PR #18, opened 2018-08-10
command/certificate: Make file names in docs more consistent

APTy:
### Description
As a first time user, it's nice when the examples in the docs are cohesive. For example, I wanted to try out a general flow of (1) create self-signed root, (2) issue intermediate, and (3) issue leaf.

With consistent file names (root-ca.crt for root, intermediate-ca.crt for intermediate), the examples flow together a little better, which makes testing out this flow much simpler on the user. I also changed the ordering a little bit.


mmalone:
Nice catch and good point. This looks good to me. Merging!

PR #20, opened 2018-08-13
Cert/inspect/stdin

APTy:
This started as simply wanting to allow for `step certificate inspect -` as a synonym for `step certificate inspect /dev/stdin`, which is a fairly common convention in CLI tools. It allows for more ergonomic shell piping.

Example:

```
cat /etc/ssl/certs/ca-certificates.crt | step certificate inspect -
```
This PR includes a few changes:

#### errs
One thing I noticed was that `errs.FileError` wasn't properly interpreting `os` errors, so I fixed that and added tests to validate its behavior.

#### utils
Next, I added `utils.ReadFile`, which returns all of the data from the named file, but allows for `-` for reading from stdin. This enables us to re-use the same behavior across the project.

#### command/certificate/inspect
Finally, I incorporated the new `utils.ReadFile` function into the `inspect` command, and added docs to reflect its usage.

Let me know what you think, thanks! :smile:

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -72,6 +72,14 @@
   revision = "2b8494104d86337cdd41d0a49cbed8e4583c0ab4"
 
 [[projects]]
+  digest = "1:a2c1d0e43bd3baaa071d1b9ed72c27d78169b2b269f71c105ac4ba34b1be4a39"
+  name = "github.com/davecgh/go-spew"
+  packages = ["spew"]
+  pruneopts = "UT"
+  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
+  version = "v1.1.0"
+
+[[projects]]
   branch = "master"
   digest = "1:4ee452f8994700dcab9e816aef1cb9eb2317218734c6ccf5135746e6c19f3dce"
   name = "github.com/golang/lint"
@@ -163,6 +171,14 @@
   version = "v0.8.0"
 
 [[projects]]
+  digest = "1:0028cb19b2e4c3112225cd871870f2d9cf49b9b4276531f03438a88e94be86fe"
+  name = "github.com/pmezard/go-difflib"
+  packages = ["difflib"]
+  pruneopts = "UT"
+  revision = "792786c7400a136282c1664665ae0a8db921c6c2"
+  version = "v1.0.0"
+
+[[projects]]
   digest = "1:6c7a3f738e37a1c7ad3d56122a34932140654d51a57e01f8613cdf3eaf050911"
   name = "github.com/pquerna/otp"
   packages = [
@@ -230,6 +246,17 @@
   ]
   pruneopts = "UT"
   revision = "d84eaafe274f9dc1f811ebfbb073e18c466e2a44"
+
+[[projects]]
+  digest = "1:c40d65817cdd41fac9aa7af8bed56927bb2d6d47e4fea566a74880f5c2b1c41e"
+  name = "github.com/stretchr/testify"
+  packages = [
+    "assert",
+    "require",
+  ]
+  pruneopts = "UT"
+  revision = "f35b8ab0b5a2cef36673838d662e249dd9c94686"
+  version = "v1.2.2"
 
 [[projects]]
   branch = "master"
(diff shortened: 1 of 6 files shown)


mmalone:
Thanks @APTy, this looks great! I think @dopey is giving the code a once over as well. Separately, I have a stylistic question to consider before merging:

We've gone back and forth a bit on using `-` to indicate reading from stdin vs simply omitting the filename. In other words, your example could simply be `cat ... | step certificate inspect` without the `-`. Obviously this is not an either/or thing: we could support omitting the filename to read from stdin *and* using `-` to explicitly read from stdin.

Relatedly, if we generalize the `-`-reads-from-stdin behavior by including it in `utils`, as suggested, are we going to create bizarre or unexpected behavior in commands that might read multiple files? What if someone runs `step certificate create foo foo.crt foo.key --ca - --ca-key -` for instance? What is desired / expected behavior? I honestly don't know and maybe it's just not worth worrying about.

APTy:
> We've gone back and forth a bit on using `-` to indicate reading from stdin vs simply omitting the filename.

It's a good question. For places where the filename is accepted as a positional argument (like this case), I think it's purely stylistic, and I'm cool with whichever way your team likes. Consistency is crucial.

In places where the filenames is accepted as a flag, that's probably where it makes the most sense to use `-`. I can update this to run with an omitted filename if you like!

dopey:
Addition looks great @APTy. @mmalone is doing some (hopefully quick) legal diligence around contributor licensing before merging. 

Separately, we were discussing the use of the `require` library for testing. We have our own internal library (`assert`) that we’ve been using. We’ve been considering changing to some open source library, but @maraino  is out of town and we want him to participate in that discussion. In the meantime we’re going to accept the PR once legal diligence is done. We’re also going to open an issue to discuss testing libraries, and hope that you would contribute any thoughts you have on that matter.

APTy:
The DCO is a pretty lightweight model, the folks at scytale use it for some of their OSS projects. Happy to do something like that if it works on your end: https://github.com/spiffe/spiffe/blob/master/CONTRIBUTING.md#sending-a-pull-request

APTy:
Happy to collaborate on discussions around testing libraries! Sorry I added `stretchr/testify/require` absentmindedly. Refactored to use `smallstep/assert` and will wait on a decision before adding extra dependencies

dopey:
@APTy link to the issue on testing libraries: https://github.com/smallstep/cli/issues/21. 

I actually prefer your use of `require` here b/c our `assert` methods don't break the flow of the test unless you call one of the methods that explicitly do that. e.g. `assert.FatalError()` or `assert.Fatal()`. I just tested the latest commit w/ `require` removed and if I set the file to one that exists in `TestFileError()` then the test panics b/c the flow continues but `err` is `nil`. This type of undocumented and unexpected behavior is a strong reason for switching to a library like `require`.  Before you change anything again (*wince, sorry!*) let's make sure that's ok w/ @mmalone.

dopey:
Ok, I think the change is better w/ your original version (using `require`). It will also serve as a reminder that we need to revisit our choice of testing package. If we end up not switching to `require` I'll go back and use the logically equivalent versions of `assert`, which will be uglier :)

Would you mind moving your PR back one commit?

APTy:
Done! Also I can squash it down to a single commit once you're happy with everything, to clean up the history a bit

mmalone:
Re: legal the question I have to answer is around patents. We chose MIT license which does not address patents at all (spiffe is Apache 2.0, which does). I’d like to use DCO or something substantially but, in the interest of good governance for the project and for smallstep, inc., I think I need to reach out to our council on this patent thing. I can see three potential outcomes: we use DCO (or nothing at all / standard GitHub inbound=outbound EULA) and things are cool, we use a CLA like the [Apache individual CLA](https://www.apache.org/licenses/icla.pdf) that explicitly addresses patent issues, or we switch to Apache 2.0 with a DCO.

If you’re ok with any of these potential outcomes I don’t see any reason we can’t move forward and merge this for now. Once we figure this out I’ll just have to get these contributions covered retroactively, which may be slightly more manual than the `git`-based flow but I’m sure won’t be too difficult.

dopey:
Sweet! If you wouldn't mind squashing it down I'll merge it in. Thanks!

PR #22, opened 2018-08-14
command/crypto/jwk/create: Add initial x5c support

APTy:
### Background
Opening this to start a bit of conversation on `x5c` support with JWKs (`step crypto jwk create`).

First, I think that re-using `--from-pem` is the right way to go, as opposed to creating a `--from-certificate` flag. Certificates behave much like public keys, with the added benefit that the key is signed and attested by some authority. Since the `--from-pem` flag already supports pubkeys, adding certificate support here makes some sense.

### Proposed Changes
This PR refactors some of the code in `jose/generate.go` and adds a switch case to allow for a JWK to be generated from a `*x509.Certificate` type.

### Example
Usage:
```
$ step crypto jwk create foo-jwks.json nopriv --from-pem path/to/foo-bundle.crt
```
Output:
```
$ cat foo-jwks.json 
{
  "use": "sig",
  "kty": "EC",
  "kid": "VX4i6RnH4axV_b9TIyPluPZCnligoiJ7rEuN5Wj5Un8",
  "crv": "P-256",
  "alg": "ES256",
  "x": "oltVQwlQoGgzKudo3yhloiZc2yPiwS2-MU4-ATTqOfk",
  "y": "0gcN9AYe5n9txmNKHEdshaXiT1VUGguCJ5cB0GpT4Ds",
  "x5c": [
    "MIIBlzCCAT2gAwIBAgIQc8XmPQaefVFuO2wRNrb/pjAKBggqhkjOPQQDAjAaMRgwFgYDVQQDEw9pbnRlcm1lZGlhdGUtY2EwHhcNMTgwODE0MDQzOTMyWhcNMTgwODE1MDQzOTMyWjAOMQwwCgYDVQQDEwNmb28wWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAASiW1VDCVCgaDMq52jfKGWiJlzbI+LBLb4xTj4BNOo5+dIHDfQGHuZ/bcZjShxHbIWl4k9VVBoLgieXAdBqU+A7o3EwbzAOBgNVHQ8BAf8EBAMCBaAwHQYDVR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUFBwMCMB0GA1UdDgQWBBRfZtG6POzQqzguPOZ6Js3ZW2EN4DAfBgNVHSMEGDAWgBTMny0YJ7oJDdhF9s1tEi2GPY/GHTAKBggqhkjOPQQDAgNIADBFAiBVVd58gE7vqK5yuZ77dMSgANzDeXt1u2hWgK2vB/EK7QIhAPyz0FD2zL6O0QCfO3stEgaPhcpWGWxmHBtpM5QdLQET",
  ]
}
```

### Limitations
No support for certificate chains. `jose.GenerateJWKFromPEM()` uses `pemutil.Read()` to parse a named file into its private key, public key, or certificate form. The `pemutil.Read()` function currently returns an error if more than one PEM block exists in the file (as is the case with certificate chains). This will be addressed in a subsequent PR.

### Open Questions
The JWK creation command takes two required arguments, the private and public JWK files:
```
$ step crypto jwk create <public-jwk-file> <private-jwk-file>
```
In the case of public keys and certificates, the `<private-jwk-file>` param is simply discarded, and no file is generated. This may also be a good time to talk about whether that API is the right one. I don't know if I have a good answer, but since encrypting the private JWK file requires a user prompt, maybe prompting for a filename (with a reasonable default) is also acceptable.

EDIT: added example usage

### DIFF
--- a/jose/generate.go
+++ b/jose/generate.go
@@ -15,6 +15,16 @@ import (
 	"golang.org/x/crypto/ed25519"
 )
 
+const (
+	jwksUsageSig = "sig"
+	jwksUsageEnc = "enc"
+)
+
+var (
+	ErrAmbiguousCertKeyUsage = errors.New("jose/generate: certificate's key usage is ambiguous, it should be for signature or encipherment, but not both (use --subtle to ignore usage field)")
+	ErrNoCertKeyUsage        = errors.New("jose/generate: certificate doesn't contain any key usage (use --subtle to ignore usage field)")
+)
+
 // GenerateJWK generates a JWK given the key type, curve, alg, use, kid and
 // the size of the RSA or oct keys if necessary.
 func GenerateJWK(kty, crv, alg, use, kid string, size int) (jwk *JSONWebKey, err error) {
@@ -57,10 +67,15 @@ func GenerateJWKFromPEM(filename string) (*JSONWebKey, error) {
 		if err != nil {
 			return nil, err
 		}
+		use, err := keyUsageForCert(crt)
+		if err != nil {
+			return nil, err
+		}
 		return &JSONWebKey{
 			Key:          key.PublicKey,
 			Certificates: []*realx509.Certificate{crt},
 			Algorithm:    algForKey(key.PublicKey),
+			Use:          use,
 		}, nil
 	default:
 		return nil, errors.Errorf("error parsing %s: unsupported key type '%T'", filename, key)
@@ -78,6 +93,41 @@ func algForKey(key crypto.PublicKey) string {
 	default:
 		return ""
 	}
+}
+
+func keyUsageForCert(cert *realx509.Certificate) (string, error) {
+	isDigitalSignature := containsUsage(cert.KeyUsage,
+		realx509.KeyUsageDigitalSignature,
+		realx509.KeyUsageContentCommitment,
+		realx509.KeyUsageCertSign,
+		realx509.KeyUsageCRLSign,
+	)
+	isEncipherment := containsUsage(cert.KeyUsage,
+		realx509.KeyUsageKeyEncipherment,
+		realx509.KeyUsageDataEncipherment,
+		realx509.KeyUsageKeyAgreement,
+		realx509.KeyUsageEncipherOnly,
+		realx509.KeyUsageDecipherOnly,
+	)
+	if isDigitalSignature && isEncipherment {
+		return "", ErrAmbiguousCertKeyUsage
+	}
+	if isDigitalSignature {
+		return jwksUsageSig, nil
+	}
+	if isEncipherment {
+		return jwksUsageEnc, nil
+	}
+	return "", ErrNoCertKeyUsage
+}
+
+func containsUsage(usage realx509.KeyUsage, queries ...realx509.KeyUsage) bool {
+	for _, query := range queries {
+		if usage&query == query {
+			return true
+		}
+	}
+	return false
 }
 
 func generateECKey(crv, alg, use, kid string) (*JSONWebKey, error) {
--- a/jose/generate.go
+++ b/jose/generate.go
@@ -67,9 +67,10 @@ func GenerateJWKFromPEM(filename string, subtle bool) (*JSONWebKey, error) {
 		if err != nil {
 			return nil, err
 		}
-		use, err := keyUsageForCert(crt)
-		if err != nil {
-			if !subtle {
+		var use string
+		if !subtle {
+			use, err = keyUsageForCert(crt)
+			if err != nil {
 				return nil, err
 			}
 		}
--- a/jose/generate.go
+++ b/jose/generate.go
@@ -1,14 +1,17 @@
 package jose
 
 import (
+	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	realx509 "crypto/x509"
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/cli/crypto/pemutil"
 	"github.com/smallstep/cli/crypto/randutil"
+	"github.com/smallstep/cli/pkg/x509"
 	"golang.org/x/crypto/ed25519"
 )
 
@@ -38,25 +41,38 @@ func GenerateJWKFromPEM(filename string) (*JSONWebKey, error) {
 	}
 
 	switch key := key.(type) {
-	case *rsa.PrivateKey, *rsa.PublicKey:
-		return &JSONWebKey{Key: key}, nil
-	case *ecdsa.PrivateKey:
+	case *rsa.PrivateKey, *rsa.PublicKey, *ecdsa.PrivateKey, *ecdsa.PublicKey, ed25519.PrivateKey, ed25519.PublicKey:
 		return &JSONWebKey{
 			Key:       key,
-			Algorithm: getECAlgorithm(key.Curve),
+			Algorithm: algForKey(key),
 		}, nil
-	case *ecdsa.PublicKey:
+	case *x509.Certificate:
+		// have: step-cli x509 Certificate
+		// want: crypto/x509 Certificate
+		crt, err := realx509.ParseCertificate(key.Raw)
+		if err != nil {
+			return nil, err
+		}
 		return &JSONWebKey{
-			Key:       key,
-			Algorithm: getECAlgorithm(key.Curve),
-		}, nil
-	case ed25519.PrivateKey, ed25519.PublicKey:
-		return &JSONWebKey{
-			Key:       key,
-			Algorithm: EdDSA,
+			Key:          key.PublicKey,
+			Certificates: []*realx509.Certificate{crt},
+			Algorithm:    algForKey(key.PublicKey),
 		}, nil
 	default:
 		return nil, errors.Errorf("error parsing %s: unsupported key type '%T'", filename, key)
+	}
+}
+
+func algForKey(key crypto.PublicKey) string {
+	switch key := key.(type) {
+	case *ecdsa.PrivateKey:
+		return getECAlgorithm(key.Curve)
+	case *ecdsa.PublicKey:
+		return getECAlgorithm(key.Curve)
+	case ed25519.PrivateKey, ed25519.PublicKey:
+		return EdDSA
+	default:
+		return ""
 	}
 }
 
--- a/jose/generate.go
+++ b/jose/generate.go
@@ -44,7 +44,7 @@ func GenerateJWK(kty, crv, alg, use, kid string, size int) (jwk *JSONWebKey, err
 
 // GenerateJWKFromPEM returns an incomplete JSONWebKey using the key from a
 // PEM file.
-func GenerateJWKFromPEM(filename string) (*JSONWebKey, error) {
+func GenerateJWKFromPEM(filename string, subtle bool) (*JSONWebKey, error) {
 	key, err := pemutil.Read(filename)
 	if err != nil {
 		return nil, err
@@ -69,7 +69,9 @@ func GenerateJWKFromPEM(filename string) (*JSONWebKey, error) {
 		}
 		use, err := keyUsageForCert(crt)
 		if err != nil {
-			return nil, err
+			if !subtle {
+				return nil, err
+			}
 		}
 		return &JSONWebKey{
 			Key:          key.PublicKey,
--- a/jose/generate.go
+++ b/jose/generate.go
@@ -21,8 +21,8 @@ const (
 )
 
 var (
-	ErrAmbiguousCertKeyUsage = errors.New("jose/generate: certificate's key usage is ambiguous, it should be for signature or encipherment, but not both (use --subtle to ignore usage field)")
-	ErrNoCertKeyUsage        = errors.New("jose/generate: certificate doesn't contain any key usage (use --subtle to ignore usage field)")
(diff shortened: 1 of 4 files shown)


mmalone:
Oof. This opens a bag of worms. Love it though!

**Thoughts on the commit**

Super happy to see a contribution here. Thank you!

Some feedback:

  * I'm ok with overloading `--from-pem` for now. That doesn't seem problematic.
  * I'd like to see [`x5t#S256`](https://tools.ietf.org/html/rfc7517#section-4.9) added by default in this case, because why not?
  * I'd definitely like to see certificate chains supported. Happy to discuss potential solutions there if you haven't come up with one already.

There is one sort of gnarly area. [RFC 7517 says](https://tools.ietf.org/html/rfc7517#section-4.7) that if a JWK has an `x5c` member then the other members of the JWK (e.g., `use`, `alg`, and "other information"?) must be "semantically consistent" with the "related fields" in the certificate. The spec is vague about what "other information" is relevant, what the "related fields" are, and what "semantically consistent" means. That makes building a generic tool to handle this particular field challenging. I'm not sure if "other information" includes the JWK `iss` member, and whether that must match the certificate issuer, for example..? If so, how do you map a DN to a string?

What seems clear is that `alg` and `use` must match, so we should probably enforce that. I suppose the "related field" for the `use` member is the [`X509v3 Key Usage` field](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) in a certificate, and I suppose the `"sig"` value would correspond to `"Digital Signature"` in a certificate (`digitalSignature` in the spec), and `"enc"` would correspond to `dataEncipherment`..? For the `alg` member the "related field" must be the `Public Key Algorithm` in the subject's public key information..?

It might be useful to reach out to the spec authors, but I suspect that if they had good answers they would have put them in the spec. Sigh.

**JWK create requires two arguments**

I agree that the API is problematic. I'm ambivalent about positional arguments for another reason: it's hard to remember which order the files go in (public, then private). Command line flags are *so much more verbose* in this case though...

I'd prefer not to use prompts as the only input mechanism because it makes scripting harder. We're actually working towards eliminating the need for password prompts by leveraging other facilities like HSMs and OS key management facilities (e.g., keychain on OS X).

I'm not really sure what the answer is here because there are so many use cases to consider. Here are some `step crypto jwk create` incantations off the top of my head, and what I'd expect the result to be:

  * `step crypto jwk create --from-pem foo.crt` -- write public JWK to stdout
  * `step crypto jwk create foo.pub.json --from-pem foo.crt` -- write public JWK to `foo.pub.json`
  * `step crypto jwk create foo.pub.json foo.key.json --from-pem foo.key` -- write public JWK to `foo.pub.json` and encrypted private key to `foo.key.json`
  * `step crypto jwk create foo.key.json --from-pem foo.key` -- write encrypted private key to `foo.key.json`
  * `step crypto jwk create --from-pem foo.key` -- write encrypted private key to stdout (I think it makes more sense to write *just* the private key, as the public key can always be derived by doing ` | step crypto jwk public`)

I don't have an answer at the moment. I'm hopeful that if we collect all the variations and use cases it will become clearer which ones we should support and how to support them. In the meantime: ¯\\_(ツ)_/¯

**`x5c` during JWT validation**

Supporting `x5c` during JWT validation raises more questions. If an `x5c` member appears in a JWT header, or in the JWK you're using to verify a JWT, then:

  * Do you validate the certificate chain by default? What if it doesn't validate?
  * What root certificate(s) do you use, if any?
  * Can the `x5c` certificate be self-signed?
  * What if the certificate key doesn't match the JWK key?
  * What if other information in the JWK isn't "semantically consistent" with information in the certificate? Once again, what does "semantically consistent" even mean?
  * If there are other key sources (e.g., `jku`, `x5u`, or `jwk` headers) which do we check first? Or maybe we check them all and they all have to be consistent?

My biggest question though is this: what are the use cases for `x5c` over using [a JWK Set at some URL](https://www.googleapis.com/oauth2/v3/certs)? I'm sure use cases exist, but I don't know what they are. That makes it hard to answer the other questions above.

Further reading:

  * RFC7517 (JWK) Sections 4.6 and 4.7: https://tools.ietf.org/html/rfc7517#section-4.6
  * RFC7515 (JWS/JWT) Section 6: https://tools.ietf.org/html/rfc7515#section-6
  * RFC7515 Appendix D (Key Selection): https://tools.ietf.org/html/rfc7515#appendix-D

APTy:
> I'd like to see x5t#S256 added by default in this case, because why not?

Agree. I would've assumed that the `jose` library would have added that automatically. I'll look into how to add this field.

> I'd definitely like to see certificate chains supported. Happy to discuss potential solutions there if you haven't come up with one already.

I have a rough branch ([diff](https://github.com/APTy/cli/compare/jwks/x5c...APTy:jwks/x5c-2)) ready with these changes.

> if a JWK has an `x5c` member then the other members of the JWK must be "semantically consistent" with the "related fields" in the certificate

To clarify a bit, the related fields MUST be semantically consistent; however, inclusion of those related fields is OPTIONAL.

- `alg`: I agree, let's make this consistent with the cert.
- `use`: I'm not sure we can determine this, couldn't an X.509 certificate signal for both "enc" and "sig"? I'm okay with omitting this field if that's the case.
- `iss`: I believe that's part of JWT, not JWK, right?

> `jwk create` requires two arguments [...]

Let's punt on this for now. I'm OK with just documenting its usage well for now.

> `x5c` during JWT validation

This is a great point, and thanks for providing the Appendix D resource on Key Selection, I hadn't read it yet; however, I've come to infer some of it's content over time. Generic validation is *very* hard, since it's largely driven by application context. Can we pick this up in a subsequent issue? I think it's important to tackle, but trying to figure out if we need to figure it out before we merge this PR.

mmalone:
Shit. I wanted this to be short but it's another wall of text. Sorry. To be honest, one of the reasons we punted on `x5c` and converting certificates to JWKs was because it's underspecified and deceptively hard to get right. I'm totally down with continuing this conversation though, but apologize if this is more than you bargained for!

> I have a rough branch ([diff](https://github.com/APTy/cli/compare/jwks/x5c...APTy:jwks/x5c-2)) ready with these changes.

High level, this looks good to me. I'd like @dopey to take a look (and @maraino when he's back from vacation). I'm curious whether there's any overlap / opportunity for code re-use with [@dopey's --bundle support](https://github.com/smallstep/cli/commit/85e55adbaedb0228960b3934717d310474f1f93d).

My only concern is whether there are any uses of the `pemutil.Parse()` API that expect a single certificate and will blow up if a list of certificates comes back (since we're returning an `interface{}` type). I found an example [in x509util/identity.go](https://github.com/smallstep/cli/blob/master/crypto/x509util/identity.go#L49) that looks suspicious. Could be fine though.

> To clarify a bit, the related fields MUST be semantically consistent; however, inclusion of those related fields is OPTIONAL.

True. We've been trying to set `kty`, `alg`, `use`, `kid`, and `crv` (for EC keys) as a best practice though. In the case of converting a certificate to a JWK it'd be nice if we could map them. There are subtle security problems with using keys for unintended purposes or with unintended algorithms (more below).

> `use`: I'm not sure we can determine this, couldn't an X.509 certificate signal for both "enc" and "sig"? I'm okay with omitting this field if that's the case.

This is a tricky / subtle subject. The short answer is that yes, an X.509 certificate can signal for both. We could omit `use` in this case, but I think we should gate that behavior with `--subtle` or maybe even `--insecure`. Alternatively we could set the `key_ops` member, but that's more work and doesn't necessarily solve the problem.

Here's why: it can actually be cryptographically dangerous to use the same key for both encryption and signing, or with unintended algorithms. Signing keys also have a different life cycle than encryption keys. When you're done signing stuff with a key you generally want to discard it so it's no longer at risk of loss. When you're done encrypting stuff with a key you generally want to keep it around to decrypt the stuff that was encrypted. So best practice is to use different keys for these operations. We do want to encourage best practices, and our tool actually enforces this in places: it always puts a `use` on your JWKs ("sig" by default) and if you try to sign something with a key marked for encryption use (or vice versa) it'll error.

The JWK `use` member enforces this best practice by construction: it's a string, not an array, so you have to pick either `enc` or `sig`, not both. For a bit more corroboration check out [section 4.7 of the JWK spec](https://tools.ietf.org/html/rfc7517#section-4.3) discussing the `key_ops` member (which is more granular than `use`, and we don't support yet). It says:

> Multiple unrelated key operations SHOULD NOT be specified for a key
   because of the potential vulnerabilities associated with using the
   same key with multiple algorithms.  Thus, the combinations "sign"
   with "verify", "encrypt" with "decrypt", and "wrapKey" with
   "unwrapKey" are permitted, but other combinations SHOULD NOT be used.

[Section 4.2 of the JWK spec](https://tools.ietf.org/html/rfc7517#section-4.2) says that "the `enc` value is also to be used for public keys used for key agreement operations" which makes me think that a "normal" X509 certificate that you'd use for TLS should result in a JWK with `"use": "enc"`. That's weird because a lot of "normal" X509 end entity TLS certificates have "Digital Signature" set as a key use.

If that's not confusing and frustrating enough, [RFC5280 says](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) that if extended key usage is set (true for TLS end entity certificates; typically set to TLS Web [Client/Server] Authentication) "then the certificate MUST only be used for one of the purposes indicated". I guess that means we have no business converting those certs into JWKs because, according to the spec, there's nothing we could use that JWK for since JWKs can't be used for TLS.

The `key_ops` member is more nuanced than `use`, but it's also a lot more complex to implement and use correctly (as indicated by the quote from its description above). There's probably a mapping that could be done between [RFC5280 Key Usage](https://tools.ietf.org/html/rfc5280#section-4.2.1.3) and [JWK Key Ops](https://tools.ietf.org/html/rfc7517#section-4.3), and maybe to JWK use. Here's a first pass:

| RFC5280 | JWK key_ops | JWK use | notes |
|----------|---------------|----------|-------|
| digitalSignature | sign | sig | |
| nonRepudiation | sign | sig | |
| keyEncipherment | wrapKey? | enc | weird because "key wrapping" is usually a symmetric operation and "key encryption / encipherment" is usually asymmetric |
| dataEncipherment | encrypt | enc | |
| keyAgreement | wrapKey? | enc | diffie-hellman is an example of key agreement. [RFC7517 says](https://tools.ietf.org/html/rfc7517#section-4.2) "The `enc` value is also to be used for public keys used for key agreement operations." |
| keyCertSign | ??? | ??? | |
| cRLSign | ??? | ??? | |
| encipherOnly | ??? | ??? | |
| decipherOnly | ??? | ??? | |

So yea, the problem remains that X509 certs tend to include conflicting operations (like `digitalSignature` and `keyEncipherment`) which the JWK RFC explicitly says you SHOULD NOT do. That said, since it's a *SHOULD NOT* and not a *MUST NOT* I think I'm ok with allowing mixed operations, but I think we should probably force the use of the `--subtle` flag because at this point I don't even know what the implications are here.

This is easy enough with the `key_ops` attribute, but I'm not sure what to do with `use`. It has to be either `enc` or `dec`. I suppose the only option is to leave it out if there's a conflict.

Unfortunately this is an area where things seem pretty underspecified. I can't find any [documents from the IETF JOSE Working Group](https://datatracker.ietf.org/group/jose/documents/) that clarify the situation. Maybe they'll publish a best practices doc or something at some point.

As I mentioned before, use cases would probably help clarify too.

> `iss`: I believe that's part of JWT, not JWK, right?

You're absolutely right. My bad. The other one that probably needs to be "semantically consistent" is `kty`. But that should be easy.

> Can we pick [`x5c` in JWTs] up in a subsequent issue?

Yes. More than enough to talk about here already :).

jose/generate.go
```diff
@@ -38,28 +41,41 @@ func GenerateJWKFromPEM(filename string) (*JSONWebKey, error) {
...
 	switch key := key.(type) {
-	case *rsa.PrivateKey, *rsa.PublicKey:
-		return &JSONWebKey{Key: key}, nil
-	case *ecdsa.PrivateKey:
-		return &JSONWebKey{
-			Key:       key,
-			Algorithm: getECAlgorithm(key.Curve),
-		}, nil
-	case *ecdsa.PublicKey:
+	case *rsa.PrivateKey, *rsa.PublicKey, *ecdsa.PrivateKey, *ecdsa.PublicKey, ed25519.PrivateKey, ed25519.PublicKey:
```
maraino: I would separate RSA keys from the others so it's easier to see that the algorithm is not automatically set in those cases.

APTy:
Sweet, I'll try to fix this up a bit today and circle back for final review

jose/generate.go
```diff
@@ -1,17 +1,30 @@
...
 )
 
+const (
+	jwksUsageSig = "sig"
+	jwksUsageEnc = "enc"
+)
+
+var (
+	ErrAmbiguousCertKeyUsage = errors.New("jose/generate: certificate's key usage is ambiguous, it should be for signature or encipherment, but not both (use --subtle to ignore usage field)")
+	ErrNoCertKeyUsage        = errors.New("jose/generate: certificate doesn't contain any key usage (use --subtle to ignore usage field)")
```
maraino: Can you add comments to this variables, the linter is failing because of this.

APTy: actually turns out we're not using them external to this package yet, so making them private for now

maraino: Sounds good

Issue #29, opened 2018-09-19
Log HTTP requests to OAuth HTTP server


mmalone:
If an unexpected request comes into the OAuth HTTP server we should log it. We should also consider having the server remain up after such a request, but we we need to think about any security considerations before doing so. Currently, the OAuth subcommand uses the `httptest` server and closes down after one request, whether or not that request is a successful OAuth redirect. If a client has some plugin (e.g., https://www.wappalyzer.com/) that sends a request to the server ahead of the OAuth redirect, the server will shutdown and the redirect won't be properly handled. At a minimum, we should output enough information that an end user can debug what's going on.

PR #30, opened 2018-09-29
Log invalid http requests into oauth server

gabrie30:
Log HTTP requests to OAuth HTTP server #29

### Description
This PR will provide more logging around invalid http requests into the oauth server. 

Currently, if any 3rd party apps or plugins, such as wappalyzer, make the first request into oauth server the redirect cannot be handled and there are no logs indicating what is happening or how to fix.

In the case of wappalyzer it's sending a request to /robots.txt, which can be used to give users a useful debugging suggestion.

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -562,6 +562,9 @@ func (o *oauth) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 
 	if code == "" || state == "" {
 		fmt.Printf("Invalid request received from: %v%v\n\n", req.RemoteAddr, req.RequestURI)
+		if req.RequestURI == "/robots.txt" {
+			fmt.Printf("** You may have an app or browser plugin that needs to be turned off **\n\n")
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


command/oauth/cmd.go
```diff
@@ -558,13 +558,20 @@ func (o *oauth) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 
-	code := q.Get("code")
+	code, state := q.Get("code"), q.Get("state")
+
+	if code == "" || state == "" {
+		fmt.Printf("Invalid request received from: %v%v\n\n", req.RemoteAddr, req.RequestURI)
+		if req.RequestURI == "/robots.txt" {
+			fmt.Printf("** You may have an app or browser plugin that needs to be turned off **\n\n")
+		}
+	}
```
maraino: Just logging the issue won't avoid the server shutdown. Here we should return a 403 or 404 in those cases, but without using the badRequest method, as it will send a signal to stop the server. 
We should also use os.Stderr to log those errors.

PR #32, opened 2018-10-19
Issue 79: Add a json reporter to quantify the extend of the cli's documenation

sourishkrout:
### Description
Implemented a simple word/line counter report based on the HTML page generation written to the output ./html directory.

https://github.com/smallstep/ca-component/issues/79

### DIFF
--- a/usage/help.go
+++ b/usage/help.go
@@ -31,6 +31,10 @@ func HelpCommand() cli.Command {
 				Usage: "The export <directory> for HTML docs.",
 			},
 			cli.StringFlag{
+				Name:  "report",
+				Usage: "The report <directory> for JSON report based on HTML docs.",
+			},
+			cli.StringFlag{
 				Name:  "markdown",
 				Usage: "The export <directory> for Markdown docs.",
 			},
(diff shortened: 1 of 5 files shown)


usage/html.go
```diff
@@ -108,14 +111,30 @@ func htmlHelpAction(ctx *cli.Context) error {
...
 	for _, cmd := range ctx.App.Commands {
-		if err := htmlHelpCommand(ctx.App, cmd, path.Join(dir, cmd.Name)); err != nil {
+		if err := htmlHelpCommand(ctx.App, cmd, path.Join(dir, cmd.Name), report); err != nil {
 			return err
 		}
 	}
+
+	// report
+	repjson := path.Join(dir, "report.json")
+	rjw, err := os.Create(repjson)
```
maraino: This will override a previous report. If you think this can be an issue you can use utils.WriteFile() that will ask the user for consent.

sourishkrout: Cool. I think I will stick with what's there now. The default behavior of the HTML help is to overwrite existing files.

usage/help.go
```diff
@@ -30,6 +30,10 @@ func HelpCommand() cli.Command {
 				Name:  "html",
 				Usage: "The export <directory> for HTML docs.",
 			},
+			cli.StringFlag{
+				Name:  "report",
+				Usage: "The report <directory> for JSON report based on HTML docs.",
+			},
```
maraino: The way that the code works, looks like this should be a `cli.BoolFlag` instead of a string with the directory, with this, we will generate the report with `step help --html <directory> --report`

sourishkrout: OK. That works for me too. I'll make the changes.

PR #45, opened 2018-12-11
Renew daemon

maraino:
### Description
Add support for a renew daemon that will renew periodically a certificate.

### DIFF
--- a/command/ca/renew.go
+++ b/command/ca/renew.go
@@ -38,6 +38,13 @@ func renewCertificateCommand() cli.Command {
 **step ca renew** command renews the given certificates on the certificate
 authority and writes the new certificate to disk either overwriting <crt-file>
 or using a new file if the **--out**=<file> flag is used.
+å
+With the **--daemon** flag the command will update periodically the given
+certificate. By default it will renew the certificate before 2/3 of the validity
+time of the certificate has passed. It uses a random jitter to avoid multiple
+instances running at the same time. The time before the expiration can be
+changed using the **--expires-in** flag, or a fixed period can be used with the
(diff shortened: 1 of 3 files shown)


command/ca/renew.go
```diff
@@ -0,0 +1,415 @@
...
+		Action: command.ActionFunc(renewCertificateAction),
+		Usage:  "renew a valid certificate",
+		UsageText: `**step ca renew** <crt-file> <key-file>
+		[**--ca-url**=<uri>] [**--root**=<file>]
+		[**--out**=<file>] [**--expires-in**=<duration>] [**--force**]`,
+		Description: `
+**step ca renew** command renews the given certificates on the certificate
+authority and writes the new certificate to disk either overwriting <crt-file>
+or using a new file if the **--out**=<file> flag is used.
+å
```
dopey: weird character. I'll do a grammatical once over once I'm through reading the rest of the commit.

maraino: Fixed.

command/ca/renew.go
```diff
@@ -0,0 +1,458 @@
...
+
+Renew the certificate and reload nginx:
+'''
+$ step ca renew --daemon --exec "nginx -s reload" internal.crt internal.key 
+'''
+
+Renew the certificate and convert it to DER:
+'''
+$ step ca renew --daemon --renew-period 16h \
+  --exec "step certificate format --force --out internal.der internal.crt" \
```
dopey: Why use --force if --out is different from the original file?

maraino: because this is going to run periodically multiple times

PR #46, opened 2018-12-12
fix ReadCertPool function

dopey:
* document acceptable input
* don't fail if input is not either a file or directory
* trim space in comma delineated file list

### DIFF
--- a/crypto/x509util/crt.go
+++ b/crypto/x509util/crt.go
@@ -13,17 +13,18 @@ import (
 )
 
 // ReadCertPool loads a certificate pool from disk.
+// *path*: a file, a directory, or a comma-separated list of files.
 func ReadCertPool(path string) (*realx509.CertPool, error) {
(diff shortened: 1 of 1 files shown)


crypto/x509util/crt.go
```diff
@@ -13,17 +13,18 @@ import (
 )
 
 // ReadCertPool loads a certificate pool from disk.
+// *path*: a file, a directory, or a comma-separated list of files.
 func ReadCertPool(path string) (*realx509.CertPool, error) {
 	info, err := os.Stat(path)
-	if err != nil {
-		return nil, errors.WithStack(err)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, errors.Wrapf(err, "os.Stat %s failed", path)
```
maraino: What happens if it's a not exists error? This snippet will show a nice error:
```go
if err != nil {
        return nil, errs.FileError(err, path)
}
```

PR #49, opened 2018-12-20
Audit of default key usage for root and intermediate certs

dopey:
* root and intermediate certs only require the CertSign key usage.
do not need key encipherment or digital signature by default.
* intermediate certificate should not have client/server auth extended
key usage.


dopey:
We're moving forwards without digital signature on root/intermediates. As of this moment, we don't have a use case for it. If users want/need it they'll let us know.

PR #52, opened 2018-12-22
Allow to add custom validity in `step certificate create`

### LINKED ISSUE #51: Support configuration of certificate validity window in `step certificate create` command.
We should probably stick to the the system we use for `step ca certificate` where we support optional --not-before and --not-after flags.

maraino:
### Description
Add support for --not-before and --not-after flags in `step certificate create`. Fixes #51

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -82,6 +82,14 @@ $ step certificate create foo foo.crt foo.key --profile leaf \
   --ca ./intermediate-ca.crt --ca-key ./intermediate-ca.key
 '''
 
+Create a leaf certificate and key with custom validity:
+
+'''
+$ step certificate create foo foo.crt foo.key --profile leaf \
+  --ca ./intermediate-ca.crt --ca-key ./intermediate-ca.key \
+  --not-before 24h --not-after 2160h
+'''
+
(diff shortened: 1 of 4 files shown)


command/certificate/create.go
```diff
@@ -217,6 +241,18 @@ func createAction(ctx *cli.Context) error {
...
 
+	notBefore, ok := flags.ParseTimeOrDuration(ctx.String("not-before"))
+	if !ok {
+		return errs.InvalidFlagValue(ctx, "not-before", ctx.String("not-before"), "")
+	}
+	notAfter, ok := flags.ParseTimeOrDuration(ctx.String("not-after"))
+	if !ok {
+		return errs.InvalidFlagValue(ctx, "not-after", ctx.String("not-after"), "")
+	}
+	if !notAfter.IsZero() && !notBefore.IsZero() && notBefore.After(notAfter) {
```
dopey: Only thing I'm not sure about is whether this check wouldn't be better performed in the WithNotBeforeAfterDuration function.

maraino: The check is there because I can return a proper error, it does not make sense to return a flag related error in WithNotBeforeAfterDuration.

PR #58, opened 2019-01-14
It's 2019

sourishkrout:
### Description
Replace mention of `2018` with `2019`. We've moved licenses from MIT to APL-2.0 as of smallstep/cli/issues/34. Let's be sure the debian package reflects that.

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -3,11 +3,11 @@ Upstream-Name: step
 Source: https://github.com/smallstep/cli
 
 Files: *
-Copyright: 2018 Smallstep Labs, Inc.
-License: MIT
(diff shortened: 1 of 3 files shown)


debian/copyright
```diff
@@ -3,11 +3,11 @@ Upstream-Name: step
 Source: https://github.com/smallstep/cli
 
 Files: *
-Copyright: 2018 Smallstep Labs, Inc.
-License: MIT
+Copyright: 2019 Smallstep Labs, Inc.
+License: Apache
```
sourishkrout: @dopey the debian docs say `Apache` for APL-1.0 & APL-2.0. Can you take a look too just to be sure this will work? Thanks

Issue #56, opened 2019-01-03
Fingerprint certificate of remote server


mmalone:
It'd be nice if we could `fingerprint` a certificate on a remote machine the same way we can `inspect` one. That is, I'd like to be able to do the following:

```
step certificate fingerprint https://smallstep.com
```

dopey:
Added the ability to fingerprint a certificate on a remote machine.

```
step certificate fingerprint https://smallstep.com
```

PR #61, opened 2019-01-16
Fingerprint certificate of remote server

dopey:
* addresses #56

### Description
Please describe your pull request.

💔Thank you!

### DIFF
--- a/command/certificate/fingerprint.go
+++ b/command/certificate/fingerprint.go
@@ -2,10 +2,12 @@ package certificate
 
 import (
 	"crypto/sha256"
+	"crypto/x509"
 	"encoding/hex"
 	"fmt"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/smallstep/cli/crypto/pemutil"
 	"github.com/smallstep/cli/errs"
 	"github.com/urfave/cli"
@@ -31,7 +33,36 @@ Get the fingerprint for a root certificate:
 '''
 $ step certificate fingerprint /path/to/root_ca.crt
 0d7d3834cf187726cf331c40a31aa7ef6b29ba4df601416c9788f6ee01058cf3
+'''
+
+Get the fingerprint for a remote certificate:
+'''
+$ step certificate fingerprint https://smallstep.com
+0d7d3834cf187726cf331c40a31aa7ef6b29ba4df601416c9788f6ee01058cf3
 '''`,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name: "roots",
+				Usage: `Root certificate(s) that will be used to verify the
+authenticity of the remote server.
+
+: <roots> is a case-sensitive string and may be one of:
+
+    **file**
+	:  Relative or full path to a file. All certificates in the file will be used for path validation.
+
+    **list of files**
+	:  Comma-separated list of relative or full file paths. Every PEM encoded certificate from each file will be used for path validation.
+
+    **directory**
+	:  Relative or full path to a directory. Every PEM encoded certificate from each file in the directory will be used for path validation.`,
+			},
+			cli.BoolFlag{
+				Name: "insecure",
+				Usage: `Use an insecure client to retrieve a remote peer certificate. Useful for
+debugging invalid certificates remotely.`,
+			},
+		},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


command/certificate/fingerprint.go
```diff
@@ -40,9 +71,25 @@ func fingerprintAction(ctx *cli.Context) error {
...
-		return err
+	var (
+		crt      *x509.Certificate
+		err      error
+		roots    = ctx.String("roots")
+		insecure = ctx.Bool("insecure")
+		crtFile  = ctx.Args().First()
+	)
+
+	if strings.HasPrefix(crtFile, "https://") {
```
maraino: It doesn't have to be `https://, we can should accept things like `tcp://` or `tls://`

We should do the same in `step certificate inspect`, use a method for both.

command/certificate/inspect.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/smallstep/certinfo"
 	"github.com/smallstep/cli/crypto/x509util"
 	"github.com/smallstep/cli/errs"
-	x509 "github.com/smallstep/cli/pkg/x509"
+	stepx509 "github.com/smallstep/cli/pkg/x509"
```
maraino: 👍 nice

command/certificate/fingerprint.go
maraino: Add docs for the prefixes supported.

command/certificate/remote.go
```diff
@@ -0,0 +1,58 @@
...
+	conn, err := tls.Dial("tcp", addr, tlsConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to connect")
+	}
+	conn.Close()
+	return conn.ConnectionState().PeerCertificates, nil
+}
+
+func trimURLPrefix(url string) (bool, string, string) {
+	for _, prefix := range urlPrefixes {
```
maraino: url = strings.ToLower(url)

command/certificate/remote.go
```diff
@@ -0,0 +1,58 @@
...
+			return true, prefix, strings.TrimPrefix(url, prefix)
+		}
+	}
+	return false, "", ""
+}
+
+// isURL returns true if the input string is formatted as a URL with one of the
+// documented prefixes.
+func isURL(str string) bool {
+	for _, prefix := range urlPrefixes {
```
maraino: str = strings.ToLower(str)

command/certificate/remote.go
maraino: But it will return the URL lowercase, can this be a problem? 
We can always use a temporary URL and return the original one with `url[len(prefix):]`

command/certificate/remote.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	conn, err := tls.Dial("tcp", addr, tlsConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to connect")
+	}
+	conn.Close()
+	return conn.ConnectionState().PeerCertificates, nil
+}
+
+func trimURLPrefix(url string) (bool, string, string) {
```
maraino: Usually, the methods have the bool return values as the last one (if there's no error)

command/certificate/lint.go
```diff
@@ -93,30 +94,17 @@ func lintAction(ctx *cli.Context) error {
...
-		addr := strings.TrimPrefix(crtFile, "https://")
-		if !strings.Contains(addr, ":") {
-			addr += ":443"
-		}
-		conn, err := tls.Dial("tcp", addr, &tls.Config{RootCAs: rootCAs})
+	if isURL, _, addr := trimURLPrefix(crtFile); isURL {
+		peerCertificates, err := getPeerCertificates(addr, roots, insecure)
 		if err != nil {
-			return errors.Wrapf(err, "failed to connect")
+			return errors.WithStack(err)
```
maraino: if getPeerCertificates returns an error from the `pkg/errors` package it already have a stack, there's no need to replace it as we might drop part of the stack.

command/certificate/remote.go
```diff
@@ -0,0 +1,58 @@
...
+		if strings.HasPrefix(url, prefix) {
+			return true, prefix, strings.TrimPrefix(url, prefix)
+		}
+	}
+	return false, "", ""
+}
+
+// isURL returns true if the input string is formatted as a URL with one of the
+// documented prefixes.
+func isURL(str string) bool {
```
maraino: Are you using this method? If not the linter will fail

PR #71, opened 2019-01-31
Sans

dopey:
### Add SAN support to `step certificate create` and `step ca certificate`

Currently the CLI does not support adding Subject Alternative Names. The user can provide one cmd line arg which doubles as a Common Name and the only subject alternative name (in `step certificate create` it is not duplicated in the SANs).

This PR adds the ability to configure SANs from the cli. If no SANs are configured then:

1) `step ca certificate` will duplicate the <subject> cmd line arg as both a subject common name and the only SAN.
2) `step certificate create` will use the <subject> cmd line arg as a subject common name and have no SANs.

If at least one SAN is configured (via the cmd line) then only those SANs which are explicitly configured will be set.


dopey:
I will remove the dep on 'sans' branch of certificates before merging (update the dependency and run dep ensure). Just wanted it for testing.

command/ca/certificate.go
```diff
@@ -25,16 +25,18 @@ func newCertificateCommand() cli.Command {
 		Name:   "certificate",
 		Action: command.ActionFunc(newCertificateAction),
 		Usage:  "generate a new private key and certificate signed by the root certificate",
-		UsageText: `**step ca certificate** <hostname> <crt-file> <key-file>
+		UsageText: `**step ca certificate** <common-name> <crt-file> <key-file>
```
maraino: For consistency with `step ca token` let's call this `<subject>`

Issue #79, opened 2019-02-20
step certificate fingerprint <cert-file> fails when bundle (leaf+intermediate) is presented


sourishkrout:
### Subject of the issue
Fingerprint throws an error when <cert-file> contains more than one cert which is common for leaf+intermediate cert bundles.

### Steps to reproduce
```
$ step certificate fingerprint remote_site.crt
```

Where `remote_site.crt` is:

```
-----BEGIN CERTIFICATE-----
MIICEjCCAbmgAwIBAgIRAJPWY9Bd2wgIPKHeWs4zahMwCgYIKoZIzj0EAwIwJDEi
MCAGA1UEAxMZU21hbGxzdGVwIEludGVybWVkaWF0ZSBDQTAeFw0xOTAyMjAwMDA1
NDFaFw0xOTAyMjEwMDA1NDFaMBAxDjAMBgNVBAMTBW5naW54MFkwEwYHKoZIzj0C
AQYIKoZIzj0DAQcDQgAEks8Ektkp9QL9hNIxv1jwvaig7mTzSeBfNn4sR3UOQKTG
fyC+/1GyULnujFZJVDyCXOiynWsrVeuwTmNZq7Y7WaOB3zCB3DAOBgNVHQ8BAf8E
BAMCBaAwHQYDVR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUFBwMCMB0GA1UdDgQWBBTb
tzlT8soA7CmfQzEWc+zDw101czAfBgNVHSMEGDAWgBTx0ojLlksfaHeKFaCqNymi
jl6rdjAQBgNVHREECTAHggVuZ2lueDBZBgwrBgEEAYKkZMYoQAEESTBHAgEBBBVt
YXJpYW5vQHNtYWxsc3RlcC5jb20EK0RtQXRadDJFaG1acl9pVEpKMzg3ZnI0TWQy
TmJ6TVhHZFhRTlcxVVdQWGswCgYIKoZIzj0EAwIDRwAwRAIgCLAEz4niywfTAsXe
msFt1sB1D5S7NQq8WWCArDonzzYCIBOAIcdyZv2pM4QhU0/auGgjOJDPmpbuGuJV
lWXj+wKd
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MIIBxjCCAWugAwIBAgIQAYoOWhdChUmmKzlc0DWcWDAKBggqhkjOPQQDAjAcMRow
[...]

PR #81, opened 2019-02-21
Improvements on certificate commands

maraino:
### Description
This PR introduces the following improvements:
* Add support for `--short` flag in `step certificate inspect` (smallstep/certificates#30)
* Add support for `--bundle` flag in `step certificate fingerprint`, and print just the first one if the flag is not passed. (#79)
* Add short certificate information on `step certificate install` and `step certificate uninstall` (smallstep/step#149)

Depends on smallstep/certinfo/pull/2


command/certificate/fingerprint.go
```diff
@@ -37,7 +42,14 @@ $ step certificate fingerprint /path/to/root_ca.crt
...
 '''
 $ step certificate fingerprint https://smallstep.com
-0d7d3834cf187726cf331c40a31aa7ef6b29ba4df601416c9788f6ee01058cf3
+e2c4f12edfc1816cc610755d32e6f45d5678ba21ecda1693bb5b246e3c48c03d
+'''
+
+Get the fingerprints for a remote certificate with its intemediate:
+'''
+$ step certificate fingerprint --bundle https://smallstep.com
+e2c4f12edfc1816cc610755d32e6f45d5678ba21ecda1693bb5b246e3c48c03d
```
dopey: maybe something like 
```
0: e2c4f12edfc1816cc610755d32e6f45d5678ba21ecda1693bb5b246e3c48c03d
1: 25847d668eb4f04fdd40b12b6b0740c567da7d024308eb6c2c96fe41d9de218d
```
would make this more clear? nitpicky suggestion.

command/certificate/fingerprint.go
```diff
@@ -71,27 +87,34 @@ func fingerprintAction(ctx *cli.Context) error {
...
+		certs, err = pemutil.ReadCertificateBundle(crtFile)
 		if err != nil {
 			return err
 		}
 	}
 
-	sum := sha256.Sum256(crt.Raw)
-	fmt.Println(strings.ToLower(hex.EncodeToString(sum[:])))
+	if !bundle {
+		certs = certs[:1]
```
dopey: certs[0]?

command/certificate/inspect.go
```diff
@@ -246,108 +287,57 @@ func inspectAction(ctx *cli.Context) error {
...
+		return nil
+	default:
+		return errs.InvalidFlagValue(ctx, "format", format, "text, json")
+	}
+}
 
-		if _, addr, isURL := trimURLPrefix(crtFile); isURL {
-			peerCertificates, err := getPeerCertificates(addr, roots, insecure)
+func inspectCertificateRequest(ctx *cli.Context, block *pem.Block) error {
+	format, short := ctx.String("format"), ctx.Bool("short")
```
dopey: we should probably InvalidFlagValue if you request format=json and short, right?

command/certificate/install.go
```diff
@@ -172,7 +173,14 @@ func installAction(ctx *cli.Context) error {
 		}
 	}
 
-	ui.Printf("Certificate %s has been installed.", filename)
+	fmt.Printf("Certificate %s has been installed.\n", filename)
```
dopey: any reason for switching from ui.print here and in the uninstallAction?

command/certificate/fingerprint.go
maraino: we need a slice

crypto/pemutil/pem.go
```diff
@@ -165,6 +165,47 @@ func ReadCertificate(filename string) (*realx509.Certificate, error) {
...
+func ReadCertificateBundle(filename string) ([]*realx509.Certificate, error) {
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return nil, errs.FileError(err, filename)
+	}
+
+	// PEM format
+	if bytes.HasPrefix(b, []byte("-----BEGIN ")) {
+		var block *pem.Block
+		var bundle []*realx509.Certificate
```
dopey: since you're here we could probably just change all the `realx509` -> `x509`, and `x509` -> `stepx509`. shrug

command/certificate/fingerprint.go
maraino: Fixed with b3a4af9bf79a849cdd3dd17f2143dc522fa9e162

command/certificate/fingerprint.go
maraino: We need a slice.

command/certificate/inspect.go
maraino: Fixed with c95f1d3c1b56f4c1ed70c6f280d477d93bd71ba4

command/certificate/install.go
maraino: The actual reason was that `ui.Print` does not exist :)
And I thought, this is the only output of this command, and there's no reason to do it in Stderr (ui always writes there to allow pipes and >)

crypto/pemutil/pem.go
maraino: Fixed with 4d2bd33c8d5ffb8735e2629505e5b7bce2375b2b

command/certificate/install.go
dopey: har har. Ok makes sense.

PR #96, opened 2019-03-21
Use standard x509 as much as possible

### LINKED ISSUE #95: Use crypto/x509 as main x509 package in x509util
### Description
`cli/crypto/x509util` is using our internal x509 implementation that adds support for Ed25519. We can change it to use the standard one and use only the internal for the specific cases where it's really required.

The CA does not support Ed25519 yet, so we only need to support it for `step certificate` commands.

maraino:
### Description
This PR tries to remove the use of the internal `cli/pkg/x509` as much as possible without affecting to the current functionality.

Fixes #95

### DIFF
--- a/crypto/x509util/convert.go
+++ b/crypto/x509util/convert.go
@@ -3,12 +3,44 @@ package x509util
 import (
 	"crypto/x509"
 
+	"github.com/pkg/errors"
 	stepx509 "github.com/smallstep/cli/pkg/x509"
 )
 
-// ToStepX509 converts a x509.Certificate from the standard library to the step
-// version of the x509.Certificate.
-func ToStepX509(cert *x509.Certificate) *stepx509.Certificate {
+// ParseCertificate parses a single certificate from the given ASN.1 DER data.
+func ParseCertificate(asn1Data []byte) (*x509.Certificate, error) {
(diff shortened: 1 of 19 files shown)


crypto/x509util/convert.go
```diff
@@ -0,0 +1,205 @@
...
+// the standard version of the x509.Certificate.
+func ToX509Certificate(cert *stepx509.Certificate) *x509.Certificate {
+	return &x509.Certificate{
+		Raw:                         cert.Raw,
+		RawTBSCertificate:           cert.RawTBSCertificate,
+		RawSubjectPublicKeyInfo:     cert.RawSubjectPublicKeyInfo,
+		RawSubject:                  cert.RawSubject,
+		RawIssuer:                   cert.RawIssuer,
+		Signature:                   cert.Signature,
+		SignatureAlgorithm:          x509.SignatureAlgorithm(cert.SignatureAlgorithm),
```
dopey: what if the sigalg was ED25519? Won't this lead to an invalid x509.Certificate?

maraino: Yes and no, SignatureAlgorithm is just an `int`, the problem would happen when you are signing or marshaling that certificate, not at this time.

PR #98, opened 2019-03-22
Use default token duration in step ca certificate / sign

### LINKED ISSUE #97: Use default token duration in step ca certificate and sign
### Description
The commands `step ca certificate` and `step ca sign` should only use the `--not-after` and `--not-before` flags for certificate validity, not token validity.

maraino:
### Description
This PR forces the use of default token validity when `step ca certificate` and `step ca sign` are used without the `--token` flag. The token validity is irrelevant as the token is used immediately.

Fixes #97

### DIFF
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -299,7 +299,7 @@ func (f *certificateFlow) GenerateToken(ctx *cli.Context, subject string, sans [
 	return newTokenFlow(ctx, subject, sans, caURL, root, "", "", "", "", time.Time{}, time.Time{})
 }
 
-// Sign signs the token using the online or the offline certificate authority.
+// Sign signs the CSR using the online or the offline certificate authority.
 func (f *certificateFlow) Sign(ctx *cli.Context, token string, csr api.CertificateRequest, crtFile string) error {
 	client, err := f.getClient(ctx, csr.Subject.CommonName, token)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


command/ca/certificate.go
```diff
@@ -266,10 +267,11 @@ func (f *certificateFlow) getClient(ctx *cli.Context, subject, token string) (ca
 	return ca.NewClient(caURL, options...)
 }
 
+// GenerateToken generates the token using the token flow or the online mode.
```
dopey: "offline token flow"?

dopey: GenerateToken generates a token for immediate use (therefore only default validity values will be used). The token is generated either with the offline token flow or the online mode.

command/ca/certificate.go
```diff
@@ -304,9 +296,10 @@ func (f *certificateFlow) GenerateToken(ctx *cli.Context, subject string, sans [
 		}
 	}
 
-	return newTokenFlow(ctx, subject, sans, caURL, root, "", "", "", "", notBefore, notAfter)
+	return newTokenFlow(ctx, subject, sans, caURL, root, "", "", "", "", time.Time{}, time.Time{})
 }
 
+// Sign signs the token using the online or the offline certificate authority.
```
dopey: Sign signs the CSR using the online or the offline certificate authority. -- right? (CSR not token, or am I confused)

command/ca/certificate.go
maraino: offline flow, yes

PR #104, opened 2019-04-02
Add passive revocation

dopey:
Add passive Revocation to the CLI.

Passive revocation prevents a certificate from being renewed in the future.

The command should be under the `step ca` branch of subcommands.


command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+
+func revokeCertificateCommand() cli.Command {
+	return cli.Command{
+		Name:   "revoke",
+		Action: command.ActionFunc(revokeCertificateAction),
+		Usage:  "revoke a certificate",
+		UsageText: `**step ca revoke**
+[**--crt**=<certificate>] [**--key**=<key>] [**--token**=<ott>]
+[**-serial**=<serial-number>] [**--ca-url**=<uri>] [**--root**=<file>]
+[**--reason**=<reason>] [**--reasonCode**=<code>] [**-offline**]`,
```
maraino: `--serial` is not currently being used and `<serial-number>` is missing as a positional argument.

command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+		if len(keyFile) == 0 {
+			return errs.RequiredWithFlag(ctx, "crt", "key")
+		}
+		if len(token) > 0 {
+			errs.IncompatibleFlagWithFlag(ctx, "crt", "token")
+		}
+		if len(serial) > 0 {
+			errs.IncompatibleFlagWithFlag(ctx, "crt", "serial")
+		}
```
maraino: Incompatible with example in help: `step ca revoke --crt mike.crt --key mike.key 308893286343609293989051180431574390766`

command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+		crt, err := pemutil.ReadCertificateBundle(crtFile)
+		if err != nil {
+			return err
+		}
+		serial = crt[0].SerialNumber.String()
```
maraino: If we allow serial as a positional argument we should check it.

command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+
+		// If there is no token then we must be doing a Revoke over mTLS.
+		cert, err := tls.LoadX509KeyPair(crtFile, keyFile)
+		if err != nil {
+			return errors.Wrap(err, "error loading certificates")
+		}
+		if len(cert.Certificate) == 0 {
+			return errors.New("error loading certificate: certificate chain is empty")
+		}
+		rootCAs, err := x509util.ReadCertPool(ctx.String("root"))
```
maraino: Sometimes root is loaded from the default location

command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+
+func (f *revokeFlow) Revoke(ctx *cli.Context, serial, token string) error {
+	client, err := f.getClient(ctx, serial, token)
+	if err != nil {
+		return err
+	}
+
+	reason := ctx.String("reason")
+	// Convert the reasonCode flag to an OCSP revocation code.
+	reasonCode, err := ReasonCodeToNum(ctx.String("reasonCode"))
```
maraino: ReasonCode is validated very late in the flow. We should validate it before

command/ca/revoke.go
```diff
@@ -0,0 +1,391 @@
...
+	// default to 0
+	if rc == "" {
+		return 0, nil
+	}
+
+	if code, err := strconv.Atoi(rc); err == nil {
+		return code, nil
+	}
+
+	code, found := RevocationReasonCodes[strings.ToLower(strings.Replace(rc, " ", "", 1))]
```
maraino: Replace all spaces 1=> -1

maraino: Do not use strings.ReplaceAll as it was just added recently

command/ca/revoke.go
dopey: The reason I'm not checking it is that you can just call `step ca revoke` (no args) and it will prompt you for a serial number.

command/ca/revoke.go
dopey: I don't understand what you are recommending / what you would like changed ??

maraino: https://github.com/smallstep/cli/blob/master/command/ca/certificate.go#L286-L292

PR #111, opened 2019-04-24
Cloud identities

maraino:
### Description
This PR add support for using the new cloud provisioners in the `step ca` commands:
* GCP instance identities
* AWS instance identities
* Azure instances using OAuth tokens

Related to https://github.com/smallstep/certificates/pull/65

### DIFF
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -34,7 +34,7 @@ func certificateCommand() cli.Command {
 		Action: command.ActionFunc(certificateAction),
 		Usage:  "generate a new private key and certificate signed by the root certificate",
 		UsageText: `**step ca certificate** <subject> <crt-file> <key-file>
-		[**--token**=<token>] [**--ca-url**=<uri>] [**--root**=<file>]
+		[**--token**=<token>]  [**--issuer**=<name>] [**--ca-url**=<uri>] [**--root**=<file>]
 		[**--not-before**=<time|duration>] [**--not-after**=<time|duration>]
 		[**--san**=<SAN>]`,
 		Description: `**step ca certificate** command generates a new certificate pair
@@ -89,6 +89,7 @@ $ step ca certificate --token $(step oauth --oidc --bare) joe@example.com joe.cr
 '''`,
 		Flags: []cli.Flag{
 			tokenFlag,
+			provisionerIssuerFlag,
 			caURLFlag,
 			rootFlag,
 			notBeforeFlag,
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -22,7 +22,6 @@ import (
 	"github.com/smallstep/cli/crypto/x509util"
 	"github.com/smallstep/cli/errs"
 	"github.com/smallstep/cli/flags"
-	"github.com/smallstep/cli/jose"
 	"github.com/smallstep/cli/token"
 	"github.com/smallstep/cli/ui"
 	"github.com/smallstep/cli/utils"
@@ -133,16 +132,9 @@ func certificateAction(ctx *cli.Context) error {
 		return err
 	}
 
-	var isStepToken bool
 	if len(tok) == 0 {
 		if tok, err = flow.GenerateToken(ctx, subject, sans); err != nil {
 			return err
-		}
-		isStepToken = isStepCertificatesToken(tok)
-	} else {
-		isStepToken = isStepCertificatesToken(tok)
-		if isStepToken && len(sans) > 0 {
-			return errs.MutuallyExclusiveFlags(ctx, "token", "san")
 		}
 	}
 
@@ -158,6 +150,9 @@ func certificateAction(ctx *cli.Context) error {
 
 	switch jwt.Payload.Type() {
 	case token.JWK: // Validate that subject matches the CSR common name.
+		if ctx.String("token") != "" && len(sans) > 0 {
+			return errs.MutuallyExclusiveFlags(ctx, "token", "san")
+		}
 		if strings.ToLower(subject) != strings.ToLower(req.CsrPEM.Subject.CommonName) {
 			return errors.Errorf("token subject '%s' and common name '%s' do not match", req.CsrPEM.Subject.CommonName, subject)
 		}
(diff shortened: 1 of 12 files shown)


command/ca/token.go
```diff
@@ -513,20 +526,29 @@ func provisionerPrompt(ctx *cli.Context, provisioners provisioner.List) (provisi
...
-			id = p.Key.KeyID
+			name = "Key ID"
+			value = p.Key.KeyID + " (" + p.Name + ")"
 		case *provisioner.OIDC:
-			name = p.Name
-			id = p.ClientID
+			name = "Client ID"
+			value = p.ClientID + " (" + p.Name + ")"
+		case *provisioner.GCP:
+			name = "Provisioner"
```
dopey: Not sure Provisioner is the most clear name here. maybe Cloud Identity?

command/ca/certificate.go
```diff
@@ -131,40 +132,54 @@ func certificateAction(ctx *cli.Context) error {
...
-		// Validate that the subject matches an email SAN
+	case token.OIDC: // Validate that the subject matches an email SAN
 		if len(req.CsrPEM.EmailAddresses) == 0 {
 			return errors.New("unexpected token: payload does not contain an email claim")
 		}
 		if email := req.CsrPEM.EmailAddresses[0]; email != subject {
 			return errors.Errorf("token email '%s' and argument '%s' do not match", email, subject)
 		}
+	case token.AWS: // Validate that the subject matches the instance id
+		if strings.ToLower(subject) != strings.ToLower(req.CsrPEM.Subject.CommonName) {
```
dopey: I think the error messages here are confusing (not a result of this commit, because I don't think you changed the wording). Saying that the token subject doesn't match the csr is misleading. The CSR is generated from the token. So what's actually happening is that the `subject` input from the cli does not match the identity derived from the token. 

The current phrasing is confusing to me and I think we should change for all the cases. We don't need to check that the token matches the CSR b/c authority.AuthorizeSign will do that for us.

command/ca/certificate.go
```diff
@@ -233,25 +229,38 @@ func (f *certificateFlow) getClient(ctx *cli.Context, subject, token string) (ca
...
 	if err != nil {
 		return nil, errors.Wrap(err, "error parsing flag '--token'")
 	}
-	var claims tokenClaims
-	if err := tok.UnsafeClaimsWithoutVerification(&claims); err != nil {
-		return nil, errors.Wrap(err, "error parsing flag '--token'")
-	}
-	if strings.ToLower(claims.Subject) != strings.ToLower(subject) {
-		return nil, errors.Errorf("token subject '%s' and CSR CommonName '%s' do not match", claims.Subject, subject)
+	switch jwt.Payload.Type() {
```
dopey: I don't understand the necessity to perform these checks.

The only place I see calling `getClient` uses the subject from the `CSR` (line 316) which was generated from the token (in the context of the certificate action). In the CreateSignRequest method you populate the CN field of the CSR (from the token) and then in this method you are verifying that the value from the CSR matches the value from the token. Unless I'm missing something (which I probably am) I can't see the reason to check a value that you just set. 

I'm a bit confused.

command/ca/token.go
```diff
@@ -513,20 +526,29 @@ func provisionerPrompt(ctx *cli.Context, provisioners provisioner.List) (provisi
 	}
 
 	if len(provisioners) == 1 {
-		var id, name string
+		var name, value string
 		switch p := provisioners[0].(type) {
```
dopey: Maybe use `switch p.GetType()` as you did above.

command/ca/token.go
maraino: That `name` causes that the CLI shows something like when there's only one:
```
✔ Provisioner: My Provisioner Name (GCP)
```
I think is the right name, because the user should know the concept of provisioners.

In fact, I would prefer to unify everything to something like:
```
✔Provisioner: Name (TYPE) [(other info)]
```
So it can look like:
```
✔Provisioner: mariano@smallstep.com (JWK) (kid: asdfasdfasdfasfasdf)
or 
✔Provisioner: Google (OIDC) (client: asdfasdfasdfasdfasfd)
```
or something similar

command/ca/certificate.go
maraino: I'll clarify a little bit these errors, those errors won't be seen by the user unless we introduce a bug when we generate the CSR

command/ca/token.go
dopey: What's confusing (to me) is that the `Name` attribute for JWK and OIDC is more specific (in that same block), but the one for cloud identities is just `Provisioner`. They are all provisioners, so to me that `Name` should be more specific (e.g. type of Provisioner) if you're going to do that for JWK and OIDC.

command/ca/certificate.go
dopey: Yea, clarification will help, especially for us when debugging.

command/ca/certificate.go
maraino: We're not generating the certificate from the token if you use `step ca sign`

dopey: I see. So the CertificateFlow is reused for other actions. In this case the checks are redundant, but they won't always be.

command/ca/token.go
maraino: We want to use p as type, for example, to get struct properties or call methods that are not part of the provisioner interface.

dopey: gotcha 👍

command/ca/offline.go
```diff
@@ -228,6 +229,23 @@ func (c *offlineCA) GenerateToken(ctx *cli.Context, typ int, subject string, san
 		return "", err
 	}
 
+	switch p := p.(type) {
+	case *provisioner.OIDC: // Run step oauth
```
dopey: just leaving a marker here for myself, hard to tell if it was addressed in a future commit.

command/ca/provisioner/add.go
```diff
@@ -15,33 +15,48 @@ import (
...
-)
-
 func addCommand() cli.Command {
 	return cli.Command{
 		Name:   "add",
 		Action: cli.ActionFunc(addAction),
 		Usage:  "add one or more provisioners the CA configuration",
 		UsageText: `**step ca provisioner add** <name> <jwk-file> [<jwk-file> ...]
-		[**--ca-config**=<file>] [**--create**] [**--password-file**=<file>]`,
+**--ca-config**=<file> [**--type**=JWK]  [**--create**] [**--password-file**=<file>]
```
dopey: shouldn't this go in examples instead of usage text?

maraino: This method is quite complicated, so I wanted to make make the usage more verbose, with the main different methods to use it

It looks something like:
```
USAGE
      step ca provisioner add name jwk-file [jwk-file ...] --ca-config=file
      [--type=JWK] [--create] [--password-file=file]

      step ca provisioner add name --type=OIDC --ca-config=file [--client-id=id]
      [--client-secret=secret] [--configuration-endpoint=url] [--domain=domain]
      [--admin=email]...

      step ca provisioner add --type=[AWS|Azure|GCP] --ca-config=file
      [--aws-account=id] [--gcp-service-account=name] [--gcp-project=name]
      [--azure-tenant=id] [--azure-resource-group=name] [--instance-age=duration]
      [--disable-custom-sans] [--disable-trust-on-first-use]
```

PR #114, opened 2019-05-17
Add test command step fileserver

maraino:
### Description
This PR adds a new hidden command `step fileserver`. 

This command starts a file server on a given directory, it supports HTTP and HTTPS with a given certificate file.

### DIFF
--- a/command/fileserver/fileserver.go
+++ b/command/fileserver/fileserver.go
@@ -20,8 +20,8 @@ func init() {
 		Hidden: true,
 		Action: command.ActionFunc(fileServerAction),
 		Usage:  "start an HTTP(S) server serving the contents of a path",
-		UsageText: `step fileserver <root>
-		[--address=<address>] [--cert=<path>] [--key=<path>`,
+		UsageText: `step fileserver <dir>
+		[--address=<address>] [--cert=<path>] [--key=<path>]`,
 		Description: `**step fileserver** command starts an HTTP(S) server serving the contents of a file
 system.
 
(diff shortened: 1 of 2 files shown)


command/fileserver/fileserver.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:   "fileserver",
+		Hidden: true,
+		Action: command.ActionFunc(fileServerAction),
+		Usage:  "start an HTTP(S) server serving the contents of a path",
+		UsageText: `step fileserver <root>
+		[--address=<address>] [--cert=<path>] [--key=<path>`,
```
dopey: missing close bracket after key

command/fileserver/fileserver.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/smallstep/cli/command"
+	"github.com/urfave/cli"
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:   "fileserver",
+		Hidden: true,
+		Action: command.ActionFunc(fileServerAction),
+		Usage:  "start an HTTP(S) server serving the contents of a path",
```
dopey: maybe say "contents of a directory" since you're calling the arg <dir>

maraino: 👍

maraino: the usage would be too long to keep it in one line, leaving it as path

PR #115, opened 2019-06-12
Add command for base64 encoding/decoding 

maraino:
### Description
This PR adds a new command `step base64` to base64 encoding or decoding.

Decoding format is automatically detected, while the encoding supports the standard encoding as well as url encoding both with or without padding.

### DIFF
--- a/command/base64/base64.go
+++ b/command/base64/base64.go
@@ -1,0 +1,136 @@
+package base64
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/command"
+	"github.com/smallstep/cli/utils"
+	"github.com/urfave/cli"
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:      "base64",
+		Action:    command.ActionFunc(base64Action),
+		Usage:     "encodes and decodes b64 data.",
+		UsageText: `**step base64** [**-d**|**--decode**] [**-r**|**--raw**] [**-u**|**--url**]`,
+		Description: `**step base64** implements base64 encoding as specified by RFC 4648.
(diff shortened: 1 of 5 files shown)


command/base64/base64.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/smallstep/cli/command"
+	"github.com/smallstep/cli/utils"
+	"github.com/urfave/cli"
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:      "base64",
+		Action:    command.ActionFunc(base64Action),
+		Usage:     "encodes and decodes b64 data.",
```
dopey: I don't think we normally have a period at the end of the usage

command/base64/base64.go
```diff
@@ -0,0 +1,136 @@
...
+VGhpcyBpcyB0aGUgc3RyaW5nIHRvIGVuY29kZQ
+'''
+
+Encode to base64 using the url encoding:
+'''
+$ echo 'abc123$%^&*()_+-=~' | step base64 -u
+YWJjMTIzJCVeJiooKV8rLT1-Cg==
+'''
+
+Decode an url encoded base64 string:
```
dopey: should these be equivalent? weird example to show if they are both the same.

maraino: The second one with the `-u` flag enforces the URL encoding. Without it will autodetect it. 
We can get rid of the second example if we want, the most common case should be without the `-u`.

dopey: sure, I just mean that based on the examples it appears that the two don't actually do anything different. which is probably not what you're trying to show.

maraino: I'll add a comment.

command/base64/base64.go
maraino: Fixed with 592fd53

PR #119, opened 2019-07-25
SSH CA - CLI support

maraino:
### Description
This PR adds support for signing SSH certificates.


maraino:
@dopey: Make sure to update dependencies if you are going to merge this PR

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+)
+
+func certificateCommand() cli.Command {
+	return cli.Command{
+		Name:   "certificate",
+		Action: command.ActionFunc(certificateAction),
+		Usage:  "sign a SSH certificate using the the SSH CA",
+		UsageText: `**step ssh certificate** <key-id> <key-file>
+		[**--host**] [**--sign**]`,
```
dopey: All all the other flags

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+$ step ssh certificate --host internal.example.com ssh_host_ecdsa_key
+'''
+
+Sign an SSH public key and generate a host certificate:
+'''
+$ step ssh certificate --host --sign \
+	internal.example.com ssh_host_ecdsa_key.pub
+'''
+
+Generate a new key pair, and a certificate with custom principals (user/host names):
```
dopey: won't this sign an existing public key?

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+'''
+$ step ssh certificate --principal max --principal mariano --sign \
+	ops@work id_ecdsa
+'''
+
+Sign an SSH public key generating a certificate with given token:
```
dopey: Won't this generate a new key pair?

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+	caClient, err := flow.GetClient(ctx, subject, token)
+	if err != nil {
+		return err
+	}
+
+	var sshPub ssh.PublicKey
+	var pub, priv interface{}
+
+	if isSign {
+		// Used given public key
```
dopey: use

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+		}
+	}
+
+	caClient, err := flow.GetClient(ctx, subject, token)
+	if err != nil {
+		return err
+	}
+
+	var sshPub ssh.PublicKey
+	var pub, priv interface{}
```
dopey: don't need to define these at this level

dopey: nvm, I guess you're using `priv` later.

command/ssh/certificate.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+
+	if !isSign {
+		ui.PrintSelected("Private Key", keyFile)
+		ui.PrintSelected("Public Key", pubFile)
+	}
+	ui.PrintSelected("Certificate", crtFile)
+
+	// Attempt to add key to agent
+	if err := sshAddKeyToAgent(subject, resp.Certificate.Certificate, priv); err != nil {
```
dopey: Isn't it possible that `priv` is `nil` here? If `isSign` then we never set `priv`.

crypto/pki/pki.go
```diff
@@ -178,6 +185,18 @@ func New(public, private, config string) (*PKI, error) {
 	if p.intermediateKey, err = getPath(private, "intermediate_ca_key"); err != nil {
 		return nil, err
 	}
+	if p.sshHostCert, err = getPath(public, "ssh_host_key.pub"); err != nil {
```
dopey: wondering about consistency here. in `step ssh certificate` you use `-cert.pub` to denote certificate files. Why not do the same here? Otherwise I assume it's just a pub key.

crypto/pki/pki.go
```diff
@@ -272,6 +291,35 @@ func (p *PKI) GenerateIntermediateCertificate(name string, rootCrt *x509.Certifi
...
+		}
+		sshKey, err := ssh.NewPublicKey(pub)
+		if err != nil {
+			return errors.Wrapf(err, "error converting public key")
+		}
+		_, err = pemutil.Serialize(priv, pemutil.WithFilename(privNames[i]), pemutil.WithPassword(password))
+		if err != nil {
+			return err
+		}
+		if err = utils.WriteFile(pubNames[i], ssh.MarshalAuthorizedKey(sshKey), 0600); err != nil {
```
dopey: a bit confusing because here you're writing a public key but the variable name you're using is `...cert`. Let's just be consistent, unless there is a good reason not to be.

crypto/pki/pki.go
```diff
@@ -299,6 +347,12 @@ func (p *PKI) tellPKI() {
 	ui.PrintSelected("Root fingerprint", p.rootFingerprint)
 	ui.PrintSelected("Intermediate certificate", p.intermediate)
 	ui.PrintSelected("Intermediate private key", p.intermediateKey)
+	if p.enableSSH {
+		ui.PrintSelected("SSH user certificate", p.sshUserCert)
```
dopey: consider saying `SSH user CA ...  & SSH host CA ...`. Otherwise it may not be clear that these are for use by the CA (for signing).

flags/flags.go
```diff
@@ -52,38 +57,102 @@ unset, default is P-256 for EC keys and Ed25519 for OKP keys.
...
-	Usage: `Do not ask for a password to encrypt a private key. Sensitive key material will
+	// NoPassword is a cli.Flag used to avoid using a password to encrypt private
+	// keys.
+	NoPassword = cli.BoolFlag{
+		Name: "no-password",
+		Usage: `Do not ask for a password to encrypt a private key. Sensitive key material will
 be written to disk unencrypted. This is not recommended. Requires **--insecure** flag.`,
-}
+	}
+
```
dopey: I assume we should be switching to using these global flags, rather than the private ones defined in /ca and /certificate and /ssh.

command/ssh/certificate.go
maraino: Yes, the commands are reversed.

command/ssh/certificate.go
maraino: Yes, the commands are reversed.

command/ssh/certificate.go
dopey: The answer to that question is 'yes'. As a fix I've added another flag (--private-key) so that you can still specify the private key for adding to ssh agent. If `priv == nil` then we won't attempt to add to the ssh agent.

flags/flags.go
dopey: I've switched to using the global flags where possible

crypto/pki/pki.go
dopey: Gonna leave it as is for now because I don't wanna break stuff, but this is still an open question for me. I think the file names should be "..-cert.pub" since that's how I understand these keys will be used.

maraino: They are public keys not certificates, renaming the variable to something like sshHostPubKey and sshUserPubKey might be clearer.

PR #120, opened 2019-07-26
Added configurable callback listener

jibingeo:
### Description
Not every identity provider allows a dynamic port number in the allowed callback URL. So this helps to configure static port number for the callback listener. 

#### Usage
`step oauth --listen 127.0.0.1:8080 ...`

💔Thank you!

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -9,6 +9,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -129,6 +130,11 @@ func init() {
 				Name:  "jwt",
 				Usage: "Generate a JWT Auth token instead of an OAuth Token (only works with service accounts)",
 			},
+			cli.StringFlag{
+				Name:  "listen",
+				Value: "127.0.0.1:8080",
+				Usage: "Callback listener URL (default: 127.0.0.1:8080)",
+			},
 			cli.BoolFlag{
 				Name:   "implicit",
 				Usage:  "Uses the implicit flow to authenticate the user. Requires **--insecure** and **--client-id** flags.",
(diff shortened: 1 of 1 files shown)


command/oauth/cmd.go
```diff
@@ -404,11 +415,26 @@ func disco(provider string) (map[string]interface{}, error) {
 	return details, err
 }
 
+func (o *oauth) NewServer() (*httptest.Server, error) {
+	l, err := net.Listen("tcp", o.CallbackListener)
+	if err != nil {
+		return nil, err
```
maraino: Can you wrap using `errors.Wrap(f)` errors coming from the standard packages. This help us to get traces from errors.

command/oauth/cmd.go
```diff
@@ -404,11 +415,26 @@ func disco(provider string) (map[string]interface{}, error) {
...
+	}
+	srv := httptest.NewUnstartedServer(o)
+	srv.Listener.Close()
+	srv.Listener = l
+	srv.Start()
+	return srv, nil
+}
```
maraino: After removing the default value you should change this method. The easiest thing would be to use `httptest.NewServer` if `o.CallbackListerer` is `""`, this will also take care of weird configurations where tcp6 is the only stack available.

And with a custom address, instead of opening an extra listener because of `httptest.NewUnstartedServer` you can just use define the httptest.Server:
```go
srv := &httptest.Server{
	Listener: l,
	Config:   &http.Server{Handler: o},
}
```

jibingeo:
I am using  https://auth0.com

jibingeo:
Not needed. I will wait for the next release. 

Thanks

mmalone:
@jibingeo thank you for your contribution

If you have any other ideas / feedback you wanna talk about drop us a line [on gitter](https://gitter.im/smallstep/community) or shoot me an email (mike @ [our domain](https://smallstep.com/)). Understanding use cases is super valuable for us right now so if you're able to share anything -- even a couple sentences -- about what you're up to we'd love to hear about it.

Keep on rockin'! 💕

PR #128, opened 2019-08-15
Seb/new website frontmatter

sourishkrout:
### Description
Auto-gen'd reference docs will move from `jekyll` to `hugo.js` which requires additional frontmatter (backwards compatible) & different files names (breaking change; behind flag).

### DIFF
--- a/usage/help.go
+++ b/usage/help.go
@@ -41,6 +41,10 @@ func HelpCommand() cli.Command {
 				Name:  "report",
 				Usage: "Writes a JSON report to the HTML docs directory.",
 			},
+			cli.BoolFlag{
+				Name:  "hugo",
+				Usage: "Writes hugo.js (vs jekyll) compatible markdown files",
+			},
 		},
 	}
 }
--- a/usage/help.go
+++ b/usage/help.go
@@ -43,7 +43,7 @@ func HelpCommand() cli.Command {
 			},
 			cli.BoolFlag{
 				Name:  "hugo",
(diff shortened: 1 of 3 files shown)


usage/html.go
```diff
@@ -36,46 +36,57 @@ func markdownHelpAction(ctx *cli.Context) error {
...
 	}
-	markdownHelpPrinter(w, mdAppHelpTemplate, ctx.App)
+	markdownHelpPrinter(w, mdAppHelpTemplate, "", ctx.App)
 	if err := w.Close(); err != nil {
 		return errs.FileError(err, index)
 	}
 
+	// preserve jekyll compatibility for transition period
+	fileName := "index.md"
+	if ctx.IsSet("hugo") {
```
maraino: This should be `ctx.Bool("hugo")`, because `--hugo=false` will set hugo, but it will be false.

sourishkrout: done

usage/help.go
```diff
@@ -41,6 +41,10 @@ func HelpCommand() cli.Command {
 				Name:  "report",
 				Usage: "Writes a JSON report to the HTML docs directory.",
 			},
+			cli.BoolFlag{
+				Name:  "hugo",
+				Usage: "Writes hugo.js (vs jekyll) compatible markdown files",
```
alanctkc: Small nit.. `hugo` instead of `hugo.js` (a Go tool)

PR #129, opened 2019-08-20
Minor fix: Hugo wants underscores in dirs with subdirs only

sourishkrout:
### Description
Misunderstood the requirements for underscores in `index.md` file names. Only `_index.md` for directories that have no sub-directories.


sourishkrout:
The build's failing with an unrelated error which looks like dependencies are failing to download. 🤷‍♂

PR #130, opened 2019-08-22
Max/lint

dopey:
gometalinter has been deprecated. 

Switch to golangci-lint, simplify `make lint`, fix lots of linting errors (mostly `err` shadow variables).


command/ca/offline.go
```diff
@@ -231,7 +231,8 @@ func (c *offlineCA) GenerateToken(ctx *cli.Context, typ int, subject string, san
 
 	switch p := p.(type) {
 	case *provisioner.OIDC: // Run step oauth
-		out, err := exec.Step("oauth", "--oidc", "--bare",
+		var out []byte
+		out, err = exec.Step("oauth", "--oidc", "--bare",
```
maraino: I would prefer to configure the linter so it doesn't force you to do this change.

PR #151, opened 2019-09-16
Enable SSH

### LINKED ISSUE #149: When initializing the CA with --ssh the global claims in the ca.json should set `enableSSHCA: true`
### Provisioners are not initially enabled for SSH even after init-ing the CA using `--ssh`
After initializing `step-ca` with `--ssh` I would expect that the default provisioner is enabled for ssh. Otherwise the experience is very confusing. The user is told they can't create ssh certificates, even though they just initialized the CA using the `--ssh` option.

### Expected behaviour
SSH certificates should work out of the box without extra user intervention in the ca.json (which is currently required).

### Additional context
Consider setting the global claims attribute to `true` (if `--ssh`) and letting new provisioners set it to false when necessary.

maraino:
### Description
Enable SSH by default if `step ca init --ssh` is  used.

Fixes #149

### DIFF
--- a/command/ca/provisioner/add.go
+++ b/command/ca/provisioner/add.go
@@ -135,6 +135,10 @@ By default it will accept any SAN in the CSR.`,
 with the same instance will be accepted. By default only the first request
 will be accepted.`,
 			},
+			cli.BoolFlag{
+				Name:  "ssh",
+				Usage: `Enable SSH on the new providers.`,
+			},
 		},
 		Description: `**step ca provisioner add** adds one or more provisioners
 to the configuration and writes the new configuration back to the CA config.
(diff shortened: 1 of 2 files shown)


maraino:
@dopey yes, there's a global enableSSH, but as this type of access can be more sensitive we decided to not enable it by default, and I only enabled it for the default provisioner on the init.
But we can decide to enable it globally.

command/ca/provisioner/add.go
```diff
@@ -135,6 +135,10 @@ By default it will accept any SAN in the CSR.`,
 with the same instance will be accepted. By default only the first request
 will be accepted.`,
 			},
+			cli.BoolFlag{
+				Name:  "ssh",
+				Usage: `Enable SSH on the new providers.`,
```
dopey: Enable SSH on the new provisioner?

PR #162, opened 2019-09-25
Add x5c provisioner capabilities

dopey:
Certificates PR: https://github.com/smallstep/certificates/pull/114


utils/cautils/token_flow.go
```diff
@@ -110,6 +111,41 @@ func NewTokenFlow(ctx *cli.Context, typ int, subject string, sans []string, caUR
...
+			return tokenGen.RevokeToken(subject, token.WithX5CFile(x5cCertFile, jwk.Key))
+		case SSHUserSignType:
+			return tokenGen.SignSSHToken(subject, provisioner.SSHUserCert, sans,
+				certNotBefore, certNotAfter, token.WithX5CFile(x5cCertFile, jwk.Key))
+		case SSHHostSignType:
+			return tokenGen.SignSSHToken(subject, provisioner.SSHHostCert, sans,
+				certNotBefore, certNotAfter, token.WithX5CFile(x5cCertFile, jwk.Key))
+		default:
+			return tokenGen.Token(subject, token.WithX5CFile(x5cCertFile, jwk.Key))
+		}
```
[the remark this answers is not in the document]
maraino: It's enough code now that we might want to refactor a bit.

utils/cautils/token_flow.go
```diff
@@ -266,18 +302,33 @@ func OfflineTokenFlow(ctx *cli.Context, typ int, subject string, sans []string,
...
 }
 
+func allowX5CProvisionerFilter(p provisioner.Interface) bool {
+	switch p.GetType() {
+	case provisioner.TypeX5C:
+		return true
+	default:
+		return false
+	}
+}
```
maraino: In one line `return p.GetType() == provisioner.TypeX5C`

token/options.go
```diff
@@ -157,3 +160,32 @@ func WithKid(s string) Options {
...
+		}
+
+		if err = x509util.VerifyCertKey(certs[0], key); err != nil {
+			return errors.Wrap(err, "x5c")
+		}
+
+		if certs[0].KeyUsage&x509.KeyUsageDigitalSignature == 0 {
+			return errors.New("x5c: certificate/private-key pair used to sign " +
+				"x5c token cannot be used for digital signature")
+		}
```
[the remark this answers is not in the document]
dopey: For posterity (since we already had a similar discussion in certificates), Verify cannot check KeyUsage extensions (only ExtKeyUsage). I will however change this to check every certificate in the chain.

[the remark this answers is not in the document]
dopey: Check that, I think only the leaf certificate should require the digital signature extension.

command/ca/ca.go
```diff
@@ -85,27 +85,63 @@ $ step ca renew internal.crt internal.key \
...
+Use this flag to define an ACME server other than the Step CA. If this flag is
+absent and an ACME provisioner has been selected then the '--ca-url' flag must be defined.`,
+	}
+
+	acmeContactFlag = cli.StringSliceFlag{
+		Name: "contact",
+		Usage: `<email-addresses> for contact as part of the ACME protocol. These contacts
+may be used to warn of certificate expration or other certificate lifetime events.
+Use the '--contact' flag multiple times to configure multiple contacts.`,
+	}
```
maraino: Perhaps `The <email-address> ...` so it displays `--contact=email-address` instead of `--contact=email-addresses`

command/ca/provisioner/add.go
```diff
@@ -107,38 +106,50 @@ Use the flag multiple times to configure multiple service accounts.`,
...
+    **Azure**
+    : Uses Microsoft Azure identity tokens.
+
+    **ACME**
+    : Uses the ACME protocol to create certificates.
+
+    **X5C**
+    : Uses an X509 Certificate / private key pair to sign provisioning tokens.
+`,
+			},
```
maraino: type should be one of the first flag to appear if not the first. The order in this list is important.

maraino: Make sure related flags are next to each other.

command/ca/sign.go
```diff
@@ -82,6 +83,12 @@ $ step ca sign foo.csr foo.crt \
 --acme https://acme-staging-v02.api.letsencrypt.org/directory
 '''`,
 		Flags: []cli.Flag{
+			acmeFlag,
+			acmeContactFlag,
+			acmeHTTPListenFlag,
+			acmeStandaloneFlag,
+			acmeWebrootFlag,
+			consoleFlag,
```
maraino: I don't think acme and console flags are the most important, move them down.

command/crypto/jws/sign.go
```diff
@@ -30,6 +32,8 @@ be written to STDOUT.
 
 For examples, see **step help crypto jws**.`,
 		Flags: []cli.Flag{
+			flags.X5cCert,
+			flags.X5cKey,
```
maraino: Order is important, move them down.

command/crypto/jwt/sign.go
```diff
@@ -57,6 +59,8 @@ A JWT signed using JWS has three parts:
 
 For examples, see **step help crypto jwt**.`,
 		Flags: []cli.Flag{
+			flags.X5cCert,
+			flags.X5cKey,
```
maraino: Move them down.

maraino: Reuse `--key`? Whatever looks better.

command/crypto/jws/sign.go
maraino: Reuse `--key`? Whatever looks better.

command/crypto/jwt/sign.go
```diff
@@ -340,6 +360,19 @@ func signAction(ctx *cli.Context) error {
...
+		certs, err := pemutil.ReadCertificateBundle(x5c)
+		if err != nil {
+			return errors.Wrap(err, "error reading cert file for x5c header")
+		}
+		strs := make([]string, len(certs))
+		for i, cert := range certs {
+			strs[i] = base64.StdEncoding.EncodeToString(cert.Raw)
+		}
+		so.WithHeader("x5c", strs)
+	}
```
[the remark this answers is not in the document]
maraino: Do we?

crypto/x509util/crt.go
```diff
@@ -21,6 +25,39 @@ func Fingerprint(cert *x509.Certificate) string {
...
+	case ed25519.PublicKey:
+		priv, ok := key.(ed25519.PrivateKey)
+		if !ok {
+			return errors.New("private key type does not match public key type")
+		}
+		if !bytes.Equal(priv.Public().(ed25519.PublicKey), pub) {
+			return errors.New("private key does not match public key")
+		}
+	default:
+		return errors.New("unknown public key algorithm")
```
maraino: unsupported instead of unknown and perhaps write the type (`%T`)

flags/flags.go
```diff
@@ -152,6 +152,17 @@ but can accept a different configuration file using '--ca-config>' flag.`,
 $STEPPATH/config/ca.json`,
 		Value: filepath.Join(config.StepPath(), "config", "ca.json"),
 	}
+
+	X5cCert = cli.StringFlag{
+		Name:  "x5c-cert",
+		Usage: "Certificate (<chain>) to store in the 'x5c' header of a JWT.",
```
maraino: ... in PEM format.

utils/cautils/token_generator.go
```diff
@@ -77,24 +86,187 @@ func (t *TokenGenerator) Token(sub string, opts ...token.Options) (string, error
...
+	}
+	if p.ListenAddress != "" {
+		args = append(args, "--listen", p.ListenAddress)
+	}
+	out, err := exec.Step(args...)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(out)), nil
+}
```
maraino: 👏 for refactoring this

utils/cautils/token_generator.go
```diff
@@ -77,24 +86,187 @@ func (t *TokenGenerator) Token(sub string, opts ...token.Options) (string, error
...
+// loadJWK loads a JWK from a combination of the
+func loadJWK(ctx *cli.Context, p *provisioner.JWK, tokAttrs tokenAttrs) (jwk *jose.JSONWebKey, kid string, err error) {
+	var opts []jose.Option
+	if passwordFile := ctx.String("password-file"); len(passwordFile) != 0 {
+		opts = append(opts, jose.WithPasswordFile(passwordFile))
+	}
+
+	if keyFile := ctx.String("key"); len(keyFile) == 0 {
+		if p == nil {
+			return nil, "", errors.New("offline mode with no CA configuration file must specify a JWK file (--key)")
```
maraino: I think the offline mode always requires a ca.json.

maraino: And it looks like we're erroring with this message before checking if it is offline.

utils/cautils/token_generator.go
```diff
@@ -77,24 +86,187 @@ func (t *TokenGenerator) Token(sub string, opts ...token.Options) (string, error
...
+		}
+		kid = p.Key.KeyID
+		// If provisioner is not nil then we must be using the offlineCA.
+		var encryptedKey string
+		if ctx.IsSet("offline") {
```
maraino: It should be `ctx.Bool("offline")`, even if it's not common `--offline=false` is possible

command/crypto/jwt/sign.go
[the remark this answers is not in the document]
dopey: It's being done in `jose.ValidateX5C(...)`.

command/ca/provisioner/add.go
dopey: Why is it important that type be one of the first flags? I sorted the flags alphabetically because I think it was pretty chaotic before that. We have hundreds of lines of flag definitions and it was hard to know if the flag you cared about was already defined. 

Is there any reason to keep related flags next to one another (other than that they are functionally related)? We should discuss coming up with a consistent pattern, because from my perspective it was chaos.

command/ca/sign.go
dopey: So, I didn't realize this, but it seems the order here is reflected as the order in the help doc. That's rather unfortunate, as it makes the flag array pretty difficult to reason about, given our mix of global, local, and in the moment defined flags. I down for updating this, but it would be great if we came up with something consistent.

command/crypto/jws/sign.go
dopey: I can make it "key, x5c-key", but I would like to have `x5c-key` available since we're using that in `step ca certificate`.

command/crypto/jwt/sign.go
dopey: optically I think it's nice to have `x5c-key` available, but as you said I can add it to the definition of the `key` flag.

command/ca/sign.go
dopey: moved these around to a more appropriate order

utils/cautils/token_generator.go
dopey: changed the error and added some documentation for the function, since there's a lot of conditionals and pathways to take in there.

command/crypto/jwt/sign.go
```diff
@@ -226,48 +225,28 @@ func signAction(ctx *cli.Context) error {
...
-	if len(x5cCertFile) > 0 || len(x5cKeyFile) > 0 {
-		if len(x5cCertFile) == 0 {
-			return errs.RequiredWithProvisionerTypeFlag(ctx, "X5C", "x5c-cert")
-		}
+	if len(x5cCertFile) > 0 {
 		if len(x5cKeyFile) == 0 {
-			return errs.RequiredWithProvisionerTypeFlag(ctx, "X5C", "x5c-key")
+			if x5cKeyFile = key; len(x5cKeyFile) == 0 {
+				return errs.RequiredWithOrFlag(ctx, "x5c-cert", "key", "x5c-key")
+			}
```
maraino: This is weird, x5c-key and key are aliases now, is `urfave/cli` not setting both?

command/crypto/jwt/sign.go
```diff
@@ -226,48 +225,28 @@ func signAction(ctx *cli.Context) error {
...
 		}
-		if jwks != "" {
-			return errs.MutuallyExclusiveFlags(ctx, "x5c-key", "jwks")
+		if len(jwks) > 0 {
+			return errs.MutuallyExclusiveFlags(ctx, "x5c-cert", "jwks")
 		}
+		isX5C = true
 	} else {
+		if len(x5cKeyFile) > 0 {
+			return errs.RequiredWithFlag(ctx, "x5c-key", "x5c-cert")
```
maraino: We can also use `errs.RequiredWithOrFlag`

command/crypto/jwt/sign.go
dopey: You are right. I didn't expect that. Just removed the second clause and bumped it up.

command/crypto/jwt/sign.go
dopey: completely removed this clause. If you define `key` then `x5c-key` will also be defined because they are aliases.

PR #177, opened 2019-10-17
Fix typo on install url for debian

Carles-Figuerola:
### Description

This is just a simple fix for the url for the debian package, the version has a `v` in front of it for the branch but not the package filename itself


maraino:
Gràcies @Carles-Figuerola

Carles-Figuerola:
De res @maraino, gràcies per la rapidesa :)

PR #179, opened 2019-10-18
Update cli to use the newly added CertChainPEM in the step-ca API

dopey:
* update offline and online tokens to use the new API.
* the old API is still accessible, but it is deprecated and we'll be
looking to push all new clients to use the new API.

### DIFF
--- a/utils/cautils/certificate_flow.go
+++ b/utils/cautils/certificate_flow.go
@@ -199,15 +199,17 @@ func (f *CertificateFlow) Sign(ctx *cli.Context, token string, csr api.Certifica
 		return err
 	}
 
-	serverBlock, err := pemutil.Serialize(resp.ServerPEM.Certificate)
-	if err != nil {
-		return err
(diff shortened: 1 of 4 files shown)


utils/cautils/certificate_flow.go
```diff
@@ -199,15 +199,17 @@ func (f *CertificateFlow) Sign(ctx *cli.Context, token string, csr api.Certifica
...
-	caBlock, err := pemutil.Serialize(resp.CaPEM.Certificate)
-	if err != nil {
-		return err
+	var data []byte
+	for _, certPEM := range resp.CertChainPEM {
+		pemblk, err := pemutil.Serialize(certPEM)
+		if err != nil {
+			return errors.Wrap(err, "error serializing from step-ca API response")
+		}
+		data = append(data, pem.EncodeToMemory(pemblk)...)
```
maraino: This looks ok for CertChainPEM, but if it's empty you should create the certificate using ServerPEM and CaPEM.

PR #181, opened 2019-10-29
Kubernetes service account tokens 

dopey:


command/ca/certificate.go
```diff
@@ -139,6 +140,11 @@ $ step ca certificate foo.internal foo.crt foo.key \
 			acmeWebrootFlag,
 			acmeContactFlag,
 			acmeHTTPListenFlag,
+			cli.StringFlag{
+				Name: "k8ssa-token-path",
+				Usage: `Configure the <file> from which to read the kubernetes service account token.
+By default this location will be /var/run/secrets/kubernetes.io/serviceaccount/token.`,
+			},
```
maraino: We should add the default value here using the `cli.StringFlag` property `Value`.

command/ca/provisioner/add.go
```diff
@@ -252,6 +268,11 @@ $ step ca provisioner add acme-smallstep --type ACME --ca-config ca.json
 Add an X5C provisioner.
 '''
 $ step ca provisioner add x5c-smallstep --type X5C --x5c-root x5cRoot.crt
+'''
+
+Add a K8s Service Account provisioner.
+'''
+$ step ca provisioner add my-kube-provisioner --type K8sSA --pem-keys keys.pub
```
maraino: Can we add to the docs how to get the keys?

command/ca/provisioner/add.go
```diff
@@ -588,6 +611,73 @@ func addX5CProvisioner(ctx *cli.Context, name string, provMap map[string]bool) (
...
+	}
+
+	var pubKeyBytes []byte
+	for _, k := range pemKeys {
+		blk, err := pemutil.Serialize(k)
+		if err != nil {
+			return nil, errors.Wrap(err, "error serializing pem key")
+		}
+		pubKeyBytes = append(pubKeyBytes, pem.EncodeToMemory(blk)...)
+	}
```
maraino: Is there any point in deserializing and serializing the keys?

maraino: In most cases, the serialization will be equivalent to `pemKeysB`.

command/ca/provisioner/add.go
```diff
@@ -588,6 +611,73 @@ func addX5CProvisioner(ctx *cli.Context, name string, provMap map[string]bool) (
...
+	p := &provisioner.K8sSA{
+		Type:    provisioner.TypeK8sSA.String(),
+		Name:    name,
+		Claims:  getClaims(ctx),
```
maraino: There's no SSH support in the CA for service accounts.

command/ca/certificate.go
maraino: This is also common for `step ca certificate/token/sign`, unify them in a variable.

command/ca/sign.go
```diff
@@ -106,6 +107,11 @@ $ step ca sign foo.csr foo.crt \
 			acmeWebrootFlag,
 			acmeContactFlag,
 			acmeHTTPListenFlag,
+			cli.StringFlag{
+				Name: "k8ssa-token-path",
+				Usage: `Configure the <file> from which to read the kubernetes service account token.
+By default this location will be /var/run/secrets/kubernetes.io/serviceaccount/token.`,
+			},
```
maraino: Use a unique var for this flag in different files. Set default in `Value`.

command/ca/token.go
```diff
@@ -171,6 +172,11 @@ be invalid for any other API request.`,
 				Name:  "ssh",
 				Usage: `Create a token for authorizing an SSH certificate signing request.`,
 			},
+			cli.StringFlag{
+				Name: "k8ssa-token-path",
+				Usage: `Configure the <file> from which to read the kubernetes service account token.
+By default this location will be /var/run/secrets/kubernetes.io/serviceaccount/token.`,
+			},
```
maraino: Use a unique var for this flag in different files. Set default in `Value`.

go.mod
```diff
@@ -3,51 +3,34 @@ module github.com/smallstep/cli
...
-	golang.org/x/net v0.0.0-20181029044818-c44066c5c816
-	golang.org/x/sys v0.0.0-20181031143558-9b800f95dbbc // indirect
-	google.golang.org/appengine v1.5.0 // indirect
+	golang.org/x/crypto v0.0.0-20190313024323-a1f597ede03a
+	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	gopkg.in/square/go-jose.v2 v2.4.0
 	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect
 )
+
+replace github.com/smallstep/certificates => ../certificates
```
maraino: Note that this should go away before merging.

command/ca/provisioner/add.go
dopey: Gives us the opportunity to fail when you're adding the provisioner, rather than failing when you attempt to sighup/restart the CA. I'd say it's nice to have slightly more assurance that the CA won't error out on restart.

command/ca/provisioner/add.go
dopey: I have no idea how to get these keys. I don't think they're accessible w/out using the TokenReview API. I think mike's intention of offering this as v1 was to just have "something" out there that we could iterate on quickly for a v2.

command/ca/provisioner/add.go
dopey: I assumed the `getClaims` method did more than just setting up `ssh` claims. If more is added to that method over time, it might be good to have it here by default. But I can also remove it if you'd prefer.

command/ca/provisioner/add.go
maraino: sounds good

command/ca/provisioner/add.go
maraino: You can return an error is --ssh and --type == K8sSA.

command/ca/provisioner/add.go
maraino: sounds good.

PR #188, opened 2019-11-19
ssh certificate nil principals for OIDC

dopey:
* only when there are no explicit principals requested.

### DIFF
--- a/command/ssh/certificate.go
+++ b/command/ssh/certificate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/smallstep/cli/crypto/sshutil"
 	"github.com/smallstep/cli/errs"
 	"github.com/smallstep/cli/flags"
+	"github.com/smallstep/cli/jose"
 	"github.com/smallstep/cli/ui"
 	"github.com/smallstep/cli/utils"
 	"github.com/smallstep/cli/utils/cautils"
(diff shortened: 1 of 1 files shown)


command/ssh/certificate.go
```diff
@@ -285,6 +287,24 @@ func certificateAction(ctx *cli.Context) error {
 		sshAuPubBytes = sshAuPub.Marshal()
 	}
 
+	// NOTE: For OIDC token the principals should be completely empty unless
+	// defined explicitly on the command line using the `--principal` flag.
+	// The OIDC provisioner is responsible for setting default principals by
+	// using an identity function.
+	j, err := jwt.ParseSigned(token)
```
maraino: use `jose.ParseSigned(token)`

command/ssh/certificate.go
```diff
@@ -285,6 +287,24 @@ func certificateAction(ctx *cli.Context) error {
...
+	if err != nil {
+		return errors.Wrap(err, "error parsing token")
+	}
+	// The ID will be just the clientID stored in azp.
+	var payload tokenPayload
+	if err := j.UnsafeClaimsWithoutVerification(&payload); err != nil {
+		return errors.Wrap(err, "err parsing token claims")
+	}
+	// If it's an OIDC token and and no principals were explicitly set ...
+	if len(payload.AuthorizedParty) > 0 && !ctx.IsSet("principal") {
```
maraino: It's probably better to do this for any token with an `email` in the payload.

PR #192, opened 2019-12-10
Expired cert

dopey:


command/ssh/checkHost.go
```diff
@@ -61,7 +65,27 @@ func checkHostAction(ctx *cli.Context) error {
 		return err
 	}
 
-	resp, err := client.SSHCheckHost(ctx.Args().First())
+	id, err := ca.LoadDefaultIdentity()
+	if err != nil {
+		return errors.Wrap(err, "error loading the deault x5c identity")
```
maraino: we should probably return something add something like `please run 'step ssh login <identity>'`

go.mod
```diff
@@ -24,6 +26,7 @@ require (
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	golang.org/x/sys v0.0.0-20191008105621-543471e840be
 	gopkg.in/square/go-jose.v2 v2.4.0
+	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect
 )
 
-// replace github.com/smallstep/certificates => ../certificates
+replace github.com/smallstep/certificates => ../certificates
```
maraino: Make sure to use a commit instead of this.

token/options.go
```diff
@@ -165,7 +165,24 @@ func WithX5CFile(certFile string, key interface{}) Options {
 		if err != nil {
 			return errors.Wrap(err, "error validating x5c certificate chain and key for use in x5c header")
 		}
-		c.SetHeader("x5c", certStrs)
+		c.SetHeader("x4c", certStrs)
```
maraino: `x4c`? `x5c`?

token/options.go
```diff
@@ -165,7 +165,24 @@ func WithX5CFile(certFile string, key interface{}) Options {
...
+// a use case where the user would prefer not to validate the certificate chain
+// before returning it. Presumably the user would then perform their own validation.
+// NOTE: here be dragons. Use WithX5CFile unless you know what you are doing.
+func WithX5CAllowInvalidFile(certFile string, key interface{}) Options {
+	return func(c *Claims) error {
+		certStrs, err := jose.ValidateX5C(certFile, key)
+		if err != nil {
+			return errors.Wrap(err, "error validating x5c certificate chain and key for use in x5c header")
+		}
+		c.SetHeader("x5cAllowInvalid", certStrs)
```
maraino: Perhaps would be better to use x5c and a custom bool header to indicate that might be invalid, perhaps something with the word "insecure".

dopey: I tried to explain in the comment why I couldn't just use `x5c` as the header. It's basically a special word in their implementation and there's no way to pull the certificate chain from the token without also validating it. That's a problem since we're specifically building this for the case of invalid certificates. Basically I have to use a keyword that is not `x5c` as the key.

command/ssh/checkHost.go
```diff
@@ -61,7 +65,27 @@ func checkHostAction(ctx *cli.Context) error {
 		return err
 	}
 
-	resp, err := client.SSHCheckHost(ctx.Args().First())
+	id, err := ca.LoadDefaultIdentity()
```
maraino: A user won't have an identity certificate unless the CA is configured with mTLS support, so perhaps we should check that first, `client.Version()` will return a struct with the `RequireClientAuthentication` field. We should only add the token if this is required.

token/options.go
maraino: Can we change it to x5cInsecure ?

dopey: changed to `x5cInsecure`

command/ssh/checkHost.go
```diff
@@ -60,8 +64,34 @@ func checkHostAction(ctx *cli.Context) error {
...
+				"/ssh/check-host", "", time.Time{}, time.Time{}, jwk)
+			tok, err = tokenGen.Token("sub", token.WithX5CInsecureFile(id.Certificate, jwk.Key))
+			if err != nil {
+				return errors.Wrap(err, "error generating identity x5c token for /ssh/check-host request")
+			}
+		}
+	}
 
-	resp, err := client.SSHCheckHost(ctx.Args().First())
+	resp, err := client.SSHCheckHost(ctx.Args().First(), tok)
```
maraino: Do `hostname := ctx.Args().First()` before so it's more clear

command/ssh/checkHost.go
```diff
@@ -60,8 +64,34 @@ func checkHostAction(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	version, err := client.Version()
```
maraino: 👍 We need to remove the retry function above.

jose/parse.go
```diff
@@ -225,6 +227,87 @@ func ParseKeySet(filename string, opts ...Option) (*jose.JSONWebKey, error) {
...
+		if err != nil {
+			return nil, errors.Wrap(err, "error parsing x5c cert")
+		}
+		certs[i] = cert
+	}
+	return certs, nil
+}
+
+// X5cInsecureHdrKey is the key used to store the x5cInsecure cert chain in the JWT header.
+var X5cInsecureHdrKey = "x5cInsecure"
```
maraino: We should try to use nicer names, `X5cInsecureKey` looks nicer to me.

PR #197, opened 2019-12-23
Allow kty, crv, and size key mods for ACME

dopey:


dopey:
Gotcha. Just took another look at it. Still not entirely sure why it's implemented that way, but it seems this PR still works so 🤷‍♂. Appreciate you calling that out though.

PR #198, opened 2020-01-07
Remove custom x509 package

maraino:
### Description
This PR removes the use of the custom x509 package and uses Go 1.13+ x509 as it supports Ed25519.


integration/command.go
```diff
@@ -59,6 +59,7 @@ type CLICommand struct {
 }
 
 // CLIOutput represents the output from executing a CLICommand.
+// nolint:unused
```
dcow: Why not remove this?

command/ssh/proxy.go
```diff
@@ -154,6 +158,7 @@ func makeAddress(host, port string) string {
 	}
 }
 
+// nolint:unused
```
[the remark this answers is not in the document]
maraino: This was part of an experiment of an ssh client support in step, right now it's disabled, but we might want to add support for it in the future, we can certainly remove it as it will be in git, but I kind of resolve it on a different way.

I'll remove it, as I don't think it's something that we're going to use soon.

integration/command.go
maraino: This is part of a framework Mike and I created to be able to do integration tests, it was a little hard to do and run those tests consistently especially because of the way the input was managed. It didn't go far, but it's something that we can improve. I've just removed the linting error.

PR #199, opened 2020-01-10
oauth: Add terminal redirect URL option

dcow:
Add a `--redirect-url` string option to the boostrap command that allows
the terminal redirect in the OAuth flow to be configured. This is useful
if you want to display a more interesting page than the locally hosted
basebones "Success" page.

If the user configures step-cli using the `ssh config` command and passes
the `--team` option, set the redirect url to:

    https://smallstep.com/app/teams/sso/success

so that the flow ends on a page that instructs users on what to do next.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -207,6 +207,11 @@ be stored in the 'sshpop' header.`,
 		Name:  "team",
 		Usage: "The team <name> used to bootstrap the environment.",
 	}
+
+	RedirectURL = cli.StringFlag{
+		Name:  "redirect-url",
+		Usage: "Terminal OAuth redirect <url>.",
+	}
 )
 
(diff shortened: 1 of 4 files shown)


flags/flags.go
```diff
@@ -207,6 +207,11 @@ be stored in the 'sshpop' header.`,
 		Name:  "team",
 		Usage: "The team <name> used to bootstrap the environment.",
 	}
+
+	RedirectURL = cli.StringFlag{
+		Name:  "redirect-url",
+		Usage: "Terminal OAuth <redirect-url>.",
```
maraino: It will look nicer if the usage was `"Terminal OAuth redirect <url>."` so the flag is displayed as `--redirect-url=url`.

dcow: done

PR #201, opened 2020-01-11
teams: Add --endpoint option

dcow:
Allow users to specify the endpoint that `step ssh config --team` uses to
pull inital configuration. If no endpoint is specified, default to

    https://api.smallstep.com/v1/teams/<name>/cas/default

If the user-specified endpoint contains any "<>" sequences, replace them
with the team name.

### DIFF
--- a/command/ssh/config.go
+++ b/command/ssh/config.go
@@ -31,7 +31,7 @@ func configCommand() cli.Command {
 		[**--team=name] [**--host**] [**--set=<key=value>**]
 		[**--dry-run**] [**--roots**] [**--federation**]
 		[**--force**] [**--ca-url**=<uri>] [**--root**=<file>]
-		[**--offline**] [**--ca-config**=<path>]`,
+		[**--offline**] [**--ca-config**=<path>] [**--team-url=<url>**]`,
 		Description: `**step ssh config** configures SSH to be used with certificates. It also supports
 flags to inspect the root certificates used to sign the certificates.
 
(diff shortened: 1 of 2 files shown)


utils/cautils/teams.go
```diff
@@ -23,16 +22,27 @@ type bootstrapAPIResponse struct {
...
-	if err != nil {
-		return errors.Wrapf(err, "error parsing %s", apiEndpoint)
+	apiEndpoint := ctx.String("endpoint")
+	if apiEndpoint == "" {
+		// Use the default endpoint..
+		u := url.URL{
+			Scheme: "https",
+			Host:   "api.smallstep.com",
+			Path:   "/v1/teams/" + name + "/cas/default",
+		}
```
dcow: Anyone know off hand if URL has `Validate()` func?

command/ssh/config.go
```diff
@@ -73,6 +73,11 @@ $ step ssh config --set User=joe --set Bastion=bastion.example.com
 				Name:  "host",
 				Usage: `Configures a SSH server instead of a client.`,
 			},
+			cli.StringFlag{
+				Name: "endpoint",
```
maraino: I think something more specific would be better `--team-endpoint` or `--team-url`

utils/cautils/teams.go
maraino: Not really, because it can build different kinds of URIs, you can do one after url.Parse()

command/ssh/config.go
maraino: Probably the latter, as you're asking for a url.

PR #216, opened 2020-02-08
Fix local development doc

mkontani:
- remove vtest description. See https://github.com/smallstep/cli/commit/71ee01bbfa698835c9bfac73458a95113181abef
- fix linter desc metalinter -> golangci-lint. See #130

### Description
I tried and found some deprecated guide descriptions as above.

Thank you.


dopey:
Good spots! Thanks @mkontani .

PR #230, opened 2020-03-13
ssh: Add machine-id hash to host certificate

dcow:
Alternatively, allow the user to specify their own UUID. Adding an ID
derived from the machine ID allows us to authorize hosts to access their
own resources by ID. The machine-id is not supposed to be sent around as
a raw UUID. So we HMAC it with an application "secret" and use the first
sixteen bytes of the resulting sha256 sum to as the entropy source when
generating a new "random" UUIDv4.

### DIFF
--- a/command/ssh/certificate.go
+++ b/command/ssh/certificate.go
@@ -3,8 +3,13 @@ package ssh
 import (
 	"bytes"
 	"crypto"
+	"crypto/hmac"
+	"crypto/sha256"
+	"io/ioutil"
+	"net/url"
 	"strings"
 
+	"github.com/google/uuid"
 	"github.com/pkg/errors"
 	"github.com/smallstep/certificates/api"
 	"github.com/smallstep/certificates/authority/provisioner"
@@ -29,7 +34,7 @@ func certificateCommand() cli.Command {
 		Action: command.ActionFunc(certificateAction),
 		Usage:  "sign a SSH certificate using the the SSH CA",
 		UsageText: `**step ssh certificate** <key-id> <key-file>
-[**--host**] [**--sign**] [**--principal**=<string>] [**--password-file**=<path>]
+[**--host**] [--host-id**] [**--sign**] [**--principal**=<string>] [**--password-file**=<path>]
 [**--provisioner-password-file**=<path>] [**--add-user**]
 [**--not-before**=<time|duration>] [**--not-after**=<time|duration>]
 [**--token**=<token>] [**--issuer**=<name>] [**--ca-url**=<uri>]
@@ -139,6 +144,7 @@ $ step ssh certificate --token $TOKEN mariano@work id_ecdsa
 			flags.Token,
 			sshAddUserFlag,
 			sshHostFlag,
+			sshHostIDFlag,
 			sshPasswordFileFlag,
 			sshPrincipalFlag,
 			sshPrivateKeyFlag,
@@ -167,6 +173,7 @@ func certificateAction(ctx *cli.Context) error {
 	// Flags
 	token := ctx.String("token")
 	isHost := ctx.Bool("host")
+	hostID := ctx.String("host-id")
 	isSign := ctx.Bool("sign")
 	isAddUser := ctx.Bool("add-user")
 	principals := ctx.StringSlice("principal")
@@ -254,6 +261,28 @@ func certificateAction(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
+
+		// All host identity certs need a URI SAN to work with our ssh API.
+		if isHost {
+			var u = uuid.Nil
+			if hostID == "" {
+				u, err = deriveMachineID()
+				if err != nil {
+					return err
+				}
+			} else {
+				u, err = uuid.Parse(hostID)
+				if err != nil {
+					return err
+				}
+			}
+			uri, err := url.Parse(u.URN())
+			if err != nil {
+				return err
+			}
+			csr.URIs = append(csr.URIs, uri)
+		}
+
 		identityCSR = *csr
 		identityKey = key
 	}
@@ -428,4 +457,28 @@ func marshalPublicKey(key ssh.PublicKey, subject string) []byte {
 type oidcPayload struct {
 	jose.Claims
 	Email string `json:"email"` // OIDC indicator
+}
+
+func deriveMachineID() (uuid.UUID, error) {
+	// use /etc/machine-id
+	machineID, err := ioutil.ReadFile("/etc/machine-id")
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	// 32 bytes, not secret
+	key := []byte("moon machines mortify more minds")
+	mac := hmac.New(sha256.New, key)
+	mac.Write(machineID)
+	machineHash := mac.Sum(nil)
+
+	// convert to uuid, definitely not thread-safe
+	r := bytes.NewReader(machineHash)
+	uuid.SetRand(r)
+	defer uuid.SetRand(nil)
+	u, err := uuid.NewRandom()
+	if err != nil {
+		return uuid.Nil, err
+	}
+	return u, nil
 }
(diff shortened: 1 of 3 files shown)


command/ssh/certificate.go
```diff
@@ -254,6 +261,28 @@ func certificateAction(ctx *cli.Context) error {
...
+			var u = uuid.Nil
+			if hostID == "" {
+				u, err = deriveMachineID()
+				if err != nil {
+					return err
+				}
+			} else {
+				u, err = uuid.Parse(hostID)
+				if err != nil {
+					return err
```
maraino: Wrap this using the `pkg/errors` package, we are always wrapping errors in the CLI.

command/ssh/certificate.go
```diff
@@ -254,6 +261,28 @@ func certificateAction(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
+
+		// All host identity certs need a URI SAN to work with our ssh API.
+		if isHost {
```
maraino: There's a switch on line 195 that we can use to throw an error if `--host-id` is passed without `--host`

command/ssh/certificate.go
```diff
@@ -254,6 +261,28 @@ func certificateAction(ctx *cli.Context) error {
...
+				}
+			} else {
+				u, err = uuid.Parse(hostID)
+				if err != nil {
+					return err
+				}
+			}
+			uri, err := url.Parse(u.URN())
+			if err != nil {
+				return err
```
maraino: Wrap error.

command/ssh/certificate.go
```diff
@@ -429,3 +458,27 @@ type oidcPayload struct {
...
+
+func deriveMachineID() (uuid.UUID, error) {
+	// use /etc/machine-id
+	machineID, err := ioutil.ReadFile("/etc/machine-id")
+	if err != nil {
+		return uuid.Nil, err
+	}
+
+	// 32 bytes, not secret
+	key := []byte("moon machines mortify more minds")
```
maraino: The block size of SHA256 is 64 bytes, that is the recommended size of the key.

command/ssh/certificate.go
```diff
@@ -254,6 +261,28 @@ func certificateAction(ctx *cli.Context) error {
...
 			return err
 		}
+
+		// All host identity certs need a URI SAN to work with our ssh API.
+		if isHost {
+			var u = uuid.Nil
+			if hostID == "" {
+				u, err = deriveMachineID()
+				if err != nil {
+					return err
```
maraino: We should return an error saying something like `cannot read machine-id, please use the --host-id flag`

command/ssh/certificate.go
maraino: Or probably better to use an error like `errs.InvalidFlagValue(...)`

command/ssh/certificate.go
```diff
@@ -429,3 +458,27 @@ type oidcPayload struct {
...
+	key := []byte("moon machines mortify more minds")
+	mac := hmac.New(sha256.New, key)
+	mac.Write(machineID)
+	machineHash := mac.Sum(nil)
+
+	// convert to uuid, definitely not thread-safe
+	r := bytes.NewReader(machineHash)
+	uuid.SetRand(r)
+	defer uuid.SetRand(nil)
+	u, err := uuid.NewRandom()
```
maraino: `uuid.FromBytes(machineHash[:16])` is a simpler way to do this.

dcow: That doesn't make a valid UUID though since the version and type bytes won't be set.

command/ssh/certificate.go
dcow: good idea

command/ssh/certificate.go
dcow: What do you think about this: https://github.com/smallstep/cli/pull/230?

command/ssh/certificate.go
maraino: ok makes sense.

command/ssh/certificate.go
maraino: We should not use SHA1
https://latacora.micro.blog/2018/04/03/cryptographic-right-answers.html

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: See https://github.com/smallstep/cli/pull/230#discussion_r392529744 for my proposal for a more simple method.

dcow: We could also just copy the 16 byes and set the version and type byes manually...

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: We are not trying to prevent collision attacks here. We just need to derive an id based off the machine-id plus a fixed "namespace" we specify that has a high probability of of being unique.

maraino: Well, the other implementation was closer to what `systemd` methods do.
https://github.com/systemd/systemd/blob/master/src/libsystemd/sd-id128/sd-id128.c#L273

dcow: Okay sure.

command/ssh/certificate.go
```diff
@@ -29,7 +36,7 @@ func certificateCommand() cli.Command {
 		Action: command.ActionFunc(certificateAction),
 		Usage:  "sign a SSH certificate using the the SSH CA",
 		UsageText: `**step ssh certificate** <key-id> <key-file>
-[**--host**] [**--sign**] [**--principal**=<string>] [**--password-file**=<path>]
+[**--host**] [--host-id**] [**--sign**] [**--principal**=<string>] [**--password-file**=<path>]
```
maraino: Missing `**`

command/ssh/certificate.go
```diff
@@ -194,6 +203,8 @@ func certificateAction(ctx *cli.Context) error {
 		return errs.IncompatibleFlagWithFlag(ctx, "token", "provisioner-password-file")
 	case isHost && isAddUser:
 		return errs.IncompatibleFlagWithFlag(ctx, "host", "add-user")
+	case !isHost && hostID != "":
+		return errors.New("flag '--host-id' can only be pass if '--host' is set")
```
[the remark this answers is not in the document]
dcow: done

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: done

command/ssh/certificate.go
dcow: @maraino I've looked into this. systemd uses the `machine-id` as the key and an application ID as the message. That is, a `16 byte` key, half of what's proposed here. Further, none of this matters because we're truncating the resulting hash so an attack attempting to recover the `machine-id` is not really possible. What we loose by truncating the sha is entropy space for our resulting uuid.

Anyway, sha3 and blake2 are designed to be used trivially as MACs, so I'm just going to use blake2: https://godoc.org/golang.org/x/crypto/blake2b#New I'm going to use the first 16 bytes of the machine-id as the key and a 16byte fixed value as salt so we get a hash over 32 bytes of data. Here's some more color: https://crypto.stackexchange.com/questions/1186/is-hklengthx-a-secure-mac-construction

dcow: https://github.com/smallstep/cli/pull/230/commits/c15a33968749a7abce1ef1879d69b8128d9d99b5

PR #236, opened 2020-03-27
ssh/certificate: Generate a random UUID by default

dcow:
Apparently some images don't properly handle the machine-id and it ends
up not being unique. By default play it safe and generate our own UUID.
Deriving a UUID from `/etc/machine-id` is still supported. To trigger
that behavior, pass 'machine' as the `--host-id` flag.

### DIFF
--- a/command/ssh/certificate.go
+++ b/command/ssh/certificate.go
@@ -120,6 +120,19 @@ $ step ssh certificate --host --sign \
 	internal.example.com ssh_host_ecdsa_key.pub
 '''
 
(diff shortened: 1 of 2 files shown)


dcow:
You can't use `machine` as a host-id because we don't allow host-ids that aren't uuids. Otherwise I'd agree.

maraino:
> You can't use machine as a host-id because we don't allow host-ids that aren't uuids. Otherwise I'd agree.

Sounds good

PR #241, opened 2020-04-02
Fix autocompletion for file paths in ZSH [#240]

### LINKED ISSUE #240: [BUG] ZSH Autocompletion is not autocompleting paths.
### Subject of the issue
ZSH Autocompletion currently swallows all completion for paths. This makes it mighty painful to work with `step certificates` when the user wants to work with certificates manually.

### Your environment
* OS - 
* Version:
```
Smallstep CLI/0.13.3 (darwin/amd64)
Release Date: 2019-10-02 18:26 UTC
```

### Steps to reproduce
In the console type the following:
```
step certificates fingerprint <Tab>
```

### Expected behaviour
Should start autocompleting paths.

### Actual behaviour
Pressing tab yields nothing.

### Additional context
This is happening because the ZSH autocomplete is prescribing only possible completions using `--generate-bash-completion`. 
In particular, it is calling `_describe` with options piped from `--generate-bash-completion` even if the output is empty. This causes ZSH to think that there are no possible completions.

NonLogicalDev:
Closes #240

### Description
This revision modifies the ZSH autocomplete script to check if the output of `--generate-bash-completion` is empty, and if is, it defaults to autocompleting files.

💔Thank you!

### DIFF
--- a/autocomplete/zsh_autocomplete
+++ b/autocomplete/zsh_autocomplete
@@ -1,10 +1,13 @@
-#compdef _step step
+#compdef step
 
 function _step {
   local -a opts
   opts=("${(@f)$(_CLI_ZSH_AUTOCOMPLETE_HACK=1 ${words[@]:0:#words[@]-1} --generate-bash-completion)}")
-
-  _describe 'values' opts
-
-  return
+  if [[ "${opts}" != "" ]]; then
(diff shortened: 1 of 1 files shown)


dopey:
Yesssss! This has been bothering me for a while and I've had no idea how to fix it. 

Thank you!

PR #242, opened 2020-04-02
ssh/hosts: Use tags intead of groups

dcow:
Tags are a more general, flexible, strategy.


go.mod
```diff
@@ -27,5 +31,6 @@ require (
 	gopkg.in/square/go-jose.v2 v2.4.0
 )
 
-// replace github.com/smallstep/certificates => ../certificates
+replace github.com/smallstep/certificates => ../certificates
```
dcow, line 34: Remove once https://github.com/smallstep/certificates/pull/221 is merged.

utils/cautils/offline.go
```diff
@@ -274,7 +274,7 @@ func (c *OfflineCA) SSHSign(req *api.SSHSignRequest) (*api.SSHSignResponse, erro
 		ValidAfter:  req.ValidAfter,
 		ValidBefore: req.ValidBefore,
 	}
-	cert, err := c.authority.SignSSH(publicKey, signOpts, opts...)
+	cert, err := c.authority.SignSSH(context.Background(), publicKey, signOpts, opts...)
```
dcow, line 277: @maraino is `context.Background()` correct here (and similar elsewhere)? I didn't really know what to use. The other option would be a fresh context.

maraino: This is ok.

PR #248, opened 2020-04-05
Add ARM support for step-cli

NonLogicalDev:
### Description

Partially solves: smallstep/certificates#152

This revision:
- Adds rules for building ARM Linux releases for Arm64 and ArmV7 (tested on my RPI)
- Refactors out the BUNDLE script from Makefile into a shell script
  - improves readability
  - adds support for specifying architecture and artifact directory separately from each other

Thank you!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,41 +37,45 @@ distclean: clean
 #################################################
 
 BINARY_OUTPUT=$(OUTPUT_ROOT)binary/
-BUNDLE_MAKE=v=$v GOOS_OVERRIDE='GOOS=$(1) GOARCH=$(2)' PREFIX=$(3) make $(3)bin/step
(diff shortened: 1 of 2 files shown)


dopey:
Hey, thanks so much for the contribution! 

Just pulled and tested on a release candidate (https://github.com/smallstep/cli/releases) and everything worked!

The cleanup makes a lot of sense too 👍

NonLogicalDev:
Now all is left is to fix up the `certificates` to also do ARM builds.

dopey:
Haha, shoot. Good point. Those projects used to use a common makefile which would have come in handy right about now. 

Feels like this should be pretty easy to port over. I'll try and do it before our next full release.

Issue #247, opened 2020-04-03
additional jwt headers


marcosrmendezthd:
### What would you like to be added

Can we add support for additional jwt token headers, even arbitrary ones to allow for custom validations? I'm specifically looking for `x5t` support, but I think it would be useful to allow more.

https://tools.ietf.org/html/rfc7515#page-12

    JWT=$(step crypto jwt sign --key key.json \
        --iss "issuer@example.com" \
        --aud "audience@example.com" \
        --sub "subject@example.com" \
        --x5t "..." \
        --exp $(date -v+60M +"%s"))

### Why this is needed

Would allow the tool to be used to support additional types of JWT token validations.

PR #250, opened 2020-04-05
adds set flag to allow arbitrary JWT headers for crypto jwt sign command

marcosrmendezthd:
### Description

Fixes https://github.com/smallstep/cli/issues/247 by allowing arbitrary JWT token headers.

For example:

```
    JWT=$(step crypto jwt sign --key key.json \
        --iss "issuer@example.com" \
        --aud "audience@example.com" \
        --sub "subject@example.com" \
        --set "x5t=$X5T_FINGERPRINT" \
        --exp $(date -v+60M +"%s"))

```

💔Thank you!


mmalone:
Hey @marcosrmendezthd, thanks for the PR!

I'm a little nervous about adding such a powerful generic mechanism though. It's been a while since I read the JWT spec, but aren't they pretty prescriptive about what headers are possible? It looks like you're _allowed_ to add custom headers, but I'm not sure why you'd want to. For standard headers I'd definitely prefer to have more specific mechanisms that help ensure the proper requirements are met (e.g., the certificate referenced by an `x5c` or `x5t` header should correspond to the private key used to sign the JWT).

For the `x5t` use case it seems like adding a `--x5t` (and `--x5c` and `--x5t-s256`) flag would be a better option. That way we could do more of the work ourselves and do some validation. Something like:

```
--x5c=cert-file, --certificate=cert-file
Set the `x5c` header to the certificate or certificate chain in `cert-file`. 
Certificates are represented as an array of base64-encoded (*not* base64url-
encoded) DER PKIX certificate values. Certificate chains are ordered from 
leaf to root, with each subsequent certificate certifying the previous one.
The public key in the leaf certificate must correspond to the private `--key`
used to sign the JWT.

--x5t
Include a certificate thumbprint (a.k.a., fingerprint) header. The thumbprint
is the base64url-encoded SHA-1 thumbprint (a.k.a. digest) of the DER 
encoding of the X.509 certificate corresponding to the key used to digitally
sign the JWS, specified with the `--x5c` flag.

--x5t-s256
Include a certificate thumbprint (a.k.a., fingerprint) header. The thumbprint 
is the base64url-encoded SHA-256 thumbprint (a.k.a. digest) of the DER 
encoding of the X.509 certificate [RFC5280] corresponding to the key used
to digitally sign the JWS, specified with the `--x5c` flag.
```

This wouldn't allow you to specify an `x5t` without an `x5c` though. We could tweak the `x5t` options to (optionally) take a cert file or a thumbprint to support this use case though.

If we do end up doing a generic mechanism I'm not sure `--set` is the right name for the flag. How about something more specific like `--hdr, --header`?

marcosrmendezthd:
> Hey @marcosrmendezthd, thanks for the PR!
> 
> I'm a little nervous about adding such a powerful generic mechanism though. It's been a while since I read the JWT spec, but aren't they pretty prescriptive about what headers are possible? It looks like you're _allowed_ to add custom headers, but I'm not sure why you'd want to. For standard headers I'd definitely prefer to have more specific mechanisms that help ensure the proper requirements are met (e.g., the certificate referenced by an `x5c` or `x5t` header should correspond to the private key used to sign the JWT).
> 
> For the `x5t` use case it seems like adding a `--x5t` (and `--x5c` and `--x5t-s256`) flag would be a better option. That way we could do more of the work ourselves and do some validation. Something like:
> 
> ```
> --x5c=cert-file, --certificate=cert-file
> Set the `x5c` header to the certificate or certificate chain in `cert-file`. 
> Certificates are represented as an array of base64-encoded (*not* base64url-
> encoded) DER PKIX certificate values. Certificate chains are ordered from 
> leaf to root, with each subsequent certificate certifying the previous one.
> The public key in the leaf certificate must correspond to the private `--key`
> used to sign the JWT.
> 
> --x5t
> Include a certificate thumbprint (a.k.a., fingerprint) header. The thumbprint
> is the base64url-encoded SHA-1 thumbprint (a.k.a. digest) of the DER 
> encoding of the X.509 certificate corresponding to the key used to digitally
> sign the JWS, specified with the `--x5c` flag.
> 
> --x5t-s256
> Include a certificate thumbprint (a.k.a., fingerprint) header. The thumbprint 
> is the base64url-encoded SHA-256 thumbprint (a.k.a. digest) of the DER 
> encoding of the X.509 certificate [RFC5280] corresponding to the key used
> to digitally sign the JWS, specified with the `--x5c` flag.
> ```
> 
> This wouldn't allow you to specify an `x5t` without an `x5c` though. We could tweak the `x5t` options to (optionally) take a cert file or a thumbprint to support this use case though.
> 
> If we do end up doing a generic mechanism I'm not sure `--set` is the right name for the flag. How about something more specific like `--hdr, --header`?

fair enough. i just borrowed some existing code y'all had that used that flag. i can change it to header if it makes it more palatable. i also added a flag to the `fingerprint` command to support generating the base64 and base64 url encoded fingerprint/thumbprint per [RFC7515](https://tools.ietf.org/html/rfc7515#page-12). the reason why i like the open header example is that it is open.  in our case though the header i'm adding is not used to sign the JWS.

marcosrmendezthd:
@mmalone i've changed `set` to `headers`. migrated the fingerprint options to an optional `format` flag and added more tests. anything else?

PR #251, opened 2020-04-05
Implemented multiarch docker build

christianlupus:
# Description

This PR should make multi-arch docker images possible. Unfortunately, it is impossible to use the classic `build`/`tag`/`push` routine. Instead, the main command is `buildx`. I tried to keep it as close to the original files as I could find it.

I am open to any suggestions and updates.


christianlupus:
Finally, I made it. Most probably the travis scripts should work but I could not check pushing to docker hub via travis.

dopey:
Wow! This is awesome! Thank you so much contributing this. 

We have a release planned for tomorrow(or thursday), and, as this is a rather significant change I'm weary of pulling it in without taking the time to do proper testing and code review (just so I understand). I promise this will be a quick follow once the release is out. I'll add it to our current sprint so I don't forget.

Again, this is awesome! Thanks for taking the time to do this.

christianlupus:
I rebased onto current master to simplify the git history opon merge.

PR #264, opened 2020-04-21
certificates: create: only force default SAN for leaf certs

### LINKED ISSUE #263: Creating certs with root-ca and intermediate-ca profiles includes DNS SAN extension
### Subject of the issue
When creating a Root CA or Intermediate CA certificate using `step certificate create`, the resulting certificate always includes the cert CN in a SAN extension.  CA certs probably shouldn't have SAN extensions at all and many CNs are invalid DNS SANs.

This behaviour also produces a different certificate structure than the one you get by using `step ca init`.

### Your environment
* OS - Ubuntu 18.04
* Version - step 0.14.2

### Steps to reproduce
```
$ step certificate create --profile root-ca "Example Inc. Root CA" root-ca.crt root-ca.key
Please enter the password to encrypt the private key: 
Your certificate has been saved in root-ca.crt.
Your private key has been saved in root-ca.key.

$ step certificate inspect ./root-ca.crt 
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 58419960625675612273431578323333713204 (0x2bf346705ab61c61a1292d8072348d34)
    Signature Algorithm: ECDSA-SHA256
        Issuer: CN=Example Inc. Root CA
        Validity
            Not Before: Apr 20 14:48:39 2020 UTC
            Not After : Apr 18 14:48:39 2030 UTC
        Subject: CN=Example Inc. Root CA
        Subject Public Key Info:
            Public Key Algorithm: ECDSA
                Public-Key: (256 bit)
                X:
                    14:37:1e:0f:46:61:88:b5:54:49:b5:eb:33:5d:d5:
                    09:e4:ac:e9:7d:d6:bf:27:63:7b:97:e4:a2:4b:74:
                    fa:46
                Y:
                    9f:0c:0d:7e:72:99:3b:85:f7:09:d8:3b:9d:fd:97:
                    e5:20:5f:9e:af:1a:3b:82:b0:a2:d9:56:0e:70:ff:
                    75:6b
                Curve: P-256
        X509v3 extensions:
            X509v3 Key Usage: critical
                Certificate Sign, CRL Sign
            X509v3 Basic Constraints: critical
                CA:TRUE, pathlen:1
            X509v3 Subject Key Identifier:
                B1:AF:98:27:78:63:E3:37:67:EC:AA:81:C9:17:45:07:F8:72:3D:B0
            X509v3 Subject Alternative Name:
                DNS:Example Inc. Root CA
    Signature Algorithm: ECDSA-SHA256
         30:46:02:21:00:ed:96:94:b6:6f:45:0d:b9:40:07:38:4a:c8:
         0e:d7:fe:f4:4e:aa:e8:a6:9c:d8:28:3d:4f:2d:39:93:6f:a1:
         c3:02:21:00:96:e1:b7:e5:9b:bd:54:d6:25:9f:b8:1e:1c:47:
         a7:15:2c:55:db:5f:95:e8:bc:52:a8:f2:b2:e5:39:73:68:fd
```

### Expected behaviour
Root CA and Intermediate CA certs should match those produced by `step ca init`.

### Actual behaviour
Root CA and Intermediate CA mistakenly contain SAN extension.

solidgoldbomb:
Forcing this for root CA and intermediate CA certs was resulting
in the subject (typically a company name) being interpreted as
a dnsName SAN.  This produced an invalid certificate.  It is also
highly unusual to have a root CA with a SAN at all as seen in
censys so it's best to leave it out.

Fixes #263

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -270,10 +270,6 @@ func createAction(ctx *cli.Context) error {
 	}
 
 	sans := ctx.StringSlice("san")
-	if len(sans) == 0 {
-		sans = []string{subject}
-	}
-	dnsNames, ips, emails := x509util.SplitSANs(sans)
 
 	var (
 		priv       interface{}
@@ -293,6 +289,11 @@ func createAction(ctx *cli.Context) error {
 		if err != nil {
 			return errors.WithStack(err)
 		}
+
+		if len(sans) == 0 {
+			sans = []string{subject}
+		}
+		dnsNames, ips, emails := x509util.SplitSANs(sans)
 
 		csr := &x509.CertificateRequest{
 			Subject: pkix.Name{
@@ -320,6 +321,12 @@ func createAction(ctx *cli.Context) error {
 			caKeyPath = ctx.String("ca-key")
 			profile   x509util.Profile
 		)
+
+		if (len(sans) == 0) && (prof == "leaf") {
+			sans = []string{subject}
+		}
+		dnsNames, ips, emails := x509util.SplitSANs(sans)
+
 		if bundle && prof != "leaf" {
 			return errs.IncompatibleFlagValue(ctx, "bundle", "profile", prof)
 		}


dopey:
I think this change will also prevent the subject from being a SAN in CSRs, if there are no other SANs. Which, I think, is outside of the scope of the issue/PR. 

Let's make sure we're not changing the way CSRs work.

solidgoldbomb:
Sure, I'll undo the logic change in the CSR code path to avoid the scope creep.  Should be able to push that later tonight.

In case you're interested, I'll provide my rationale for the change as originally submitted so you can understand why I thought it made sense to also change the CSR behaviour in case this hits someone in the future.

The CA/B forum only specifies that the SAN extension is required in issued certificates and that it must contain at least one entry in that SAN extension.  That's a responsibility on the CA as it creates a signed certificate, not on the Applicant (who would be creating the CSR).  I didn't check but it seems like this assertion (must have a SAN) would be better placed in the CA signing path where it should simply refuse to sign a CSR that would result in an illegal cert.  Random tools could send you malformed CSRs.

The change as originally written doesn't prevent a user from deliberately repeating the subject in the CSR SAN (with `--san`), it just prevents it from being automatically (and unavoidably) copied into the SAN.

solidgoldbomb:
Let me know if the latest version works for you.  Also, do I need to address the very small code coverage regression?

dopey:
I should have given a bit more context to my comment as well. 

I totally agree with your rationale. And that's the direction we are currently moving. We made the decision to always have a deafult SAN a few years ago (CN's were already deprecated, but at the time it still seemed like a sane default for most CSRs / certificates). I want to be careful about switching away from that in CSRs because we may have validation against that sort of thing in the CA. I just want to make sure we discuss / think through the change to make sure we don't break other invariants. Hence why I don't want to conflate the CSR change with the one for leaf certs. If you want to post it as a separate issue/PR that's definitely ok. We'll need it soon because we want to be able to create intermediate CSRs.

solidgoldbomb:
Sounds like a good approach to me.  Thanks for the review.

dopey:
You're good on the codecov btw. That tool is more useful for our other repos where we actually have decent unit testing. I just leave it on for this repo as a reminder to feel bad about my habits.

dopey:
I actually added a  small change to include `self-signed` certs as well: 676edc5773949cf200a3a240c561161097875144

PR #266, opened 2020-04-21
Add x5t header option

### LINKED ISSUE #265: Add support for x5t header
It would be great to include the `x5t` header as an option when signing JWS or JWT.

### Why this is needed
Some identity providers require the `x5t` option when issuing JWTs. For example, the Microsoft identity platform requires it when using client certificates as an assertion for generating an access token. (See https://docs.microsoft.com/en-us/azure/active-directory/develop/active-directory-certificate-credentials#header)

This is related to #247, but specifically for the `x5t` header.

davidnarayan:
### Description
Adds option to create and sign a JWS or JWT with the x5t header (the base64url-encoded fingerprint of the signing certificate).

Fixes #265 

💔Thank you!


command/crypto/jws/sign.go
```diff
@@ -201,7 +204,23 @@ func signAction(ctx *cli.Context) error {
 			return errs.MutuallyExclusiveFlags(ctx, "x5c-cert", "jwks")
 		}
 		isX5C = true
-	} else {
+	}
+
+	var isX5T bool
+	if len(x5tCertFile) > 0 {
+		if len(x5tKeyFile) == 0 {
```
dopey, line 211: I assume that `x5c-cert` and `x5t-cert` are mutually exclusive. Let's add that invariant check to both "if" blocks (even though we only need it in one - but for posterity).

flags/flags.go
```diff
@@ -195,6 +195,21 @@ $STEPPATH/config/ca.json`,
...
 	}
 
+	// X5tCert is a cli.Flag used to pass the x5t header certificate thumbprint
+	// for a JWS or JWT.
+	X5tCert = cli.StringFlag{
+		Name:  "x5t-cert",
+		Usage: "Certificate in PEM format to use for the 'x5t' header of a JWS or JWT",
+	}
+
+	// X5tKey is a cli.Flag used to pass the private key (corresponding to the x5t-cert)
```
dopey, line 205: I'm not seeing this variable used anywhere. I guess it's nice to have it defined? _shrug_

dopey:
Thanks for the PR @davidnarayan! It looks great. I'm going to check it out locally to make a few changes - specifically I just want to make the test file adhere more closely with the format that we've become accustomed to. I'll also address the couple of comments that I left and maybe pull out the common code in the `ValidateX5*` methods.

dopey:
Merged by #276

PR #269, opened 2020-04-22
Copy 'Extensions' of csr to 'ExtraExtensions' of cert ...

### LINKED ISSUE #185: certificate sign not respecting input CSR extensions
`step certificate sign` uses default leaf certificate extensions rather than using only the ones requested in the CSR.

### LINKED ISSUE #196: x509 leaf CSR should marshal Extensions to ExtraExtensions
https://github.com/smallstep/cli/blob/master/crypto/x509util/leafProfile.go#L59-L60

See documentation: https://godoc.org/crypto/x509#Certificate

dopey:
* Current code is copying these extensions in a way that any extensions
on the csr will be ignored.

Fixes #185 
Fixes #196

### DIFF
--- a/crypto/x509util/leafProfile.go
+++ b/crypto/x509util/leafProfile.go
@@ -55,8 +55,7 @@ func NewLeafProfileWithCSR(csr *x509.CertificateRequest, iss *x509.Certificate, 
 	}
 
 	sub := defaultLeafTemplate(csr.Subject, iss.Subject)
-	sub.Extensions = csr.Extensions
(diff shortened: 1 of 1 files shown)


dopey:
@maraino One thing to note is that this is the same method used by the CA when it generates a leaf profile from a CSR. So, by fixing this, we'll be allowing the CA to sign and return certificates with arbitrary extensions provided by the user. I assume that's ok, but it will be a change from how the CA currently works (not accepting any extra extensions).

PR #270, opened 2020-04-22
Change trimURLPrefix -> trimURL ...

### LINKED ISSUE #245: Step CLI breaks on URLs with trailing slashes
### Subject of the issue

```
% step certificate inspect https://google.com/
failed to connect: dial tcp: lookup google.com/: no such host
```

### Your environment

```
% step version
Smallstep CLI/0.13.3 (darwin/amd64)
Release Date: 2019-10-02 18:36 UTC
```

### Steps to reproduce

```
% step certificate inspect https://google.com/
failed to connect: dial tcp: lookup google.com/: no such host
```

### Expected behaviour

URLs with a trailing slash (or perhaps even "whole URL"?) should be supported as input.

If not that, perhaps the error returned should more accurately reflect the failure mode?

### Actual behaviour

An incorrect error is returned, describing a name lookup failure

dopey:
* Remove trailing '/' if one exists.
* Fixes for inspect, verify, lint, fingerprint

Fixes #245

### DIFF
--- a/command/certificate/remote.go
+++ b/command/certificate/remote.go
@@ -48,25 +48,24 @@ func getPeerCertificates(addr, roots string, insecure bool) ([]*x509.Certificate
 	return conn.ConnectionState().PeerCertificates, nil
 }
 
-var errNotURL = errors.New("input is not a URL")
-
-// trimURL returns the url split into prefix and suffix and a bool which
-// tells if the input string had a recognizable URL prefix.
+// trimURL returns the host[:port] if the input is a URL, otherwise returns an
+// empty string (and 'isURL:false').
 //
 // Examples:
-// trimURL("https://smallstep.com") -> "https://", "smallstep.com", true
-// trimURL("./certs/root_ca.crt") -> "", "", false
+// trimURL("https://smallstep.com/onbaording") -> "smallstep.com", true, nil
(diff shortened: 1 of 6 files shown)


command/certificate/remote.go
```diff
@@ -47,18 +47,18 @@ func getPeerCertificates(addr, roots string, insecure bool) ([]*x509.Certificate
...
-func trimURLPrefix(url string) (string, string, bool) {
+// trimURL("https://smallstep.com") -> "https://", "smallstep.com", true
+// trimURL("./certs/root_ca.crt") -> "", "", false
+// trimURL("hTtPs://sMaLlStEp.cOm") -> "hTtPs://", "sMaLlStEp.cOm", true
+func trimURL(url string) (string, string, bool) {
 	tmp := strings.ToLower(url)
 	for _, prefix := range urlPrefixes {
 		if strings.HasPrefix(tmp, prefix) {
-			return url[:len(prefix)], url[len(prefix):], true
+			return url[:len(prefix)], strings.TrimSuffix(url[len(prefix):], "/"), true
```
maraino, line 61: I would rather parse that url using `url.Parse(rawurl)` (you will need to rename the argument), and grab the `u.Host`, that would be in the format of `host` or `host:port`.

PR #276, opened 2020-04-30
Add support for x5t header option in JWS and JWT 

### LINKED ISSUE #265: Add support for x5t header

dopey:
Adds option to create and sign a JWS or JWT with the x5t header (the base64url-encoded fingerprint of the signing certificate).

Fixes #265


command/crypto/jws/sign.go
```diff
@@ -194,14 +197,36 @@ func signAction(ctx *cli.Context) error {
...
+		if ctx.IsSet("jwk") {
+			return errs.MutuallyExclusiveFlags(ctx, "x5t-cert", "jwk")
+		}
+		if len(jwks) > 0 {
+			return errs.MutuallyExclusiveFlags(ctx, "x5t-cert", "jwks")
+		}
+		isX5T = true
+	}
+
+	if !isX5C || !isX5T {
```
maraino, line 229: Looks like the proper condition here is `&&`. This is not causing a problem right now because `key` won't be empty because is an alias of `x5c-key` and `x5t-key`

flags/flags.go
```diff
@@ -195,6 +195,21 @@ $STEPPATH/config/ca.json`,
 be stored in the 'x5c' header.`,
 	}
 
+	// X5tCert is a cli.Flag used to pass the x5t header certificate thumbprint
+	// for a JWS or JWT.
+	X5tCert = cli.StringFlag{
+		Name:  "x5t-cert",
+		Usage: "Certificate in PEM format to use for the 'x5t' header of a JWS or JWT",
+	}
```
maraino, line 203: We should have a `<path>` or something tagged with `<>` so the help displays `x5t-cert=path` instead of `x5t-cert=value`

jose/validate_test.go
```diff
@@ -0,0 +1,224 @@
...
+	key, _ := x509.ParsePKCS1PrivateKey(keyBlock.Bytes)
+
+	got, err := ValidateX5T(certfile, key)
+	if err != nil {
+		t.Errorf("ValidateX5T(%q, %q) error: %s", certfile, keyfile, err)
+	}
+
+	assert.Equals(t, want, got, fmt.Sprintf("ValidateX5T(%s, %s)", certfile, keyfile))
+}
+*/
```
maraino, line 224: Remove or fix this test.

command/crypto/jws/sign.go
dopey: good spot

PR #278, opened 2020-05-05
Add a flag to force "localhost" in oauth redirect_uri

### LINKED ISSUE #169: Change OIDC response URL from 127.0.0.1 to localhost
Please change the OIDC response URL from http://127.0.0.0 to http://localhost because 127.0.0.1 causes errors with Microsoft Azure AD because they arent allow to set an application response URL without https to any target except http://localhost

With 127.0.0.1 as reply address you will receive the following errors:

AADSTS50011: The reply url specified in the request does not match the reply urls configured for the application: 'XXXXXXXX'.

irgeek:
I ran into an issue using `step oauth` to authenticate against an Amazon Cognito User Pool. Cognito requires that redirect URIs are HTTPS with the sole exception of _http://localhost_. I saw #169 was already open -- and adding a flag seemed pretty easy -- so I threw this together.

My Golang muscles are a little out of practice, so if there's anything you'd like me to change, I'd be happy to do it. Also, I did read the contributing guidelines. I haven't added any tests -- mostly because I couldn't find any tests for the oauth command. If I missed them, though, just point out where they are and I'll add a test or two.

Fixes #169

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -136,6 +136,10 @@ func init() {
 				Usage: "Callback listener <address> (e.g. \":10000\")",
 			},
 			cli.BoolFlag{
+				Name:  "localhost",
+				Usage: "Use \"localhost\" instead of \"127.0.0.1\" in redirect uri.",
+			},
+			cli.BoolFlag{
 				Name:   "implicit",
 				Usage:  "Uses the implicit flow to authenticate the user. Requires **--insecure** and **--client-id** flags.",
 				Hidden: true,
@@ -155,12 +159,13 @@ func init() {
 
 func oauthCmd(c *cli.Context) error {
 	opts := &options{
-		Provider:         c.String("provider"),
-		Email:            c.String("email"),
-		Console:          c.Bool("console"),
-		Implicit:         c.Bool("implicit"),
-		CallbackListener: c.String("listen"),
-		TerminalRedirect: c.String("redirect-url"),
+		Provider:          c.String("provider"),
+		Email:             c.String("email"),
+		Console:           c.Bool("console"),
+		Implicit:          c.Bool("implicit"),
+		CallbackListener:  c.String("listen"),
+		TerminalRedirect:  c.String("redirect-url"),
+		LocalhostRedirect: c.Bool("localhost"),
 	}
 	if err := opts.Validate(); err != nil {
 		return err
@@ -283,12 +288,13 @@ func oauthCmd(c *cli.Context) error {
 }
 
 type options struct {
-	Provider         string
-	Email            string
-	Console          bool
-	Implicit         bool
-	CallbackListener string
-	TerminalRedirect string
+	Provider          string
+	Email             string
+	Console           bool
+	Implicit          bool
+	CallbackListener  string
+	TerminalRedirect  string
+	LocalhostRedirect bool
 }
 
 // Validate validates the options.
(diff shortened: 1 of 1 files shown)


maraino:
The problem with this PR is that it does not integrates well with [step-ca](https://github.com/smallstep/certificates).

A better approach would be something like this:

```
diff --git a/command/oauth/cmd.go b/command/oauth/cmd.go
index ddb8b68..e44f0b0 100644
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -447,6 +447,16 @@ func (o *oauth) NewServer() (*httptest.Server, error) {
                Config:   &http.Server{Handler: o},
        }
        srv.Start()
+
+       // Update host to use for example localhost
+       if host != "127.0.0.1" {
+               _, p, err := net.SplitHostPort(l.Addr().String())
+               if err != nil {
+                       return nil, errors.Wrapf(err, "error parsing %s", l.Addr().String())
+               }
+               srv.URL = "http://" + host + ":" + p
+       }
+
        return srv, nil
 }

```

This makes use of the `--listen <address>` argument that it can be configured in the ca, so it properly redirects the user to a fixed port, on those identity providers that do not support random ports. My patch would work exactly as `--localhost` if you run:

```sh
step oauth --listen localhost:0
```

It's not so obvious as `--localhost` but it can be documented properly in the help.

What do you think about my option?

maraino:
I've pushed my change as #280.

maraino:
@irgeek: after talking about this internally, we think the integration with `step-ca` is important, and I've merged my PR. 

Unless there's a reason to merge this PR, I will close it in favor of using `step oauth --listen localhost:0`

irgeek:
I'm not bothered about which fix is merged. I've tested your version out and it works with Cognito. I almost fixed it the same way, but I chose to make a new flag because I've had PRs rejected in the past where I overloaded an existing flag to do a second task.

I'm confused, though, by what you mean when you say my fix does not integrate well with `step-ca` -- I don't see how this change and that repo are related. Can someone explain?

maraino:
@irgeek: sure, on `step-ca` you can configure OIDC provisioners, this provisioner will grant TLS and SSH certificates (if enabled) to users authorized by the identity provider. 

So if the identity provider, in your case Cognito, supports OpenID Connect, when you do the oauth flow, Cognito will return the id_token that contains your email. The provisioner will validate the signature with the keys that Cognito provides and a few other things, and grant a TLS certificate with your email as a SAN. Or an SSH certificate with the first part of your email as a principal, as well as your email.

If you don't provide a token, `step ca certificate` will ask `step-ca` for the list of provisioners, and with an OIDC one it will just run `step oauth` with the configuration provided by `step-ca`. 

For those providers that do not support a random port, we added the possibility to specify a listen address, so you can just configure the provisioner with `"listenAddress": ":10000"` to use `http://127.0.0.1:10000`, you can also use a custom interface there, but with my change, this can also be used to configure `http://localhost` instead of `http://127.0.0.1` in the redirect uri, just using `"listenAddress": "localhost:0"` or with a fixed port `"listenAddress": "localhost:10000"`. 

To be able to use `--localhost` we will have to add a new configuration field in `step-ca` and make use of it in `step`

mmalone:
@irgeek thanks so much for your contribution. I hate it when contributed code doesn't make it in. I hope this is the first of many PRs and the rest make it in with proper attribution :/

FYI I believe [this is the relevant code ](https://github.com/smallstep/cli/blob/90dea5d2a7bb28b47cab69570451669a11bd2470/utils/cautils/token_generator.go#L116)that @mariano is referring to. It's called from [here](https://github.com/smallstep/cli/blob/master/command/ca/certificate.go#L188) which calls [this](https://github.com/smallstep/cli/blob/90dea5d2a7bb28b47cab69570451669a11bd2470/utils/cautils/certificate_flow.go#L104) then [this](https://github.com/smallstep/cli/blob/90dea5d2a7bb28b47cab69570451669a11bd2470/utils/cautils/token_flow.go#L86) then gets [here](https://github.com/smallstep/cli/blob/90dea5d2a7bb28b47cab69570451669a11bd2470/utils/cautils/token_flow.go#L116). The CLI gets the relevant configuration from the CA, as @mariano described. We have a blog post that describes the feature in the CA [over here](https://smallstep.com/blog/easily-curl-services-secured-by-https-tls/). The definition of the `listenAddress` parameter is in the CA codebase as part of the OIDC provisioner, [over here](https://github.com/smallstep/certificates/blob/master/authority/provisioner/oidc.go#L64).

Another note on this: if you have an opportunity to bug Amazon about this I'd encourage it! OAuth best current practice is to use `http://127.0.0.1` not `http://localhost`. See [section 8.3](https://tools.ietf.org/html/bcp212#section-8.3) of [BCP212](https://tools.ietf.org/html/bcp212), which says:

> While redirect URIs using localhost (i.e., "http://localhost:{port}/{path}") function similarly to loopback IP redirects described in Section 7.3, the use of localhost is NOT RECOMMENDED.  Specifying a redirect URI with the loopback IP literal rather than localhost avoids inadvertently listening on network interfaces other than the loopback interface.  It is also less susceptible to client-side firewalls and misconfigured host name resolution on the user's device.

Not sure about the best way to bring this to Amazon's attention, or if they'll even care 🤷.

Thanks again! ❤️

PR #286, opened 2020-05-13
Allow localhost as redirect_uri in step oauth

tashian:
This fixes the `step oauth --listen` flag so that it can receive `localhost:port`, `:port`, or `127.0.0.1:port` and the `redirect_uri` will respect what is provided.

This was inspired by a gitter convo last week where someone was trying to do OIDC with AWS Cognito, and Cognito only allows `localhost` redirect URIs.

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -496,16 +496,26 @@ func (o *oauth) NewServer() (*httptest.Server, error) {
 	}
 	srv.Start()
 
-	// Update host to use for example localhost
-	if host != "127.0.0.1" {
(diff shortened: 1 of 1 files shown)


maraino:
@tashian I thought this was already fixed with the lines that you removed.

tashian:
@maraino if I just remove those lines (in `NewServer()`) and set `o.redirectURI` to `srv.URL`, it doesn't allow `localhost` to be passed to `--listen`, because the `httptest` package always turns `localhost` into `127.0.0.1` behind the scenes.

maraino:
This was already fixed with 1b5697ee0bc03390351379e45efdc5a89527726d

PR #289, opened 2020-05-20
Support for AWS Instance Metadata Service v2

joshathysolate:
### Description
This PR corresponds to PR 270 in the smallstep/certificates repo in order to add a new flag called --imds-version so that customers can specify which version(s) of the IMDS they want to support on their AWS provisioner.

By default existing AWS provisioners and any new provisioner where --imds-version is not specified will default to V2 first and then falling back to V1 if an error or an HTTP status of 400 or greater is returned by the V2 service.

### DIFF
--- a/command/ca/provisioner/add.go
+++ b/command/ca/provisioner/add.go
@@ -42,7 +42,7 @@ func addCommand() cli.Command {
 [**--pem-keys=<file>**] [**--ca-config**=<file>]...
 
 **step ca provisioner add** <name> **--type**=[AWS|Azure|GCP]
-[**--ca-config**=<file>] [**--aws-account**=<id>]
+[**--ca-config**=<file>] [**--aws-account**=<id>] [**--imds-version**=<versions>]
 [**--gcp-service-account**=<name>] [**--gcp-project**=<name>]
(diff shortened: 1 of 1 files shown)


command/ca/provisioner/add.go
```diff
@@ -491,12 +498,29 @@ func addAWSProvisioner(ctx *cli.Context, name string, provMap map[string]bool) (
 		return nil, err
 	}
 
+	// validate IMDS versions before we send them over to the CA
+	imdsVersions := ctx.StringSlice("imds-version")
+	if len(imdsVersions) == 0 {
+		imdsVersions = []string{"v2", "v1"}
+	}
```
maraino, line 505: You can set the defaults in the `cli.StringSliceFlag.Value` property.

maraino: You can also leave this as empty and relay in the default logic of the CA.

maraino:
Requires smallstep/certificates#270

jokreliable:
does https://github.com/smallstep/certificates/pull/683 make this unnecessary? (seems related)

maraino:
Yes @jokreliable this is not necessary anymore. Thanks for noticing.

PR #290, opened 2020-05-21
Added support for DN for certificate subjects

joshathysolate:
### Description
This code adds support for using DNs for certificate subjects rather than just creating a simple subject with a CN field.

If the subject of a certificate request follows the standard form of a distinguished name, for example, `CN=myhost,OU=my org,O=my copmany,L=New York,ST=New York,C=US`, then the subject is parsed and each field is added to the pkix.Name object used by functions.

If the subject does not match or an unknown field is encountered, the entire subject is treated as the CN as was the previous behavior.

The following DN fields are supported by this PR:
- **CN**: CommonName
- **OU**: OrganizationalUnit
- **O**: Organization
- **L**: Locality
- **ST**: Province (also accepts **S**, **P**, or **SP**)
- **C**: Country

This is pretty standard behavior for certificate subjects.

One other minor change is that according to the CA/B forum, a CN should always be added to the SAN attribute of any leaf certificate or CSR so some minor changes to that behavior are added as well.

### DIFF
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -11,6 +11,7 @@ import (
 	"github.com/smallstep/cli/token"
 	"github.com/smallstep/cli/ui"
 	"github.com/smallstep/cli/utils/cautils"
+	"github.com/smallstep/cli/utils/pkiutils"
 	"github.com/urfave/cli"
 )
 
@@ -211,7 +212,8 @@ func certificateAction(ctx *cli.Context) error {
 		if ctx.String("token") != "" && len(sans) > 0 {
 			return errs.MutuallyExclusiveFlags(ctx, "token", "san")
 		}
-		if !strings.EqualFold(subject, req.CsrPEM.Subject.CommonName) {
+		pkixName, _ := pkiutils.ParseSubject(subject)
+		if !strings.EqualFold(pkixName.CommonName, req.CsrPEM.Subject.CommonName) {
 			return errors.Errorf("token subject '%s' and argument '%s' do not match", req.CsrPEM.Subject.CommonName, subject)
 		}
 	case token.OIDC: // Validate that the subject matches an email SAN
--- a/command/ca/sign.go
+++ b/command/ca/sign.go
@@ -13,6 +13,7 @@ import (
 	"github.com/smallstep/cli/token"
 	"github.com/smallstep/cli/ui"
 	"github.com/smallstep/cli/utils/cautils"
+	"github.com/smallstep/cli/utils/pkiutils"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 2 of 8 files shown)


dopey:
Thanks for the PR @joshathysolate! I'm gonna look this over in more detail this afternoon, specifically with regards how this might change security considerations. 

Right now we validate the "Subject" of a JWT against the Common Name being requested for the certificate. If we're going to allow users to set more attributes in the "Subject" of the cert, should we also be validating those from the JWT? I doubt it, but want to discuss with the team.

Otherwise the PR looks good to me. We'll probably want some unit testing around `ParseSubject`, but I'm not going to bother you for that until we have a chance to discuss the change.

Again, thanks!

maraino:
@joshathysolate I'm ok of being able to change the subject like that but I'd like a small change.
 
With your changes, the token subject will look like `CN=myhost,OU=my org,O=my copmany,L=New York,ST=New York,C=US`, but the CA only validates the CN (and SANs) so I think it makes more sense to use `"sub":"myhost"` in the token, and use the rest of the properties to change the CSR. Right now as long as the CN and SANs match, you can send a CSR with custom extensions, custom properties in the subject, ... 

By the way, there's some functionality a little bit hidden to support custom Subject properties in the CA, in the authority part of the ca.json you can define a template:

```json
...
"authority": {
  "template": {
    "country": "US",
    "organization": "ACME Corp.",
    "organizationalUnit": "Coyote",
    "locality": "San Francisco",
    "province": "California",
    "streetAddress": "123 Main St"
    }
}
...
```

This might work better for your purpose.

maraino:
The authority template is used to fill those properties that are not defined.

maraino:
By the way, there should be an issue around that we haven't address yet, but we want, is to be able to define the CSR using a JSON template, that template would mainly have the subject, and extensions, but also properties to define which kind of key you want to create, the signature algorithm, ... This would be used for people that requires custom things in the certificate, but it won't be mandatory.

I'm not saying that you should do that 😉, I'm just telling you the direction we want to go.

maraino:
Hi @joshathysolate I've started to work on https://github.com/smallstep/certificates/pull/312, the "template" implementation that I talked about before. This includes some extra restrictions to avoid "unauthorized" information in the final certificate. This part would be incompatible with this PR, so I'm going to close it. Of course, you will be able to override this behavior with custom templates.

Once the part of `step-ca` is done we will work in an extended version for the cli, that will make the creating of the certificates requests way more flexible in `step`, and that combined with the templates in `step-ca` you'll be able to sign certificates that contain only the information that you want to approve.

@joshathysolate Note that with the current version you can configure the `ca.json` with a custom DN template so all your CSRs will be filled with those fields by default (if they do not include them)

```js
{
  "root": "...",
  // ...
  "authority": {
    "template": {
      "country": "US",
      "organization": "Smallstep",
      "organizationalUnit": "Smallstep Eng",
      "locality": "San Francisco",
      "province": "California",
      "streetAddress": "1 The Street St"
    },
    //...
  }
}
```

PR #303, opened 2020-06-18
Fix typo in ssh list command usage

joostd:
### Description

Fix a simple typo in the `subject` positional argument on the usage page of `step ssh list`


maraino:
Thanks for the PR. Merging right now.

PR #304, opened 2020-06-22
Support -sk key types in ssh inspect; refactor inspection code

tashian:
I've added support for the new `-sk` key types in `step ssh inspect`. Also was able to refactor the inspection code. Between `crypto/sshutil/inspect.go` and `crypto/sshutil/sshutil.go` there were two ways of determining public key type, size, and fingerprint. Now there's just one.

Here's a cert signed by an SK-ECDSA key to test with:

````
ssh-rsa-cert-v01@openssh.com AAAAHHNzaC1yc2EtY2VydC12MDFAb3BlbnNzaC5jb20AAAAgNydgSUva3aquOZKH1YPLmUUsdSfNGT3ip5z3G+U2urYAAAADAQABAAABAQDIQqOkTPD8L2xGTHhv8hrUm8qIvz9eoI2oq72s4hGlcPIPp6kEegma0wV8gAzjd3zT39V9bzwyN471tjG+45yNF1NbJiNQ72rLPwLz0FZ5zJfULZ6coqNYDF30B7URs/pGATHeU/ChcszrD33v2HWAbP1C/YzCLjarqk2NRFJOSLcee88pQeOWaCQhkazFMQG3a8N5nw31S9L5CkIUxp0XavsQONSk1WEbifDOekoheVc0+nfYX7nluZlm0zILsKm6107+rva9gwRWZvF1eRH6omTTF76CNsTT9Ed7HcdlDB+1ItXZiLycXMJ5Uga1dAtGECpWJVNFUWns9xxjD4AxAAAAAAAAAAAAAAABAAAAOHVidW50dUBlYzItMTgtMTkxLTk5LTE0Ny51cy1lYXN0LTIuY29tcHV0ZS5hbWF6b25hd3MuY29tAAAACgAAAAZ1YnVudHUAAAAAXvEssAAAAABg0Q8bAAAAAAAAAIIAAAAVcGVybWl0LVgxMS1mb3J3YXJkaW5nAAAAAAAAABdwZXJtaXQtYWdlbnQtZm9yd2FyZGluZwAAAAAAAAAWcGVybWl0LXBvcnQtZm9yd2FyZGluZwAAAAAAAAAKcGVybWl0LXB0eQAAAAAAAAAOcGVybWl0LXVzZXItcmMAAAAAAAAAAAAAAH8AAAAic2stZWNkc2Etc2hhMi1uaXN0cDI1NkBvcGVuc3NoLmNvbQAAAAhuaXN0cDI1NgAAAEEEj9hseuAR7LEe1TnVF41UAufoEMfx4m0CLwwTnxk5u471p03G6fiQyzUdWfpb6p3CNEu4MBfLoDyI8/W/U8q48gAAAARzc2g6AAAAeAAAACJzay1lY2RzYS1zaGEyLW5pc3RwMjU2QG9wZW5zc2guY29tAAAASQAAACEA/xUwZ/quc1yKYEDMOCfaIja+jRKUEON4ZaRwYiYyDRMAAAAgX3418jk6JrpihXKRgh+yakXsbLFggfZ3qrjWOp4GhUcBAAAABQ== carl-step
````

### DIFF
--- a/crypto/sshutil/sshutil.go
+++ b/crypto/sshutil/sshutil.go
@@ -62,6 +62,26 @@ func PublicKey(key ssh.PublicKey) (crypto.PublicKey, error) {
 	}
 }
 
+// Fingerprint returns the key size, fingerprint, comment and algorithm of a
+// public key.
+func Fingerprint(in []byte) (string, error) {
+	key, comment, _, _, err := ssh.ParseAuthorizedKey(in)
+	if err != nil {
+		return "", errors.Wrap(err, "error parsing public key")
+	}
+	if comment == "" {
+		comment = "no comment"
+	}
+
+	// set typ, size
+	typ, size, err := publicKeyTypeAndSize(key)
+	if err != nil {
+		return "", errors.Wrap(err, "error determining key type and size")
+	}
+
+	return fmt.Sprintf("%d %s %s (%s)", size, ssh.FingerprintSHA256(key), comment, typ), nil
+}
+
 func publicKeyTypeAndSize(key ssh.PublicKey) (string, int, error) {
 	var isCert bool
 	if cert, ok := key.(*ssh.Certificate); ok {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crypto/sshutil/sshutil.go [hunk 1]
```diff
@@ -182,49 +171,3 @@ func parseRSA(in []byte) (*rsa.PublicKey, error) {
 	key.N = w.N
 	return &key, nil
 }
-
-// parseECDSA parses an ECDSA key according to RFC 5656, section 3.1.
```
tashian, line 186: What I'm unsure about with this PR is whether we have some reason to prefer these `parse` functions over what is already offered by `crypto/ssh`? These appear to only have been used in `PublicKey` above, which was only ever used to determine the key type in `inspectPublicKey`. I don't see any references to them anywhere else.

maraino: We're not using that method anymore, but it can be useful if we want to extract information about the key of a certificate, or just extract it the from an SSH certificate, we have commands to do that in an x509 certificate, but not yet for a SSH certificate and we should.

x/crypto/ssh does not provide a way to convert a ssh.PublicKey to the actual key, this is why I created those methods. So we should keep them.

PR #305, opened 2020-06-23
Add URI capability to SplitSANs method

dopey:


command/ssh/certificate.go
```diff
@@ -309,15 +309,12 @@ func certificateAction(ctx *cli.Context) error {
...
-			// Hack to add the URI.
-			// TODO: the proper solution would be to add support in
-			// x509util.SplitSANs.
 			template := &x509.CertificateRequest{
 				Subject:        csr.Subject,
 				DNSNames:       csr.DNSNames,
 				IPAddresses:    csr.IPAddresses,
 				EmailAddresses: csr.EmailAddresses,
-				URIs:           []*url.URL{uri},
+				URIs:           append(csr.URIs, uri),
```
maraino, line 317: Prepend the generated uri. There is code that expects the uuid URI to be the first one.

crypto/x509util/crt.go
```diff
@@ -37,8 +39,12 @@ func SplitSANs(sans []string) (dnsNames []string, ips []net.IP, emails []string)
...
 			ips = append(ips, ip)
 		} else {
-			// If not IP then assume DNSName.
-			dnsNames = append(dnsNames, san)
+			if u, err := url.Parse(san); err != nil || u.Scheme == "" {
+				// If not email, ip, or url, then dns name.
+				dnsNames = append(dnsNames, san)
+			} else {
+				uris = append(uris, u)
+			}
```
maraino, line 47: This is probably not strictly ok, but I think it is good enough.

go.mod
```diff
@@ -24,14 +24,15 @@ require (
...
 	golang.org/x/crypto v0.0.0-20200323165209-0ec3e9974c59
-	golang.org/x/net v0.0.0-20191209160850-c0dbc17a3553
+	golang.org/x/net v0.0.0-20200202094626-16171245cfb2
 	golang.org/x/sys v0.0.0-20200106162015-b016eb3dc98e
 	gopkg.in/square/go-jose.v2 v2.4.0
 	howett.net/plist v0.0.0-20200419221736-3b63eb3a43b5 // indirect
 )
 
-// replace github.com/smallstep/certificates => ../certificates
+replace github.com/smallstep/certificates => ../certificates
```
maraino, line 36: Make sure to change this.

crypto/x509util/crt.go
dopey: yeah, I wasn't sure what the right answer was. 🤷‍♂️ If / when people complain, we'll fix it.

PR #309, opened 2020-07-04
s/--sans/--san in token CLI docs

mafrosis:
Should be self-explanatory 😄 

Hopefully I will follow up with a more substantial contribution than one character


maraino:
Thanks @mafrosis, merging right now.

PR #310, opened 2020-07-07
Enable key decryption in `ca renew`

### LINKED ISSUE #288: step ca renew fails with an encrypted key
If an encrypted key is used in `step ca renew`, we do not attempt to decrypt it and use it for the mTLS renewal process.

The function `tls.LoadX509KeyPair(certFile, keyFile string) (tls.Certificate, error)` expects an unencrypted key and if fails [here](https://github.com/smallstep/cli/blob/824f0c263eddb1ad463b5d6ef8afc2dce5622062/command/ca/renew.go#L237-L240).

We can check if a key is encrypted looking at their header like [here](https://github.com/smallstep/cli/blob/824f0c263eddb1ad463b5d6ef8afc2dce5622062/crypto/pemutil/pem.go#L248) and manually create the tls.Certificate if it is encrypted.

dopey:
- Fixes #288

### DIFF
--- a/command/ca/renew.go
+++ b/command/ca/renew.go
@@ -1,6 +1,7 @@
 package ca
 
 import (
+	"crypto"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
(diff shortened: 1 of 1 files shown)


bosszhit4l69 (commented):
https://t.me/joinchat/GdhNjRWs54DKcewWNiPW1g

PR #314, opened 2020-07-08
Don't reuse san and principal flags to avoid confusing usage.

### LINKED ISSUE #262: `step ca token` docs mention that several args are mutually exclusive with `-token`, but doesn't take `-token` flag.
Looks like we're re-using docs for a few flags that are used with both `step ca token` and `step ssh certificate` / `step ca certificate`. Some of those flags are documented as being "mutually exclusive with `-token`", which is confusing since `step ca token` doesn't take `-token`. Might need to be a little less clever with re-use here.

dopey:
Fixes #262

### DIFF
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -120,7 +120,7 @@ $ step ca certificate foo.internal foo.crt foo.key \
 		Flags: []cli.Flag{
 			cli.StringSliceFlag{
 				Name: "san",
-				Usage: `Add <dns|ip|email> Subject Alternative Name(s) (SANs)
+				Usage: `Add <dns|ip|email|uri> Subject Alternative Name(s) (SANs)
 that should be authorized. Use the '--san' flag multiple times to configure
 multiple SANs. The '--san' flag and the '--token' flag are mutually exclusive.`,
 			},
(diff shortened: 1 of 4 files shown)


command/ca/certificate.go
```diff
@@ -118,7 +118,12 @@ $ step ca certificate foo.internal foo.crt foo.key \
 --acme https://acme-staging-v02.api.letsencrypt.org/directory --san bar.internal
 '''`,
 		Flags: []cli.Flag{
-			sanFlag,
+			cli.StringSliceFlag{
+				Name: "san",
+				Usage: `Add <dns|ip|email> Subject Alternative Name(s) (SANs)
```
maraino, line 123: Now we support uri too, right?

command/ca/token.go
```diff
@@ -142,8 +142,20 @@ $ step ca token my-remote.hostname remote_ecdsa --ssh --host
 			certNotBeforeFlag,
 			passwordFileFlag,
 			provisionerKidFlag,
-			sanFlag,
-			sshPrincipalFlag,
+			cli.StringSliceFlag{
+				Name: "san",
+				Usage: `Add <dns|ip|email> Subject Alternative Name(s) (SANs)
```
maraino, line 147: `|uri` here too.

command/ca/certificate.go
dopey: yep, good catch

PR #318, opened 2020-07-10
Remove standard Extensions from ExtraExtensions before CreateCert

### LINKED ISSUE #317: Do not copy all csr extensions to the certificate.
Note: This only applies to the cli, because `step-ca` does not copy any extensions from the CSR to the cert. 

The golang crypto code gives precedence to any extensions in the ExtraExtensions list when converting a cert template to a signed certificate. In the absence of an x509Extension in the ExtraExtensions list, the crypto lib will build the extension from the values in the certificate template and apply some extra logic that may enforce correct certificate use and policies. 

As an example, the crypto lib will the SANextension as `Critical: true` if the Subject of cert template is empty. However, if the SANextension from the CSR is copied to the cert template, the extra logic to set `Critical: true` will not be applied.

I think it's best for our tools to not copy (from the csr to the cert template) any extensions that are also parsed by golang/crypto as x509Extensions.

dopey:
Fixes #317

### DIFF
--- a/crypto/x509util/profile.go
+++ b/crypto/x509util/profile.go
@@ -9,7 +9,6 @@ import (
 	"crypto/x509/pkix"
 	"encoding/asn1"
 	"encoding/pem"
-	"fmt"
 	"math/big"
 	"net"
 	"net/url"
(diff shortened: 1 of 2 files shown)


crypto/x509util/profile.go
```diff
@@ -21,6 +22,38 @@ import (
...
+	oidExtKeyUsage              = []int{2, 5, 29, 15}
+	oidExtExtendedKeyUsage      = []int{2, 5, 29, 37}
+	oidExtAuthorityKeyID        = []int{2, 5, 29, 35}
+	oidExtBasicConstraints      = []int{2, 5, 29, 19}
+	oidExtSubjectAltName        = []int{2, 5, 29, 17}
+	oidExtCertificatePolicies   = []int{2, 5, 29, 32}
+	oidExtNameConstraints       = []int{2, 5, 29, 30}
+	oidExtCRLDistributionPoints = []int{2, 5, 29, 31}
+	oidExtAuthorityInfoAccess   = []int{1, 3, 6, 1, 5, 5, 7, 1, 1}
+	oidStandardExtHashMap       = map[string]int{}
```
maraino, line 36: Not super important, but you should use `map[string]stuct{}` for "sets" as it takes less memory.

crypto/x509util/profile.go
```diff
@@ -21,6 +22,38 @@ import (
...
+	oidStandardExtHashMap[oidHashKey(oidExtSubjectAltName)] = 1
+	oidStandardExtHashMap[oidHashKey(oidExtCertificatePolicies)] = 1
+	oidStandardExtHashMap[oidHashKey(oidExtNameConstraints)] = 1
+	oidStandardExtHashMap[oidHashKey(oidExtCRLDistributionPoints)] = 1
+	oidStandardExtHashMap[oidHashKey(oidExtAuthorityInfoAccess)] = 1
+	oidStandardExtHashMap[oidHashKey(oidExtAuthorityInfoAccess)] = 1
+}
+
+func oidHashKey(id asn1.ObjectIdentifier) string {
+	return strings.Trim(strings.Replace(fmt.Sprint(id), " ", ".", -1), "[]")
```
maraino, line 54: You should also be able to use an oid as a key.

maraino: A simpler option to get the string is `asn1.ObjectIdentifier(id).String()`

PR #324, opened 2020-07-16
Add url functionality to `step otp verify` and more ...

dopey:
- add flags to otp generate usageText

### DIFF
--- a/command/crypto/otp/generate.go
+++ b/command/crypto/otp/generate.go
@@ -21,8 +21,8 @@ func generateCommand() cli.Command {
 		Action: command.ActionFunc(generateAction),
 		Usage:  "generate a one-time password",
 		UsageText: `**step crypto otp generate** [**--issuer**=<name>]
-[**account**=<user-name>] [**--period**=<seconds>] [**--length**=<size>]
-[**--alg**=<alg>] [*--time**=<time|duration>] [**--url**] [**--qr**]`,
+[**--account**=<user-name>] [**--period**=<seconds>] [**--length**=<size>]
+[**--alg**=<alg>] [**--url**] [**--qr**]`,
 		Description: `**step crypto otp generate** does TOTP and HTOP`,
 		Flags: []cli.Flag{
 			cli.StringFlag{
--- a/command/crypto/otp/generate.go
+++ b/command/crypto/otp/generate.go
@@ -17,10 +17,12 @@ import (
 
 func generateCommand() cli.Command {
 	return cli.Command{
-		Name:        "generate",
-		Action:      command.ActionFunc(generateAction),
-		Usage:       "generate a one-time password",
-		UsageText:   `**step crypto otp generate**`,
+		Name:   "generate",
+		Action: command.ActionFunc(generateAction),
+		Usage:  "generate a one-time password",
+		UsageText: `**step crypto otp generate** [**--issuer**=<name>]
+[**account**=<user-name>] [**--period**=<seconds>] [**--length**=<size>]
+[**--alg**=<alg>] [*--time**=<time|duration>] [**--url**] [**--qr**]`,
 		Description: `**step crypto otp generate** does TOTP and HTOP`,
 		Flags: []cli.Flag{
 			cli.StringFlag{
--- a/command/crypto/otp/verify.go
+++ b/command/crypto/otp/verify.go
@@ -74,9 +74,11 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
 
 func verifyAction(ctx *cli.Context) error {
 	var (
-		secret, algStr string
-		period         uint
-		digits         int
+		secret string
+		// defaults
+		period = uint(30)
+		digits = 6
+		algStr = "SHA1"
 	)
 
 	secretFile := ctx.String("secret")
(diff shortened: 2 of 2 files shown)


command/crypto/otp/generate.go
```diff
@@ -17,10 +17,12 @@ import (
...
-		Name:        "generate",
-		Action:      command.ActionFunc(generateAction),
-		Usage:       "generate a one-time password",
-		UsageText:   `**step crypto otp generate**`,
+		Name:   "generate",
+		Action: command.ActionFunc(generateAction),
+		Usage:  "generate a one-time password",
+		UsageText: `**step crypto otp generate** [**--issuer**=<name>]
+[**account**=<user-name>] [**--period**=<seconds>] [**--length**=<size>]
+[**--alg**=<alg>] [*--time**=<time|duration>] [**--url**] [**--qr**]`,
```
maraino, line 25: Is this gofmt?

maraino: And `[**account**=<user-name>]` has the `--` missing.

command/crypto/otp/verify.go
```diff
@@ -21,11 +23,15 @@ func verifyCommand() cli.Command {
 		Name:   "verify",
 		Action: cli.ActionFunc(verifyAction),
 		Usage:  "verify a one-time password",
-		UsageText: `**step crypto otp verify** <secret-file>
+		UsageText: `**step crypto otp verify** **--secret**=<path>
 [**--period**=<seconds>] [**--skew**=<num>] [**--length**=<size>]
-[**--alg**=<alg>] [*--time**=<time|duration>]`,
+[**--alg**=<alg>] [*--time**=<time|duration>] [**-url**]`,
```
maraino, line 28: I don't see the `--url` flag defined, and it has just one `-`.

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,73 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
 func verifyAction(ctx *cli.Context) error {
-	args := ctx.Args()
-	secretFile := args.Get(0)
+	var (
+		secret, algStr string
+		period         uint
+		digits         int
+	)
 
+	secretFile := ctx.String("secret")
```
maraino, line 84: You've changed how secret file is passed, can we keep both cases, and as this is a required parameter, can you return an `errs.RequiredFlag(..)` error.

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,73 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
 	if err != nil {
 		return errs.FileError(err, secretFile)
 	}
-	secret := string(b)
+	secret = string(b)
 	if strings.HasPrefix(secret, "otpauth://") {
-		url, err := otp.NewKeyFromURL(secret)
+		otpKey, err := otp.NewKeyFromURL(secret)
 		if err != nil {
 			return err
```
maraino, line 93: Do we need to wrap here?

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,73 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
+			return errors.Wrap(err, "error parsing TOTP Key URI in secret file")
+		}
+		q := u.Query()
+
+		secret = otpKey.Secret()
+
+		// period query param
+		periods := q["period"]
+		if len(periods) > 0 {
+			period64, err := strconv.ParseUint(periods[0], 10, 32)
```
maraino: if instead of treating the query as a map you use q.Get("period") you will get the first period passed:
```go
if period := q.Get("period"); period != "") { ... }
```

maraino: Same for other query values below.

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,73 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
+		} else {
+			algStr = ""
+		}
 	}
 
-	alg, err := algFromString(ctx, ctx.String("alg"))
+	if period == 0 || ctx.IsSet("period") {
+		period = ctx.Uint("period")
+	}
+	if digits == 0 || ctx.IsSet("period") {
```
maraino: ctx.IsSet("digits")

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,66 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
+	)
 
+	secretFile := ctx.String("secret")
 	b, err := ioutil.ReadFile(secretFile)
 	if err != nil {
 		return errs.FileError(err, secretFile)
 	}
-	secret := string(b)
+	secret = string(b)
 	if strings.HasPrefix(secret, "otpauth://") {
```
maraino, line 90: Just a question here. If the secret is passed as an url does it make sense to read it from a file? Or would something like this be more common?
```sh
step crypto otp verify otpauth://foo.bar.zar 
```
If it makes sense this usage, we can support both.

command/crypto/otp/verify.go
maraino: gofmt?

command/crypto/otp/verify.go
maraino: You've fixed this.

command/crypto/otp/verify.go
maraino: This seems to be resolved with the new commits.

command/crypto/otp/generate.go
dopey: I've just run `gofmt` on it. Don't think there were any changes. Also, `make lint` was passing.

command/crypto/otp/verify.go
dopey: Not sure what you mean by `gofmt`? `make lint` wasn't showing any errors.

command/crypto/otp/verify.go
dopey: what? where?

command/crypto/otp/generate.go
maraino: I'm just saying `gofmt` because before the values were aligned, and now they are not.

command/crypto/otp/verify.go
maraino: Same as before, if `gofmt` doesn't change it, then it should be ok.

command/crypto/otp/verify.go
dopey: I think this is more a matter of taste. From my point of view the "secret" is a password, so to be consistent with our other commands I think it makes sense to read it from a file. Open to your thoughts as well though.

command/crypto/otp/generate.go
maraino: Ok it looks like when a string with backticks takes more than one line, then it breaks the alignment. So it's ok.

command/crypto/otp/verify.go
dopey: And if we require a file the the secret will never show up in your shell logs.

maraino: Ok makes sense.

command/crypto/otp/verify.go
maraino: errors.Wrap(err, "failed to parse secret")

dopey: Oh, I see. I thought you were saying that I shouldn't be wrapping. Got confused. Sure, I'll wrap it.

command/crypto/otp/verify.go
```diff
@@ -67,23 +73,73 @@ as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or "µs"), "ms"
...
 		if err != nil {
-			return err
+			return errors.Wrap(err, "error parsing TOTP Key URI in secret file")
+		}
+		q := u.Query()
+
+		secret = otpKey.Secret()
+		// period query param
+		if periodStr := q.Get("period"); len(periodStr) > 0 {
+			period64, err := strconv.ParseUint(periodStr, 10, 64)
```
maraino, line 112: To avoid overflows converting from uint64/int64 to uint/int, I would replace the 64 to `strconv.IntSize`. Same a few lines below.

maraino: Use bitSize `0`. See https://golang.org/pkg/strconv/#ParseInt
```
The bitSize argument specifies the integer type that the result
must fit into. Bit sizes 0, 8, 16, 32, and 64 correspond to int, 
int8, int16, int32, and int64. If bitSize is below 0 or above 64, 
an error is returned.
```

PR #327, opened 2020-07-22
CA cert flexibility support

maraino:
### Description

This PR adds support for certificate templates in `step certificate create`, as well as support for sending user-provided template variables on `step ca certificate` and `step ca sign`.


command/certificate/create.go
```diff
@@ -145,33 +162,70 @@ Create a CSR and key with underlying OKP Ed25519:
 '''
 $ step certificate create foo foo.csr foo.key --csr --kty OKP --curve Ed25519
 '''
-`,
+
+Create a root certificate using a custom template. The root certificate will
+have a path length constrain that allows at least 2 intermediates:
```
dopey, line 167: constraint

command/certificate/create.go
```diff
@@ -145,33 +162,70 @@ Create a CSR and key with underlying OKP Ed25519:
...
+
+Sign a new intermediate using the previous intermediate, now with path
+length 0 using the **--profile** flag:
+'''
+$ step certificate create --profile intermediate-ca \
+  --ca intermediate_ca.crt --ca-key intermediate_ca_key \
+  "Coyote Corporation" coyote_ca.crt coyote_ca_key
+'''
+
+Create now a leaf certificate and bundle it with the two intermediate certificates and validate it:
```
dopey, line 214: remove 'now'

dopey: Also, maybe mention that the leaf profile is the default, hence why you don't specify  a profile or a template.

command/certificate/create.go
```diff
@@ -257,199 +341,306 @@ func createAction(ctx *cli.Context) error {
...
+		return err
+	}
+
+	// Create X.509 certificate
+	templateData := x509util.CreateTemplateData(subject, sans)
+	certificate, err := x509util.NewCertificate(cr, x509util.WithTemplate(template, templateData))
+	if err != nil {
+		return err
+	}
+	certTemplate := certificate.GetCertificate()
```
dopey, line 467: ugh, these names are kinda confusing. GetCertificate() in my mind would return a Certificate. But in this case it's not actually signed, so it's really just returning a certificatetemplate or some other name for that. It's kinda confusing in practice. Maybe something we can update with 0.15.1

command/certificate/create.go
```diff
@@ -257,199 +341,306 @@ func createAction(ctx *cli.Context) error {
...
+	certTemplate.NotBefore = notBefore
+	certTemplate.NotAfter = notAfter
+
+	if certTemplate.NotBefore.IsZero() {
+		certTemplate.NotBefore = time.Now()
+	}
+	if certTemplate.NotAfter.IsZero() {
+		certTemplate.NotAfter = certTemplate.NotBefore.Add(defaultValidity)
+	}
+	// Check that is not expired
```
dopey, line 482: Check that the certificate is not already expired - OR - Check expiry

command/certificate/create.go
maraino: sure

command/certificate/create.go
maraino: Fixed in d1977e4

command/certificate/create.go
maraino: d1977e4

command/certificate/create.go
maraino: Fixed in d1977e4

command/certificate/create.go
```diff
@@ -257,199 +342,306 @@ func createAction(ctx *cli.Context) error {
...
+		}
 	}
+
 	if err = utils.WriteFile(crtFile, pubBytes, 0600); err != nil {
 		return errs.FileError(err, crtFile)
 	}
 
-	if noPass {
-		_, err = pemutil.Serialize(priv, pemutil.ToFile(keyFile, 0600))
+	ui.Printf("Your certificate request has been saved in %s.\n", crtFile)
```
dopey, line 518: not a certificate request

maraino: Fixed with c81e3e8

PR #339, opened 2020-08-06
brand: macOS

jsoref:
### Description
macOS is the canonical name for Apple's current desktop operating system (and has been for a very long time). It's also the term used on the brew page that's linked from this section.


maraino:
Thanks @jsoref merging right away.

PR #346, opened 2020-08-27
Add support for SSH cert-templates

maraino:
### Description

This PR adds support for ssh certificate templates. It reverts commit b5cf069f9773f5b5a52f0c797b272a58e2b12155.

It also fixes the re-load of an offline client, failing because the lock on badgerDB, and add missing keyIDs in SSHSign requests


utils/cautils/offline.go
```diff
@@ -29,8 +29,17 @@ type OfflineCA struct {
 	configFile string
 }
 
+// offlineInstance is a singleton used for OfflineCA. The use of a singleton is
+// necessary to avoid double initialization. Double initializations are
+// sometimes not possible due to locks likek the on in badgerDb.
```
dopey, line 34: Double initializations are sometimes not possible due to locks - as seen in badgerDB.

dopey (approved):
lgtm

PR #351, opened 2020-09-03
Templates for step certificate sign

maraino:
### Description
This PR adds support for templates and profiles in `step certificate sign`

It also removes the requirement of `-` for stdin in `step certificate inspect`


dopey:
The issue that you linked above as "fixes xxx" is just the ssh cert templates PR. Did you mean to link an issue?

command/certificate/create.go
```diff
@@ -24,6 +24,7 @@ const (
 	profileSelfSigned     = "self-signed"
 	profileIntermediateCA = "intermediate-ca"
 	profileRootCA         = "root-ca"
+	profileDirect         = "direct" // Used only on sign
```
dopey, line 27: I think we should use `csr` here. I think that matches better with the implementation in `certificates` which uses csrTemplate, etc.

command/certificate/sign.go
```diff
@@ -38,18 +60,101 @@ This command returns 0 on success and \>0 if any error occurs.
...
+'''
+$ step certificate sign --bundle --not-before -1m --not-after 16h leaf.csr issuer.crt issuer.key
+'''
+
+Sign an intermediate ca:
+'''
+$ step certificate sign --profile intermediate-ca intermediate.csr issuer.crt issuer.key
+'''
+
+Sign an intermediate ca that can sign other intermediates, in this example the
```
dopey, line 85: grammar

command/certificate/sign.go
```diff
@@ -67,48 +172,226 @@ func signAction(ctx *cli.Context) error {
...
+		fmt.Sprintf("Please enter the password to decrypt %s", keyFile),
+		func(s string) ([]byte, error) {
+			return ui.PromptPassword(s)
+		}))
+	if err != nil {
+		return err
+	}
+	signer, ok := key.(crypto.Signer)
+	if !ok {
+		return errors.Errorf("key in %s does not satisfies the crypto.Signer interface", keyFile)
```
dopey, line 199: does not satisfy

dopey (approved):
lgtm

PR #368, opened 2020-10-09
Allow for encrypted key file in `step ca revoke`

### LINKED ISSUE #367: `step ca revoke --cert <cert> --key <--key>` fails on encrypted keys
### Subject of the issue
It seems that I cannot revoke a certificate created using a CSR by passing the certificate and the private key to the CA.

### Your environment
* OS - macOS
* Version - step 0.15.2

### Steps to reproduce
```console
$ step certificate create --csr foo.example.com foo.csr foo.key
Please enter the password to encrypt the private key:
Your certificate signing request has been saved in foo.csr.
Your private key has been saved in foo.key.
$ step ca sign foo.csr foo.crt
✔ Provisioner: carl@smallstep.com (JWK) [kid: ruHPHGuxX_PZZJr9HUOFZzHzuuIJHRsvDSUw1gU2pME]
✔ Please enter the password to decrypt the provisioner key:
✔ CA: https://localhost:4337
✔ Certificate: foo.crt
$ step ca revoke --cert foo.crt --key foo.key
✔ CA: https://localhost:4337
error loading certificates: tls: failed to parse private key
```

### Expected behavior
The certificate should be revoked by the CA

### Actual behavior
The `step` client fails to parse the private key.
Running `step ca revoke [serial number]` works just fine.

dopey:
Fixes #367

### DIFF
--- a/command/ca/revoke.go
+++ b/command/ca/revoke.go
@@ -409,7 +409,7 @@ func (f *revokeFlow) Revoke(ctx *cli.Context, serial, token string) error {
 
 		certPEMBytes, err := ioutil.ReadFile(certFile)
 		if err != nil {
-			return errors.Wrap(err, "error reading cert")
+			return errors.Wrap(err, "error reading certificate")
 		}
 		key, err := pemutil.Read(keyFile)
 		if err != nil {
--- a/command/ca/revoke.go
+++ b/command/ca/revoke.go
@@ -3,6 +3,8 @@ package ca
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
+	"io/ioutil"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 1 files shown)


command/ca/revoke.go
```diff
@@ -405,11 +407,22 @@ func (f *revokeFlow) Revoke(ctx *cli.Context, serial, token string) error {
 	if len(token) == 0 {
 		certFile, keyFile := ctx.String("cert"), ctx.String("key")
 
-		// If there is no token then we must be doing a Revoke over mTLS.
-		var cert tls.Certificate
-		cert, err = tls.LoadX509KeyPair(certFile, keyFile)
+		certPEMBytes, err := ioutil.ReadFile(certFile)
 		if err != nil {
-			return errors.Wrap(err, "error loading certificates")
+			return errors.Wrap(err, "error reading cert")
```
maraino, line 412: As this will become a user error is better to write `certificate` instead of `cert`

command/ca/revoke.go
```diff
@@ -405,11 +407,22 @@ func (f *revokeFlow) Revoke(ctx *cli.Context, serial, token string) error {
...
+		key, err := pemutil.Read(keyFile)
+		if err != nil {
+			return errors.Wrap(err, "error parsing key")
+		}
+		keyBlock, err := pemutil.Serialize(key)
+		if err != nil {
+			return errors.Wrap(err, "error serializing key")
+		}
+
+		cert, err := tls.X509KeyPair(certPEMBytes, pem.EncodeToMemory(keyBlock))
```
maraino, line 423: You can build the `*tls.Certificate` yourself, with the `x509.Certtificate` and private key. That is basically what `tls.X509KeyPair` does, but it adds some extra validations making sure the public key in the cert matches the private key.

So I guess it's ok like that.

maraino (approved):
Change the error message and it looks good to me.

PR #370, opened 2020-10-13
Add local part of an email and email as a principals.

maraino:
### Description
For user certificates, if an email is passed as a principal, we will include in the principals list the email, and the local-part. This imitates the behavior for OIDC provisioners on other provisioners like JWK.

On `step ssh certificate` we will only include them if no principals are passed using the `--principal` flag.

Fixes #389


dopey (approved):
lgtm

Issue #244, opened 2020-04-02
Support for PKCS12 file format inspection and creation.


NonLogicalDev:
### What would you like to be added
Support for PKCS12, both for inspection through `certificates` and `format`.

### Why this is needed
PKCS12 File format is very useful for distributing Client Certs.
It is not terribly difficult to do so with `openssl`, but it would be amazing to just need a single binary for a more or less common task.

dopey:
Thanks for opening the issue! 

I've added this to our backlog for prioritization, but I doubt we'll have time to accomplish this short term (unless we see a lot of support for it - in which case we'd reprioritize). If anyone from the community wants to jump in and take this on, we'd be happy to help/assist.

NonLogicalDev:
For research it seems the only library capable of working with PKCS12 bot read and write is:
https://godoc.org/software.sslmate.com/src/go-pkcs12

That said this particular statement does not look very good:
>  It is intended for decoding P12/PFX files for use with the crypto/tls package, and for encoding P12/PFX files for use by legacy applications which do not support newer formats. Since PKCS#12 uses weak encryption primitives, it SHOULD NOT be used for new applications.

It boggles my mind that there is seemingly no other easy way to shoving a Client Certificate onto iOS devices and MacOS keychain.

doug-fitzmaurice-rowden:
I've used the `go-pkcs12` library before and I'm interested in taking this on as P12 support in `step` would mean we could drop `openssl` and `keytool` from our build process entirely.

What should the CLI args look like?
You can make both "identity stores" (a leaf certificate and private key, and any number of intermediate certs) and "trust stores" (any number of intermediates and CAs with a special flag to make Java use them to validate other certs), so there are a few possible commands:

```
//Command Name:
step [keystore/p12/truststore]
step certificate [keystore/p12/truststore]

//Identity Stores:
step certificate p12 <crt_file> <key_file>
step p12 identity <crt_file> <key_file>
step certificate p12 <crt_file> <key_file> <intermediate_file>
step certificate p12 <crt_file> <key_file> <intermediate_file> <ca_file>

//Trust Stores:
step certificate p12 <ca_file>
step p12 trust <crt_file> <key_file>
step certificate p12 <ca_file> --trust
step certificate p12 <intermediate_file> <ca_file>
```

What are the conventions so far for args vs options, and how to combine multiple modes of operation into a command?

mmalone:
[...]

PR #372, opened 2020-10-15
Add a `p12` command to create PKCS12 store files

doug-fitzmaurice-rowden:
### Description
This PR adds a `step certificate p12` subcommand to support bundling keys and certificates together into PKCS12 / p12 / pfx files, as discussed in #244 

These are commonly used for installing client certificates into Firefox / Windows, and for configuring server certificates for Java applications.

The arguments / options are an attempt at being easy to use, without forcing the user to understand all the different types of p12 that exist.

A password is specifiable on the CLI as one is required for .p12, but is a bit pointless as the encryption is weak. The author of the upstream library this PR uses suggests using a nominal password then protecting the result file using a stronger external method:
https://github.com/SSLMate/go-pkcs12/blob/master/pkcs12.go#L33

### DIFF
--- a/command/certificate/p12.go
+++ b/command/certificate/p12.go
@@ -57,8 +57,8 @@ $ step certificate p12 trust.p12 --ca ca.crt
 flag multiple times to add multiple CAs or intermediates.`,
 			},
 			cli.StringFlag{
-				Name:  "password",
-				Usage: `Set the password to encrypt the .p12 file.`,
+				Name:  "password-file",
+				Usage: `The path to the <file> containing the password to encrypt the .p12 file`,
 			},
 			flags.Force,
 		},
@@ -74,7 +74,6 @@ func p12Action(ctx *cli.Context) error {
 	crtFile := ctx.Args().Get(1)
 	keyFile := ctx.Args().Get(2)
 	caFiles := ctx.StringSlice("ca")
-	password := ctx.String("password")
 	hasKeyAndCert := crtFile != "" && keyFile != ""
 
 	//If either key or cert are provided, both must be provided
@@ -89,11 +88,21 @@ func p12Action(ctx *cli.Context) error {
 
 	x509CAs := []*x509.Certificate{}
 	for _, caFile := range caFiles {
-		x509CA, err := pemutil.ReadCertificate(caFile)
+		x509Bundle, err := pemutil.ReadCertificateBundle(caFile)
 		if err != nil {
 			return errors.Wrap(err, "error reading CA certificate")
 		}
-		x509CAs = append(x509CAs, x509CA)
+		x509CAs = append(x509CAs, x509Bundle...)
+	}
+
+	var password string
+	var err error
+
+	if passwordFile := ctx.String("password-file"); passwordFile != "" {
+		password, err = utils.ReadStringPasswordFromFile(passwordFile)
+		if err != nil {
+			return err
+		}
 	}
 
 	if password == "" {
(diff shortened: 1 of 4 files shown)


command/certificate/p12.go
```diff
@@ -0,0 +1,140 @@
...
+		Flags: []cli.Flag{
+			cli.StringSliceFlag{
+				Name: "ca",
+				Usage: `Add a CA or intermediate certificate to the .p12 file. Use the '--ca'
+flag multiple times to add multiple CAs or intermediates.`,
+			},
+			cli.StringFlag{
+				Name:  "password",
+				Usage: `Set the password to encrypt the .p12 file.`,
+			},
```
maraino, line 62: For consistency with other commands, it makes more sense to use `--password-file`, and then just use `utils.ReadStringPasswordFromFile(filename string) (string, error)`.

command/certificate/p12.go
```diff
@@ -0,0 +1,140 @@
...
+
+	var pkcs12Data []byte
+	var err error
+
+	if hasKeyAndCert {
+		//If we have a key and certificate, we're making an identity store
+		x509Cert, err := pemutil.ReadCertificate(crtFile)
+		if err != nil {
+			return errors.Wrap(err, "error reading certificate")
+		}
```
maraino, line 115: An X.509 certificate is usually bundled with an intermediary, we do that by default if you are using `step-ca`. I think it makes sense that if we pass a bundled certificate, we should include the intermediary automatically in the list of CAs. Don't you think?
You can use `pemutil.ReadCertificateBundle(filename string) ([]*x509.Certificate, error)` to get all certificates.

maraino: What would happen if a CA is added multiple times? For example one from the cert and one from the `--ca` flag.

maraino (changes requested):
@doug-fitzmaurice-rowden This looks great.
I've proposed a couple of changes for consistency.

command/certificate/p12.go
doug-fitzmaurice-rowden: I'm not sure - I suspect it will be fine as PKCS12 is a dumb collection of certificates. I'll build a few different types and do some tests.

doug-fitzmaurice-rowden: Regarding bundles - if a bundle is passed to the `--ca` option it's easy - append all the certs to the CA list.
If one is passed to `<crt_file>`, we need to work out which one the key applies to, and which is the intermediate as they get stored in different places in the result.

Is the best way to do this extract the public key from the certs and the private key, and compare to find a match? I had a look around the codebase but I couldnt see any existing code that does such a check.

maraino: A X.509 certificate is usually a bundle like:
```
-----BEGIN CERTIFICATE-----
MII.... the server certificate
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE-----
MII.... the intermediate certificate
-----END CERTIFICATE-----
```

It can also include multiple intermediates at the end. For example, in TLS this is required because your browser only knows the root certificates and he needs the intermediate to build the validation path.

Check for example:
```sh
step certificate inspect --format pem --bundle https://www.google.com/
```

So using a PKI with the structure `root`->`intermediate`->`leaf` and using a CA like `step-ca` it will make sense to create pkcs#12 certificate like:

```sh
step ca certificate mariano@smallstep.com mariano.crt mariano.key
step certificate p12 mariano.crt mariano.key --ca $(step path)/certs/root_ca.crt
```

maraino: The server cert is always the first. The list of intermediates are in order at the end.

doug-fitzmaurice-rowden:
Thanks @maraino  - I've pushed another commit with those changes.

maraino (approved):
Looks great @doug-fitzmaurice-rowden

maraino:
@doug-fitzmaurice-rowden the PR is merged to master, we will probably do a new release soon that will include `step certificates p12`.
Thanks again.

doug-fitzmaurice-rowden:
Fantastic, thanks so much!

I have a couple of PRs pending on the upstream library that will hopefully improve compatibility with some Java tools that consume these .p12s.

If  / when those are merged I'll put up another quick PR here to update the version of `go-pkcs12` used.

PR #375, opened 2020-10-20
Add support for cloud CAS in step ca init.

maraino:
### Description
This PR adds support for CloudCAS in step ca init.

Now it's possible to start the ca.json configuration running `step ca init --ra cloudCAS`. The intermediate resource can be passed with the flag `--issuer` and the optional credentials file with `--credentials-file`.

Requires https://github.com/smallstep/certificates/pull/402

### DIFF
--- a/command/ca/init.go
+++ b/command/ca/init.go
@@ -175,17 +175,9 @@ func initAction(ctx *cli.Context) (err error) {
 		if err != nil {
 			return err
 		}
-		file := ctx.String("credentials-file")
-		if !ctx.IsSet("credentials-file") {
-			file, err = ui.Prompt("What credentials file would you like to use? [leave empty if using a service account]",
-				ui.WithValue(ctx.String("credentials-file")))
-			if err != nil {
-				return err
-			}
-		}
 		p.SetAuthorityOptions(&apiv1.Options{
 			Type:                 ctx.String("ra"),
-			CredentialsFile:      file,
+			CredentialsFile:      ctx.String("credentials-file"),
 			CertificateAuthority: iss,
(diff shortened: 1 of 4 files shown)


command/ca/init.go
```diff
@@ -137,10 +162,32 @@ func initAction(ctx *cli.Context) (err error) {
...
+			return err
+		}
+	} else {
+		iss, err := ui.Prompt("What certificate authority would you like to use? (e.g. projects/ca/locations/us-west1/certificateAuthorities/intermediate-ca)",
+			ui.WithValidateNotEmpty(), ui.WithValue(ctx.String("issuer")))
+		if err != nil {
+			return err
+		}
+		file := ctx.String("credentials-file")
+		if !ctx.IsSet("credentials-file") {
```
dopey, line 179: should we be forcing a credentials file entry here instead of just allowing it to be empty if you don't pass one in via flag?

dopey (commented):
left a comment

command/ca/init.go
maraino: You don't need a credentials-file if you're using a service account. 

I want a user to be able to run this command without the prompt, if all parameters are passed, in this case `--credentials-file ""`

dopey: Right, I don't understand why we need the prompt at all. If you don't pass `--credentials-file` then we just use the default one on the instance.

maraino: I was imitating the rest of the parameters here. But yes, it can be confusing.

dopey (approved):
Lgtm

PR #385, opened 2020-11-03
Update go-pkcs12 dependency to include fixes to Java trust compatibility

doug-fitzmaurice-rowden:
Very simple PR to bump the version of the `go-pkcs12` library used to create .P12 files.

There have been a couple of recent merges upstream that improve compatibiltiy of the created files with fussy Java libraries by adding an extra OID value when creating trust stores and adding "friendly names" to certs in the store.


maraino:
@doug-fitzmaurice-rowden Can you fix the conflicts?

doug-fitzmaurice-rowden:
@maraino Done!
Just to check as I'm new to go modules - `go get -u software.sslmate.com/src/go-pkcs12` is all that's required right?

maraino:
@doug-fitzmaurice-rowden Yes that's the command.

Because that go-pkcs12 package doesn't have version tags, the new version is the commit in the go.mod `57fc603b7f52`, a shorted version of your last commit in that package, https://github.com/SSLMate/go-pkcs12/commit/57fc603b7f52aee814e586228d071982b169db35

PR #389, opened 2020-11-23
Arch Linux installation

mettacrawler:
### Description
Please provide a working example of a command that can be used on Arch Linux to install step.

Thank you!


maraino (approved):
Thanks @mettacrawler, it looks like `yay` is the best way to install the last version of the packages.

PR #392, opened 2020-11-26
Add password flags to certificate create & sign

### LINKED ISSUE #391: Support `--password-file` option in `step certificate create` and `step certificate sign` commands

Would it be possible to add the `--password-file` option to the following commands also?
* `step certificate create`
* `step certificate sign`

### Why this is needed

We would like to automate creating additional intermediate CAs in Docker containers and having their certs be signed by another (offline) step-ca instance. To ease the process of initializing the intermediate CAs and then replace their root and intermediate certs (as described in https://smallstep.com/docs/tutorials/intermediate-ca-new-ca#the-secure-way), we would like to automate generating CSRs with a randomly generated secure private key password, and then having the CSR signed by an already set up offline step-ca which also has its private key password stored in a `./secrets/password` file.

At the moment, this cannot be automated fully because both `step certificate create` and `step certificate sign` commands require the user to enter a private key password manually.
This was already mentioned in https://github.com/smallstep/certificates/issues/154#issuecomment-576537128, but I think it has gone unnoticed.

Although not being familiar with go, after looking into the source code it seems that adding the  `--password-file` option could be implemented without much effort by copying most of the code already used in the `step ca ...` commands.

Thanks,
Stefan

dopey:
Fixes #391

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -772,8 +772,8 @@ func savePrivateKey(ctx *cli.Context, filename string, priv interface{}, insecur
 	}
 
 	var pass []byte
-	if ctx.IsSet("password-file") {
-		pass, err = ioutil.ReadFile(ctx.String("password-file"))
+	if passFile := ctx.String("password-file"); passFile != "" {
+		pass, err = ioutil.ReadFile(passFile)
 		if err != nil {
 			return errors.Wrap(err, "error reading encryptiong password from file")
 		}
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -42,8 +42,9 @@ func createCommand() cli.Command {
 		Usage:  "create a certificate or certificate signing request",
 		UsageText: `**step certificate create** <subject> <crt-file> <key-file>
 [**--csr**] [**--profile**=<profile>] [**--template**=<path>]
-[**--ca**=<issuer-cert>] [**--ca-key**=<issuer-key>] [**--password-file**=<path>]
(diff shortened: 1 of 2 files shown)


maraino (changes requested):
I believe the usage is a little bit confusing. In my opinion, we should do:

1. When creating/signing a certificate, `--password-file` should be used to encrypt the created key.
2. When creating a CSR with `--key`, `--password-file` should be used to decrypt the key.
3. When signing a certificate with `--ca`and `--ca-key`, then a new `--ca-password-file` can be used to decrypt the `--ca-key`.

command/certificate/create.go
```diff
@@ -707,17 +764,25 @@ func parseSigner(ctx *cli.Context, defaultSigner crypto.Signer) (*x509.Certifica
...
+		_, err = pemutil.Serialize(priv, pemutil.ToFile(filename, 0600))
 		return err
 	}
 
 	var pass []byte
-	pass, err := ui.PromptPassword("Please enter the password to encrypt the private key",
-		ui.WithValidateNotEmpty())
-	if err != nil {
-		return errors.Wrap(err, "error reading password")
+	if ctx.IsSet("password-file") {
```
maraino, line 775: Is probably better to check for empty string 
```go
if passwordFile := ctx.String("password-file"); passwordFile != "" {
    // ... 
}
```

`--password-file ""` marks password-file as set.

maraino (approved):
@dopey I added a comment for a small change, but it looks ok to me.

maraino (approved):

PR #393, opened 2020-11-26
Do not allow empty password in step cert create ...

### LINKED ISSUE #379: step certificate create should warn that an empty password was used to encrypt private key
When I run `step certificate create` and I get the prompt `Please enter the password to encrypt the private key:`, if I hit enter it will save a private key encrypted with a blank password. `step` should warn that this is the case, and maybe suggest using `--insecure --no-password` if an unencrypted private key is the desired output.

dopey:
- unless --no-password --insecure used

Fixes #379


maraino (approved):
lgtm

PR #395, opened 2020-12-03
Certs requests using K8sSA tokens should not use k8ssa sub

dopey:

### DIFF
--- a/command/ca/sign.go
+++ b/command/ca/sign.go
@@ -191,7 +191,7 @@ func signCertificateAction(ctx *cli.Context) error {
 		return errors.Wrap(err, "error parsing flag '--token'")
 	}
 	switch jwt.Payload.Type() {
-	case token.AWS, token.GCP, token.Azure:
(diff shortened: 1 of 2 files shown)


maraino (changes requested):
We will need to change the check that we do in `step ca sign`
https://github.com/smallstep/cli/blob/bcae4cfa2c8af7394ba64830d67246ba1a960cea/command%2Fca%2Fsign.go#L193-L201
it should be like this
https://github.com/smallstep/cli/blob/bcae4cfa2c8af7394ba64830d67246ba1a960cea/command%2Fca%2Fcertificate.go#L238

And just as a note, in k8s tokens, the subject has a meaning, by default is something like `system:serviceaccount:default:default` but you can change it using annotations to something like `system:serviceaccount:default:my-service`.

I think it make sense to do this change and set the common name to the `subject` positional argument in the cli. A user can always go back to use the actual token subject, either setting that as an argument or using templates.

maraino (approved):
👍

PR #396, opened 2020-12-13
Fixes step ssh documentation url in README.md

korhojoa:
Url went to oauth instead of ssh.


maraino:
@korhojoa Thanks for fixing it.

PR #401, opened 2020-12-21
Fix duration flags incompatible with csr

### LINKED ISSUE #400: step certificate create not-after not set correctly
Attempting to create a certificate which is valid for at least a year.  After creation, the certificate is only valid for a day.

        Validity
            Not Before: Dec 21 06:51:12 2020 UTC
            Not After : Dec 22 06:52:12 2020 UTC

step --version
Smallstep CLI/0.15.2 (linux/amd64)
Release Date: 2020-08-31 21:35 UTC

Executed, 'step certificate create --not-after 8760h --insecure --no-password --csr foo.example.com foo.csr foo.key' followed by, 'step-ca sign foo.csr foo.crt'

I also used time, 2021-12-31T07:20:50.52Z, in the first command but with the same result.  There is no example in the documentation that I saw for time; I'm unsure if the format is correct but the certificate was created.  Regardless, both time and duration are failing.  The root ca is valid until 2029.

dopey:
Fixes #400 

@alphaDev23

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -42,6 +42,7 @@ func createCommand() cli.Command {
 		Usage:  "create a certificate or certificate signing request",
 		UsageText: `**step certificate create** <subject> <crt-file> <key-file>
 [**--csr**] [**--profile**=<profile>] [**--template**=<path>]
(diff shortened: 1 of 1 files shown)


maraino (approved):
👍

PR #403, opened 2020-12-22
Add --pid-file flag to step ca renew

### LINKED ISSUE #397: Allow a `--pid-file` to be passed to `step ca renew --daemon`
Right now this command has a `--pid` option, but this couples the daemon to the service's running process. A `--pid-file` would decouple them.

I know it's possible to use some form of `--exec="kill $(cat /pid/file)"` or maybe `--exec="bash -c 'kill $(cat /pid/file)'"`, but the shell substitution is more error-prone and harder to test, especially in the context of a systemd unit, where there's a different set of substitution rules at play.

dopey:
- Fixes #397

### DIFF
--- a/command/ca/renew.go
+++ b/command/ca/renew.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"io/ioutil"
 	"log"
 	"math/rand"
 	"net/http"
@@ -38,8 +39,10 @@ func renewCertificateCommand() cli.Command {
 		Action: command.ActionFunc(renewCertificateAction),
 		Usage:  "renew a valid certificate",
 		UsageText: `**step ca renew** <crt-file> <key-file>
-[**--ca-url**=<uri>] [**--root**=<file>]
-[**--out**=<file>] [**--expires-in**=<duration>] [**--force**]`,
+[**--ca-url**=<uri>] [**--root**=<path>] [**--password-file**=<path>]
+[**--out**=<path>] [**--expires-in**=<duration>] [**--force**]
+[**--expires-in**=<duration>] [**--pid**=<int>] [**--signal**=<int>]
+[**--exec**=<string>] [**--daemon**] [**--renew-period**=<duration>]`,
 		Description: `
 **step ca renew** command renews the given certificate (with a request to the
 certificate authority) and writes the new certificate to disk - either overwriting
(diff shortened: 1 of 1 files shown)


command/ca/renew.go
```diff
@@ -227,11 +236,29 @@ func renewCertificateAction(ctx *cli.Context) error {
...
+		pidB, err := ioutil.ReadFile(pidFile)
+		if err != nil {
+			return errs.FileError(err, pidFile)
+		}
+		pid, err = strconv.Atoi(strings.TrimSpace(string(pidB)))
+		if err != nil {
+			return errs.Wrap(err, "error converting %s to integer process id", pidB)
+		}
+		if pid <= 0 {
+			return errs.InvalidFlagValue(ctx, "pidFile", strconv.Itoa(pid), "")
```
maraino, line 258: flag is `pid-file`

maraino (changes requested):
Looks good, but I detected a typo in the flag name

command/ca/renew.go
```diff
@@ -38,8 +39,10 @@ func renewCertificateCommand() cli.Command {
 		Action: command.ActionFunc(renewCertificateAction),
 		Usage:  "renew a valid certificate",
 		UsageText: `**step ca renew** <crt-file> <key-file>
-[**--ca-url**=<uri>] [**--root**=<file>]
-[**--out**=<file>] [**--expires-in**=<duration>] [**--force**]`,
+[**--ca-url**=<uri>] [**--root**=<path>] [**--password-file**=<path>]
+[**--out**=<path>] [**--expires-in**=<duration>] [**--force**]
+[**--expires-in**=<duration>] [**--pid**=<int>] [**--signal**=<int>]
+[**--exec**=<string>] [**--daemon**] [**--renew-period**=<duration>]`,
```
maraino, line 45: Can you add `**--pid-file**=<path>` here too

command/ca/renew.go
```diff
@@ -227,11 +236,29 @@ func renewCertificateAction(ctx *cli.Context) error {
 		return errs.RequiredWithFlag(ctx, "renew-period", "daemon")
 	}
 
-	pid := ctx.Int("pid")
-	if ctx.IsSet("pid") && pid <= 0 {
+	if ctx.IsSet("pid") && ctx.IsSet("pid-file") {
+		return errs.MutuallyExclusiveFlags(ctx, "pid", "pid-file")
+	}
+	var pid int
+	if pid = ctx.Int("pid"); ctx.IsSet("pid") && pid <= 0 {
```
maraino, line 243: It will be more clear if we rewrite this as:
```go
pid := ctx.Int("pid")
if ctx.IsSet("pid") && pid <= 0 {
    ...
}
```

maraino (approved):
lgtm

PR #404, opened 2020-12-26
Travis CI -> Github Actions

dopey:


.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+
+  create_release:
+    name: Create Release
+    runs-on: ubuntu-latest
+    needs: lintTestBuild
+    outputs:
+      version: ${{ steps.extract-tag.outputs.VERSION }}
+      vversion: ${{ steps.extract-tag.outputs.VVERSION }}
+      deb_version: ${{ steps.extract-tag.outputs.DEB_VERSION }}
+      gh_deb_version: ${{ steps.extract-tag.outputs.GH_DEB_VERSION }}
```
maraino, line 33: I don't see this variable/environment variable being used.

.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+          tag_name: ${{ github.ref }}
+          release_name: Release ${{ github.ref }}
+          draft: false
+          prerelease: ${{ steps.is_prerelease.outputs.IS_PRERELEASE }}
+
+  upload_git_archive:
+    name: Upload Git Archive
+    runs-on: ubuntu-latest
+    needs: create_release
+    steps:
```
maraino, line 72: We only need this for a release, not a pre-release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+          tag_name="${GITHUB_REF##*/}"
+          hub release edit $(find ./.releases -type f -printf "-a %p ") -m "" "$tag_name"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  build_upload_docker:
+    name: Build & Upload Docker Images
+    runs-on: ubuntu-latest
+    needs: create_release
+    steps:
```
maraino, line 168: We only need this for a release, not a pre-release.

.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15.6'
+      - name: Build
+        id: build
+        run: |
+          PATH=$PATH:/usr/local/go/bin:/home/admin/go/bin
+          make docker-artifacts
```
maraino, line 179: Not related to this file, but the build shows this log:
```
29 WARNING! Using -*** the CLI is insecure. Use --password-stdin.
30 docker buildx create --use --name mybuilder --platform="linux/amd64,linux/386,linux/arm,linux/arm64" || true
31 mybuilder
32 WARNING! Your password will be stored unencrypted in /home/runner/.docker/config.json.
33 Login Succeeded
34 Configure a credential helper to remove this warning. See
35 https://docs.docker.com/engine/reference/commandline/login/#credentials-store
```

maraino: I think we can reduce the amount of time to produce images if we change the way we build those images. If instead of building everything in docker, we just build the binaries in one VM, using GOOS and GOARCH, and then just COPY the appropriate binary to the docker image. We'll get rid of `FROM golang:alpine AS builder`. We are not using CGO so I believe the builds should work properly.

.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+          PATH=$PATH:/usr/local/go/bin:/home/admin/go/bin
+          make -j1 binary-linux binary-darwin binary-windows
+          mkdir -p ./.releases
+          cp ./output/binary/linux/bin/step ./.releases/step-linux-${{ needs.create_release.outputs.version }}
+          cp ./output/binary/linux/bin/step ./.releases/step-linux-latest-integration
+          cp ./output/binary/darwin/bin/step ./.releases/step-darwin-${{ needs.create_release.outputs.version }}
+          cp ./output/binary/windows/bin/step ./.releases/step-windows-${{ needs.create_release.outputs.version }}.exe
+      - name: Upload s3
+        id: upload-s3
+        uses: jakejarvis/s3-sync-action@master
```
maraino, line 208: Should we clone non-github actions?

.github/workflows/release.yml
```diff
@@ -0,0 +1,326 @@
...
+    name: Upload Windows Installer
+    runs-on: ubuntu-latest
+    needs: create_release
+    if: ${{ needs.create_release.outputs.is_prerelease }} == 'false'
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Download Existing Installer
+        id: download
+        uses: prewk/s3-cp-action@master
```
maraino, line 228: Should we clone non-github actions? There might be others.

maraino (changes requested):
Pretty much ok, I've added some comments.

.github/workflows/release.yml
dopey: Fixed the docker login warning.

dopey: With regards to speeding up docker image builds that's probably outside the scope of this PR. Let's open up another issue for it.

dopey: But I like your idea for it.

.github/workflows/release.yml
dopey: Full list: 

jakejarvis/s3-sync-action@master -- https://github.com/jakejarvis/s3-sync-action (500 stars)
prewk/s3-cp-action@master -- https://github.com/prewk/s3-cp-action (fork of s3-sync-action)
smallstep/pkgbuild-action@v1.0.2 -- https://github.com/smallstep/pkgbuild-action
ad-m/github-push-action@master -- https://github.com/ad-m/github-push-action (500 stars)
codecov/codecov-action@v1 -- https://github.com/codecov/codecov-action (550 stars)

For the ones with less stars I have done a code review.

We _could_ fork them all (like I did with the pkgbuild-action one). Depends on what we feel is safe. 
If these were already supported then I would have used the ones from github, but they're not in their lib.

.github/workflows/release.yml
dopey: see above

.github/workflows/release.yml
dopey: We can also pin to a specific SHA

.github/workflows/release.yml
maraino: Sounds good.

.github/workflows/release.yml
maraino: 🤷‍♂️ pinning to a specific SHA would be safer, popular repos can become an attack vector. I don't think GitHub can do anything about that.

.github/workflows/release.yml
```diff
@@ -0,0 +1,275 @@
...
+        id: extract-tag
+        run: |
+          VVERSION=${GITHUB_REF#refs/tags/}
+          VERSION=${GITHUB_REF#refs/tags/v}
+          DEB_VERSION=$(echo ${VERSION} | sed 's/-/~/')
+          GH_DEB_VERSION=$(echo ${VERSION} | sed 's/-/./')
+          echo "::set-output name=VVERSION::${VVERSION}"
+          echo "::set-output name=VERSION::${VERSION}"
+          echo "::set-output name=DEB_VERSION::${DEB_VERSION}"
+          echo "::set-output name=GH_DEB_VERSION::${GH_DEB_VERSION}"
```
maraino, line 47: GH_DEB_VERSION is still twice here.

.github/workflows/release.yml
```diff
@@ -0,0 +1,270 @@
...
+        with:
+          go-version: '^1.15.6'
+      - name: Build
+        id: build
+        run: |
+          PATH=$PATH:/usr/local/go/bin:/home/admin/go/bin
+          make -j1 binary-linux binary-darwin binary-windows
+          mkdir -p ./.releases
+          cp ./output/binary/linux/bin/step ./.releases/step-linux-${{ needs.create_release.outputs.version }}
+          cp ./output/binary/linux/bin/step ./.releases/step-linux-latest-integration
```
maraino, line 127: What is this? Do we need it?

maraino (approved):
lgtm, but I added a question.

.github/workflows/release.yml
dopey: Are you asking if we need the `step-linux-latest-integration`? That's the script used by the integration tests. Rather than specifying a version, they always pull `latest`.

Issue #405, opened 2021-01-03
Certificate Extractation feature request


tehmoon:
Hi Step team!

### What would you like to be added

I wrote a tool more than 2 years ago that does the following two things:
- filter PEMs: output selected PEM blocks based on some filters. This is because I heavily use OpenSSL CLI to decode x509/asn.1 and there is no easy way to do x509 decode arbitrary certificates with: `openssl s_client -showcerts | openssl x509 -noout -text` 
  - position of the blocks
  - type of block
 - extract code signing certificates from Windows Portable Executable files on linux. There is an utility that exists to sign PE files on linux, but nothing that makes it pipeable to, let's say, openssl. There is a really good PE golang standard library [https://golang.org/pkg/debug/pe/](https://golang.org/pkg/debug/pe/) where I had fun digging through the windows docs to extract the right field. The format is in PKCS7 if I recall correctly.

### Why this is needed

I completely rewrote that tool and never ported those two features, mainly because it did not make sense anymore, but I still use them since I don't know of any portable alternative and I thought of making a standalone to extract certificates. But then, I wondered if it would not be better to integrate those two features in Step, sharing the love instead of doing that on my own.

Of course, I know that everybody is busy, so I'll be happy doing it with guidance, or just share the code if the team thinks it's valuable and makes sense.

dopey:
Hey @tehmoon, thanks for opening the issue!

We've got a couple questions --

1. What's the use case for your first bullet point - "output selected PEM blocks based on filters"? We have the `step certificate inspect --format JSON` which you can then pipe to `jq`. But this  probably wouldn't allow for the expressiveness of a regular filter. So, what the need for this feature would be.

2. This sounds cool, and it's in an area where we don't have much expertise. Our main question was where code / functionality like this would live in our tool chain. If you're willing to open a PR, we'd be happy to take a closer look.

We're always open to people opening PRs or just sharing code. As you said, everyone's busy, but we're always happy to work with folks from the community who've spent time using and even modifying our tools. That's where we tend to get the best feedback.

tehmoon:
@dopey thank you for taking the time to go through the issue and to answer!

1. Thank you so much for pointing out `step certificate inspect`. I was able to successfully achieve what I wanted with the `--bundle` option!!
2. Great! I can slowly get started and look at the code. I don't know all the features yet. I'm unsure where it fits yet but I think I'd like to implement:
    1. inspect
    2. verify
[...]

PR #407, opened 2021-01-17
[WIP] Extract certificates from Portable Executable files

tehmoon:
This PR is an implementation of https://github.com/smallstep/cli/issues/405.

Read a PE ([Portable Executable](https://en.wikipedia.org/wiki/Portable_Executable)) file and extract the certificates that are used to verify the executable itself.
The format used is PKCS7 and the pointer to the certificates are in the 5th section of the headers https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers, the certificate table.

This is still a work in progress:

- [x] ~~Port my old code to stepcli~~
- [x] ~~Cleanup the code~~
- [x] ~~Discuss API design~~
- [x] Write tests
- [x] Cleanup and ready for review 

Tring it out:
```
(
  cd /tmp
  curl -OL https://1.na.dl.wireshark.org/win64/Wireshark-win64-3.4.2.exe
  curl -OL https://1.na.dl.wireshark.org/win32/Wireshark-win32-3.4.2.exe
)

# To test a 32 bits program
./bin/step crypto winpe extract /tmp/Wireshark-win32-3.4.2.exe  | step certificate inspect --format json --bundle

# To test a 64 bits program
./bin/step crypto winpe extract /tmp/Wireshark-win64-3.4.2.exe  | step certificate inspect --format json --bundle
```

You'll find more info in the code itself, I added a ton of comments everywhere.

For now I chose to create a new command called `extract-pe` under `certificate`. I was simpler for me than to fit this in `inspect`.
I did not finish writing the command API documentation, waiting on final decision about where to put this code.


tehmoon:
@maraino @dopey This PR is ready for initial feedback. There are things still left for me to do (see description), but I could use some input in terms of what the API would look like! Thank you!

command/certificate/extractPE.go
```diff
@@ -0,0 +1,168 @@
...
+
+var ErrCodeSignCertsNoCertificateTableFound = errors.New("No Certificate Table found")
+
+func extractPECommand() cli.Command {
+	return cli.Command{
+		Name:        "extract-pe",
+		Action:      cli.ActionFunc(extractPEAction),
+		Usage:       `TODO`,
+		UsageText:   `TODO`,
+		Description: `TODO`,
```
maraino, line 25: Please add docs for here. We use a slightly modified version of markdown for this, you can see how other docs are done.

command/certificate/extractPE.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+}
+
+func extractPEAction(ctx *cli.Context) error {
+	if err := errs.MinMaxNumberOfArguments(ctx, 1, 1); err != nil {
```
maraino, line 30: As only one argument is required we should use `errs.NumberOfArguments(ctx, 1)`

maraino: I suppose for this use case doesn't make sense to allow an argument through stdin, right?

command/certificate/extractPE.go
```diff
@@ -0,0 +1,168 @@
...
+		return err
+	}
+
+	file, err := os.Open(path.Clean(ctx.Args().Get(0)))
+	if err != nil {
+		return err
```
maraino, line 36: Wrap errors coming from the stdlib, they allow us to see an stack trace of the error using the environment variable STEPDEBUG. Do something like:
```go
return errors.Wrapf(err, "error opening %s", filename)
```

command/certificate/extractPE.go
```diff
@@ -0,0 +1,168 @@
...
+	var (
+		buff []byte
+		ob   *asn1.ObjectIdentifier
+	)
+
+	switch header := f.OptionalHeader.(type) {
+	case *pe.OptionalHeader32:
+		buff, ob, err = CodeSignCertsParseHeader32(header, file)
+	case *pe.OptionalHeader64:
+		buff, ob, err = CodeSignCertsParseHeader64(header, file)
```
maraino, line 55: add a default case (just in case). returning an error.

command/certificate/extractPE.go
maraino: Try to return always errors from `github.com/pkg/errors`.

command/certificate/extractPE.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	return nil
+}
+
+// Wrapper Functions
+// PE headers can be written by 64 bits architectures or 32 bits architectures.
+// 32 bits is IMHO more spread out because it's backward compatible with 64 bits.
+// I don't think it changes the fact that the program can for a 64 bits arch and the headers 32 bits.
+func CodeSignCertsParseHeader64(header *pe.OptionalHeader64, file io.ReaderAt) ([]byte, *asn1.ObjectIdentifier, error) {
```
maraino, line 74: Use the standard way to comment public methods (or make them private)
```
FuctionName blah blah blah
```

Add comment to all public methods, if not the linter will complain.

maraino (changes requested):
Hi @tehmoon, sorry for the late reply, the PR looks quite good, but I added a few comments.

And I also have a question for you, do you think extract-pe should be a command under `certificates` as it is or should it be under another subcommand. The later case would be useful if we want to add more commands.

Perhaps:
```
step winpe certificate foo.exe
or 
step certificate winpe extract foo.exe
```

What other commands can be useful to implement? Signing of executables?

tehmoon:
Hi @maraino  thxs for taking the time to review this! No worries, got all my time.

I have literally no idea and no strong opinion about the CLI aspect, I'll try my best implementing what makes sense to you all.

I think I'd like to implement the verify command, signing my be too hard for me, I would like to contribute or do it but it might take me a lot of time. Verify sounds easy though, just have to find the appropriate docs to know what are the x509 flags the signtool.exe is looking for so there isn't any security flaws.

Let me know what the right CLI API would be and I will try my best to implement it. Cheers!

maraino:
Hi @tehmoon, I've talked internally about the cli structure and we think it should be `step crypto winpe <verb> ... `, for example:
* `step crypto winpe extract ...`
* `step crypto winpe verify ...`

tehmoon:
@maraino I think `winpe` makes sense, because it still does stand for `windows pe` which is way more explicit than just `pe` in my opinion.

I'll start with moving this to the crypto api at least and address your review comments. What about the output format? right now it's just pem, do you want to have the similar flags as for certificate? ie:
```
      --format=format
          The output format for printing the introspection details.

          format is a string and must be one of:

            text
              Print output in unstructured text suitable for a human to read.

            json
              Print output in JSON format.

            pem
              Print output in PEM format.
```

maraino:
> What about the output format? right now it's just pem, do you want to have the similar flags as for certificate?

Will it differ from `step certificate inspect`? If not I think it makes more sense to just do:
```
step crypto winpe extract foo.exe | step certificate inspect --format json
```

tehmoon:
I don't think it will differ no. Sounds good this is what I had in mind as well. Will spend some time in the next days to finish this.

command/certificate/extractPE.go
tehmoon: Unfortunately no, the library needs an os.File and more specifically the io.ReadAt interface so it can go back and forth reading parts of the file.
I'm using it as part of another project, which is oriented toward streaming. The way I did it was to write all the file in a temporary location. Not so sure about storing it in a []byte because of the fact that it could be quite memory intensive.

But for this cli, yeah my guess is that stdin does not quite make sense

tehmoon:
@maraino I should have addressed all your feedback. I also updated the description with the new command. If that's good to go, I'll work on the tests.

Again, thank you so much for taking the time to help me out.

command/crypto/winpe/winpe.go
```diff
@@ -0,0 +1,191 @@
...
+
+func extractCommand() cli.Command {
+	return cli.Command{
+		Name:      "extract",
+		Action:    cli.ActionFunc(extractPEAction),
+		Usage:     "Extract certificates from Windows Portable Executable files",
+		UsageText: `**step crypto winpe extract <file>`,
+		Description: `**step crypto winpe extract** extract certificate from a Windows Portable Executable file in PEM format.
+
+For examples, see **step help crypto winpe**.
```
maraino, line 40: There are no examples in `step her crypto winpe`

maraino (changes requested):
The PR looks good, I've just added a small comment. Let me know when are you consider this PR done, and I'll take a last look before merging it.

command/crypto/winpe/winpe.go
tehmoon: Good catch. Done!

tehmoon:
@maraino Alright. I've added a test. Basically, I just grabbed the smaller windows installer I could think of: `chrome.exe`, stuck the whole file inside of the test file and basically simply writing it in a temporary file every time.

I can try to find another file, but it won't show up in the final `step` executable.

I also rebased and squashed the commits into a single and clean one.

Let me know what you think.

maraino (approved):
Thanks again @tehmoon, merging right away.

PR #412, opened 2021-01-29
Allow configuration of browser used for OAuth on macOS.

mmalone:
### Description
Allow users to specify which browser is used for OAuth flows. This is only implemented for macOS for now. The tool we're using on Linux (`xdg-open`) doesn't seem to have any simple way to set which browser is used (it always uses your system's default browser). It may be possible to extend this to Windows, but I haven't looked yet. For now I've made this a hidden option since it seems pretty niche and it's not documented, etc.

To configure which browser is used you can pass the `--browser /Path/To/Browser.app` flag into `step oauth`. For other subcommands that depend on `step oauth`, like `step ssh proxycommand` and `step ca certificate`, you can set the `STEP_BROWSER=/Path/To/Browser.app` environment variable or set `"browser": "/Path/To/Browser.app"` in your `defaults.json` (which is normally at `~/.step/config/defaults.json`).

Known issue: if you set the path to something invalid (e.g., `STEP_BROWSER=not-a-browser step oauth`), `step oauth` will fail silently. We should probably fix this before un-hiding this flag (if we ever un-hide it).

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -199,6 +199,11 @@ $ step oauth --client-id my-client-id --client-secret my-client-secret \
 				Usage:  "Allows the use of insecure flows.",
 				Hidden: true,
 			},
+			cli.StringFlag{
+				Name:   "browser",
+				Usage:  "Path to browser that to use for OAuth flow (macOS only).",
+				Hidden: true,
+			},
 			flags.RedirectURL,
 		},
 		Action: oauthCmd,
@@ -215,6 +220,7 @@ func oauthCmd(c *cli.Context) error {
 		Implicit:         c.Bool("implicit"),
 		CallbackListener: c.String("listen"),
 		TerminalRedirect: c.String("redirect-url"),
+		Browser:          c.String("browser"),
 	}
 	if err := opts.Validate(); err != nil {
 		return err
@@ -343,6 +349,7 @@ type options struct {
 	Implicit         bool
 	CallbackListener string
 	TerminalRedirect string
+	Browser          string
 }
 
 // Validate validates the options.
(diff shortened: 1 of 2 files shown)


alanchrt:
Not actually sure where it's supported (`xdg-open` doesn't seem to respect it), but the `BROWSER` env var is pretty common. This command I think would respect `BROWSER` and work across macOS / Linux and python 2 / 3.

```bash
python -m webbrowser "https://www.example.com"
```

Granted, it expects python, but most distros (especially with a web browser) should have a system version installed.

maraino (approved):
lgtm

mmalone:
It'd be nice if we could support browser overrides on other platforms, but I don't think it's worth the effort to do so right now.

I'm not convinced it's worth taking a dependency on `python` to make this work. We'd need more diligence around python 2/3 compatibility and which distros and OSes (e.g., Windows?) come with an appropriate `python`. We'd also open ourselves up to issues with hosed `python` environments (although I think that's mostly a thing of the past). Even if we were satisfied that a working `python` was mostly ubiquitous, I'd be very curious to learn _how_ `python` is doing this under the hood and whether we can just do the same thing instead of taking a dependency.

Another option that would be pretty easy to implement and allow us to mostly punt on this problem would be to support a `--browser-command` flag. Semantics would be something like _we append the URL we're opening to whatever you pass as the value of `--browser-command`_. So you could set `STEP_BROWSER_COMMAND` to `python -m webbrowser` or `BROWSER=/some/path python -m webbrowser` or `open -a /some/path`. That would be easy to implement. I'd want to spend a few minutes noodling on security considerations though.

None of this precludes us from supporting `BROWSER` in the future. That said, ideally `xdg-open` and friends would support `BROWSER` so we didn't have to. In any case, I've hidden the `--browser` flag for now because I don't think we want to promise stable semantics going forward. I think we can replace this implementation with a better alternative if/when we find one.

Long story short, I vote merging this for now and moving on with our lives. But if someone wants to pick this up and improve it, go for it, with the caveats above in mind.

Issue #215, opened 2020-02-06
Optionally force OIDC to go through login/consent screens every time


sourishkrout:
```
prompt
OPTIONAL. Space delimited, case sensitive list of ASCII string values that specifies whether the Authorization Server prompts the End-User for reauthentication and consent. The defined values are:
none
The Authorization Server MUST NOT display any authentication or consent user interface pages. An error is returned if an End-User is not already authenticated or the Client does not have pre-configured consent for the requested Claims or does not fulfill other conditions for processing the request. The error code will typically be login_required, interaction_required, or another code defined in Section 3.1.2.6. This can be used as a method to check for existing authentication and/or consent.
login
The Authorization Server SHOULD prompt the End-User for reauthentication. If it cannot reauthenticate the End-User, it MUST return an error, typically login_required.
consent
The Authorization Server SHOULD prompt the End-User for consent before returning information to the Client. If it cannot obtain consent, it MUST return an error, typically consent_required.
select_account
The Authorization Server SHOULD prompt the End-User to select a user account. This enables an End-User who has multiple accounts at the Authorization Server to select amongst the multiple accounts that they might have current sessions for. If it cannot obtain an account selection choice made by the End-User, it MUST return an error, typically account_selection_required.
```

Slamdunk:
OpenID reference: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
Google reference: https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters

+1 for this feature

PR #413, opened 2021-02-03
OAUTH: add --prompt option for Authentication Request

Slamdunk:
Close https://github.com/smallstep/cli/issues/215

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -182,8 +182,30 @@ $ step oauth --client-id my-client-id --client-secret my-client-secret \
 				Usage: "OAuth scopes",
 			},
 			cli.StringFlag{
-				Name:  "prompt",
-				Usage: "Whether the Authorization Server prompts the End-User for reauthentication and consent",
+				Name: "prompt",
+				Usage: `Whether the Authorization Server prompts the End-User for reauthentication and consent.
+OpenID standard defines the following values, but your provider may support some or none of them:
+
+    **none**
+    :   The Authorization Server MUST NOT display any authentication or consent user interface pages.
+        An error is returned if an End-User is not already authenticated or the Client does not have
+        pre-configured consent for the requested Claims or does not fulfill other conditions for
+        processing the request.
+
+    **login**
+    :   The Authorization Server SHOULD prompt the End-User for reauthentication. If it cannot
+        reauthenticate the End-User, it MUST return an error, typically login_required.
+
+    **consent**
+    :   The Authorization Server SHOULD prompt the End-User for consent before returning information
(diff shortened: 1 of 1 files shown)


maraino:
Hi @Slamdunk, thanks for the PR, do you think we should validate the prompt arguments (none, login, consent, ...)? Do you know if they are standardized or are they IID-specific?

Slamdunk:
So, from what I can tell, most things in the reference page https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest that have **not** the words MUST or REQUIRED are implemented slightly differently in each provider:

1. [Microsoft states](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-auth-code-flow#request-an-authorization-code) it implements all 4, but `The only valid values at this time are login, none, and consent.`
2. [Google doesn't mention](https://developers.google.com/identity/protocols/oauth2/openid-connect#authenticationuriparameters) the `login` value
3. Neither [Github](https://docs.github.com/en/developers/apps/authorizing-oauth-apps) nor [Gitlab](https://docs.gitlab.com/ee/api/oauth2.html) have implemented anything about `prompt`

So to me the value shouldn't be validated for 2 reasons:

1. The RFC about `prompt` is merely a suggestion
2. It would be awkward to me to have `step-cli` telling `prompt=login` is valid, but then Google ignores it

If we don't provide any validation, `step-cli` users would be (a bit more) aware that they have to check their provider for the result, and to no expect `step-cli` to provide the ultimate truth about it.

command/oauth/cmd.go
```diff
@@ -181,6 +181,10 @@ $ step oauth --client-id my-client-id --client-secret my-client-secret \
 				Name:  "scope",
 				Usage: "OAuth scopes",
 			},
+			cli.StringFlag{
+				Name:  "prompt",
+				Usage: "Whether the Authorization Server prompts the End-User for reauthentication and consent",
```
maraino, line 186: @Slamdunk I think we should add to the flag description what are the defined values, either just by name or even with a description formatting those like
https://github.com/smallstep/cli/blob/533e53e9ef691bc5c52cd548355b9b2c06eddaf1/command/crypto/jwk/create.go#L202-L215

maraino (commented):
@Slamdunk Can you add the change in the description and I will merge.

command/oauth/cmd.go
Slamdunk: Done:

image

maraino (approved):
Thanks again @Slamdunk, I've had some time again to look into this and it looks good.

Merging right away.

PR #417, opened 2021-02-10
Add --password-file and --new-password-file flags to change-pass

### LINKED ISSUE #416: Automated way to change key password
I would like to be able to non-interactively change the password for a key using `step crypto change-pass`.

### Why this is needed
The best practices document [suggests](https://smallstep.com/docs/step-ca/certificate-authority-server-production#use-strong-passwords-and-store-them-well)

> After initializing your CA, we recommend that you immediately change the password for the intermediate CA private key

Currently the only option to change the password is to respond to the interactive prompts. As a result, it would be difficult to automate this operation with a utility like Ansible.

I [asked](https://github.com/smallstep/certificates/discussions/467) how to automate the `step crypto change-pass` command on the `certificates` repo discussion, and both @tashian and @maraino mentioned possibly adding a `--password-file` flag to `step crypto change-pass` to address this issue. I would like to respectfully suggest that two flags might be necessary: one to indicate the password to decrypt and another to indicate the password to encrypt.

Additionally, @maraino posted the following command sequence (using bash pipes) to accomplish a non-interactive key password change.

```bash
cat $(step path)/secrets/intermediate_ca_key | \
    step crypto key format --password-file /tmp/old-password.txt | \
    step crypto key format --password-file /tmp/new-password.txt
```

dopey:
- allow the command to be completely non-interactive

fixes #416

### DIFF
--- a/command/crypto/change-pass.go
+++ b/command/crypto/change-pass.go
@@ -24,7 +24,8 @@ func changePassCommand() cli.Command {
 		Action: command.ActionFunc(changePassAction),
 		Usage:  "change password of an encrypted private key (PEM or JWK format)",
 		UsageText: `**step crypto change-pass** <key-file>
-[**--out**=<file>] [**--insecure**] [**--no-password**]`,
+[**--out**=<path>] [**--password-file**=<path>] [**--new-password-file**=<path>]
+[**--insecure**] [**--no-password**]`,
 		Description: `**step crypto change-pass** extracts and decrypts
 the private key from a file and encrypts and serializes the key to disk using a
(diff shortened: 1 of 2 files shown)


maraino (approved):
At some point we need to migrate to use `go.step.sm/crypto` but looks ok. For example we have 
```go
func EncryptJWK(jwk *JSONWebKey, passphrase []byte) (*JSONWebEncryption, error)
```
See https://pkg.go.dev/go.step.sm/crypto@v0.8.0/jose#EncryptJWK

PR #422, opened 2021-02-16
Add no-password to step certificate p12

### LINKED ISSUE #414: Enable a flag to not require a password for p12 file
In order to support certain automated functions for certificate packaging and then import into the Windows Cert Store,  `step certificate p12` should allow for no-password to be required.

Something like an `--insecure` or `--no-password` parameter option.

maraino:
### Description

This PR adds the flags `--no-password --insecure` to `step certificate p12` to encrypt a PKCS#12 archive with a blank password.

Fixes #414

### DIFF
--- a/command/certificate/p12.go
+++ b/command/certificate/p12.go
@@ -51,7 +51,7 @@ Package a CA certificate into a "trust store" for Java applications:
 $ step certificate p12 trust.p12 --ca ca.crt
 '''
 
-Package a certificate and private key with a blank password:
+Package a certificate and private key with an empty password:
 
 '''
 $ step certificate p12 --no-password --insecure foo.p12 foo.crt foo.key
--- a/command/certificate/p12.go
+++ b/command/certificate/p12.go
@@ -49,6 +49,12 @@ Package a CA certificate into a "trust store" for Java applications:
 
 '''
 $ step certificate p12 trust.p12 --ca ca.crt
+'''
+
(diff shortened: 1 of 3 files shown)


command/certificate/p12.go
```diff
@@ -96,26 +112,26 @@ func p12Action(ctx *cli.Context) error {
...
+				return err
+			}
 		}
-	}
 
-	if password == "" {
-		pass, err := ui.PromptPassword("Please enter a password to encrypt the .p12 file")
-		if err != nil {
-			return errors.Wrap(err, "error reading password")
+		if password == "" {
```
dopey, line 125: Shouldn't we just fail here or accept that the user wanted to use an empty password. If the password file you selected was empty, it seems weird to then ask the user for another password.

dopey (approved):
I left a comment about a weird place in the logic. You don't have to change it if you don't want. Everything else looks good.

command/certificate/p12.go
maraino: I left the functionality that was already there. But I think it's better to use `--no-password --insecure` if you really want an empty string as the password

dopey: Then maybe we should just fail if the password file is empty? 

We don't normally check that value. We just use it (I think).

maraino: That also match how other commands act `step crypto jwk create`

PR #426, opened 2021-02-18
Housekeeping github actions

dopey:


.github/workflows/release.yml
```diff
@@ -7,62 +7,48 @@ on:
...
-      - name: Checkout
+      -
+        name: Checkout
         uses: actions/checkout@v2
-      - name: Setup Go
+      -
+        name: Setup Go
         uses: actions/setup-go@v2
         with:
           go-version: '^1.15.6'
```
maraino, line 25: Will this use 1.16? Do we want to be backward compatible with older versions of Go?

.github/workflows/release.yml
```diff
@@ -175,17 +160,17 @@ jobs:
...
       - name: Checkout
         uses: actions/checkout@v2
       - name: Calculate New SHAs
         id: calculate-shas
         run: |
           PKG=step-cli-bin
-          VVER=${{ needs.create_release.outputs.vversion }}
-          VER=${{ needs.create_release.outputs.version }}
+          VVER=${{ needs.test.outputs.vversion }}
+          VER=${{ needs.test.outputs.version }}
```
maraino, line 173: A few lines below we have 
```
echo "${{ secrets.ARCH_SSH_KEY }}" > /tmp/ssh_arch_key && chmod 0400 /tmp/ssh_arch_key
```
Does this key appear in the logs?

maraino (commented):
LGTM but I've added a few questions.

.github/workflows/release.yml
dopey: https://docs.github.com/en/actions/reference/encrypted-secrets#accessing-your-secrets

It shouldn't. This isn't exactly recommended, but I'm open to suggestions.

maraino (approved):
LGTM, but can se use always the same go-version, right now we're using 1.15.6 and 1.15.8.

Make also sure the oauth commands gets properly merged, I'm not sure why it appears in this PR, it's already in master.

PR #433, opened 2021-02-24
Do not require output file in `step ca root` and related commands.

maraino:
### Description
This change allows to print the root or federated roots instead of
downloading to a given file. The commands affected are:
  * step ca root
  * step ca roots
  * step ca federation


dopey (approved):
lgtm

PR #435, opened 2021-02-25
[action] fix microsoft install action

### LINKED ISSUE #434: PowerShell install script installs an HTTP 404 response as "step.exe"
For v0.15.7, the step.exe binary requested by the [PowerShell install script](https://files.smallstep.com/install-step.ps1) did not land in the S3 bucket, resulting in a 404 for clients.

The URL in the PS script is https://files.smallstep.com/step-windows-foo.exe

That “foo” is suspicious... I wonder if the script was updated with the wrong binary url by the release process?

dopey:
Fixes #434

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -160,10 +160,10 @@ jobs:
           PATH=$PATH:/usr/local/go/bin:/home/admin/go/bin
           make -j1 binary-linux binary-darwin binary-windows
           mkdir -p ./.releases
-          cp ./output/binary/linux/bin/step ./.releases/step-linux-${{ needs.test.outputs.version }}
+          cp ./output/binary/linux/bin/step ./.releases/step-linux-${{ needs.create_release.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -195,7 +195,7 @@ jobs:
           DEST: ./install-step.ps1
       - name: Modify Installer
         id: modify
-        run: sed -i -e "s~step-windows-.*.exe~step-windows-foo.exe~g" ./install-step.ps1
+        run: sed -i -e "s~step-windows-.*.exe~step-windows-${{ needs.create_release.outputs.version }}.exe~g" ./install-step.ps1
```
maraino, line 198: We should have install-step.ps1 in github instead of downloading from S3

maraino (approved):
LGTM, but I added a comment that I think we should address.

.github/workflows/release.yml
maraino: Probably naming it like `install-step.ps1.tpl` to avoid any missuse

maraino (approved):

PR #436, opened 2021-03-02
[action] changing the package name of the step deb to match other resources

dopey:


tashian:
tested `make debian` on Ubuntu and installed the resulting `.deb`, lgtm

tashian (approved):

PR #447, opened 2021-03-16
add linux/mips build target

maxthomas:
this allows e.g. some routers to make use of `step cli`

### Description

hey there! huge fan of step and y'all's drive to make this ecosystem better :)

this is a small change that just adds a `linux/mips` build target. the binary size that golang produces isn't ideal for some devices, but `step` is worth it :) there may be some easy wins there for the future, although i haven't done any homework there.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -18,6 +18,7 @@ builds:
       - arm
       - arm64
       - 386
+      - mips
     goarm:
       - 7
     ignore:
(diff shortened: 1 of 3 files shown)


dopey:
Trying out a release candidate. Will merge assuming everything builds smoothly.

Thanks for the contribution, and the kind words 👍 Definitely let us know if you find some more easy wins in the future :)

dopey:
Had to make a couple additions to the goreleaser file to get this working. But all merged up now. Release candidate v0.15.14-rc4 has `mips` builds.

maxthomas:
oh, hey, right on! thanks for seeing that through, really appreciate it! will definitely be following along at the very least :)

PR #449, opened 2021-03-22
Check flag values before printing

tashian:


tashian:
@maraino I've tested this out with various flag combos, seems to work well.

ui/ui.go
```diff
@@ -85,6 +85,13 @@ func PrintSelected(name, value string, opts ...Option) error {
 	return nil
 }
 
+// Print if the value (eg. a specific flag value) is empty
```
maraino, line 88: Comments should have the format `// <method> ...`, in this case `// PrintlnIfEmpty ...`

ui/ui.go
```diff
@@ -85,6 +85,13 @@ func PrintSelected(name, value string, opts ...Option) error {
...
+func PrintlnIfEmpty(s, value string) {
+	if value == "" {
+		Println(s)
+	}
+}
```
maraino, line 93: I'm not sure if we should add this method here, it doesn't fit the way other methods work, ideally `ui.Println` should be able to detect `ui.Options` and act accordingly, but that's a bit more difficult. 

I would move this method to `command/ca/init.go` and make it private.

maraino (changes requested):
Looks good, but I would move that method to init.go. 

I can also see if I can made ui.Println to support options so we can do things like 
```go
ui.Println("What's your name?", ui.WithValue(ctx.String("name"))
```

tashian:
OK, I've addressed your feedback. The version with the options would be nice to have. If you want to do that, please add it. Or just approve and I'll merge :D

PR #450, opened 2021-03-23
Include autocomplete in distribution archives.

slewsys:
### Description
This adds the autocomplete directory to tar.gz archives which is already included in Debian step-cli packages.

💔Thank you!


dopey:
Hey @slewsys, thanks for the contribution. 

We recently switched to using `goreleaser` to manage release archives. The `make/bundle.sh` can still be used though, in case users want to package their own binaries. I'll merge up your PR and then add a small update in the goreleaser yaml that will hopefully get the `autocomplete` files uploaded in the public releases as well.

cheers!

dopey (approved):
lgtm

PR #451, opened 2021-03-24
Add support for JSON Web Keys in step crypto key commands

maraino:
### Description
The following commands now support keys in JWK format.
```
step crypto key fingerprint
step crypto key format
step crypto key inspect
```


dopey (approved):
lgtm

PR #458, opened 2021-04-13
Add --verbose option to step ssh check-host

### LINKED ISSUE #455: SCP does not work with smallstep cli  on windows power shell
### Subject of the issue
Describe your issue here.

### Your environment
* OS - Windows
* Version - 

### Steps to reproduce
it works from scp on Linux, but not from Windows Powershell.

From a fresh Windows install. If we try to scp (from Powershell not WSL) it works. After we configure step (following Client Quick Start — Smallstep), we can't scp anymore, not even scp to not step registred host. 

It seems that the scp process goes in "background", we can't type password for example.

If I comment the lines added by step at the beginning of ~/.ssh/config, scp works again.

### Expected behaviour
SCP should work with step installed

### Actual behaviour
It exits successfully but the file is never uploaded

tashian:
This fixes #455

### DIFF
--- a/command/ssh/checkHost.go
+++ b/command/ssh/checkHost.go
@@ -25,11 +25,11 @@ func checkHostCommand() cli.Command {
 		Usage:  "checks if a certificate has been issued for a host",
 		UsageText: `**step ssh check-host** <hostname>
 [**--ca-url**=<uri>] [**--root**=<file>]
-[**--offline**] [**--ca-config**=<path>]`,
+[**--offline**] [**--ca-config**=<path>] [**--quiet**]`,
 		Description: `**step ssh check-host** checks if a certificate has been issued for a host.
 
-This command returns a zero exit status then the server exists, it will return 1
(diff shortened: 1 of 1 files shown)


command/ssh/checkHost.go
```diff
@@ -47,11 +47,17 @@ $ step ssh check-host internal.smallstep.com
 			flags.Root,
 			flags.Offline,
 			flags.CaConfig,
+			cli.BoolFlag{
+				Name:  "quiet",
+				Usage: `Silently return an exit code.`,
+			},
```
maraino, line 53: This makes the command backward compatible, but we will need to make changes to the configurations.
I think it makes more sense to keep the configuration and add a `"verbose,v"` mode to show the message.
If you do `"verbose,v"`, v is an alias.

maraino (changes requested):

maraino (approved):
👍 lgtm

PR #463, opened 2021-04-15
Always specify a Proxy in all custom transports.

maraino:
### Description
This PR changes all the custom transports being used to add `Proxy: http.ProxyFromEnvironment`

Related to smallstep/certificates#535
Full fix requires smallstep/certificates#540


dopey (approved):
lgtm

PR #464, opened 2021-04-15
In Makefile always use the VERSION from ENV if not empty.

dopey:
- Even when in CI e.g. github-action or travis

Fixes broken version in Homebrew. They're also using github actions so, `GITHUB_REF` is set and overrides our environment given VERSION variable.


maraino (approved):
lgtm

Issue #467, opened 2021-04-27
step certificate verify fails to verify with custom extkeyusages


maraino:
See https://github.com/smallstep/certificates/discussions/556

The problem is that the verify command as it is will only verify certificates with the `ServerAuth` extended key usage. 

Go's `x509.VerifyOptions` has a `KeyUsage` parameter to verify in the given cert, if none is used (as it is) then it expects `x509.ExtKeyUsageServerAuth`, we can also allow anything with `x509.ExtKeyUsageAny`.

maraino:
With the change in #468 we allow any kind of cert. To fix this issue properly we might want to add a flag like `--purpose <pupose>`, where `purpose` can be `server`, `client`, ... to configure the KeyUsage with the required ones.

PR #468, opened 2021-04-27
Allow verification of any type of certs.

maraino:
### Description

This change allows to validate any type of cert, no matter the extended key usage of it. Previously only certs with server auth were considered valid.

In the future we should add some kind of 'purpose' flag and configure the verify options accordingly.

Related to #467

### DIFF
--- a/command/certificate/verify.go
+++ b/command/certificate/verify.go
@@ -171,7 +171,7 @@ func verifyAction(ctx *cli.Context) error {
 		// Support verification of any type of cert.
 		//
 		// TODO: add something like --purpose client,server,... and configure
-		// this property according.
+		// this property accordingly.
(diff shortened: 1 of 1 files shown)


command/certificate/verify.go
```diff
@@ -168,6 +168,11 @@ func verifyAction(ctx *cli.Context) error {
 		DNSName:       host,
 		Roots:         rootPool,
 		Intermediates: intermediatePool,
+		// Support verification of any type of cert.
+		//
+		// TODO: add something like --purpose client,server,... and configure
+		// this property according.
+		KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageAny},
```
dopey, line 175: accordingly

dopey (approved):
lgtm -- fix one word in comment

Issue #221, opened 2020-02-19
Feature Request: ARM releases


gclawes:
### What would you like to be added
Releases for ARM platforms (see related ticket https://github.com/smallstep/certificates/issues/152).  Creating this issue to track ARM releases for this repo.

### Why this is needed
See https://github.com/smallstep/certificates/issues/152

PR #469, opened 2021-04-28
Add armv6 for build

jabbrwcky:
### Description
This PR extends ARM support to armv6. (Original issue in #221)

I tested the built binary on the venerable RPi2 that provides DNS on my home network

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,9 @@ binary-linux:
 binary-linux-arm64:
 	$(call BUNDLE_MAKE,linux,arm64,,$(BINARY_OUTPUT)linux.arm64/)
 
+binary-linux-armv6:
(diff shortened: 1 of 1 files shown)


maraino:
Hi @jabbrwcky, we're using https://goreleaser.com now for builds, so to get it in our builds the changes have to go to https://github.com/smallstep/cli/blob/master/.goreleaser.yml

jabbrwcky:
I did not see that, @maraino . I'll update my PR accordingly

tashian:
Hi @jabbrwcky thanks for the contribution.
I created a PR that updates the GoReleaser config (it was just a one-line change) in #478 and smallstep/certificates#570, so I'm going to close this.
You should see arm6 builds in the next release.
Separately, we'll need to go through the Makefile and remove some of the old recipes that are no longer in use.

PR #479, opened 2021-05-12
Dynamically determine version number of latest GitHub release in PowerShell installer

tashian:
@dopey what do you think of this magic


dopey (approved):
lgtm

PR #481, opened 2021-05-17
Reverse insecure and --no-password in RequireWithFlag

### LINKED ISSUE #480: Error message gives inverted order of required flags
### Subject of the issue

There's an error message that reports on erroneous CLI flag usage, but it seems the wrong way around.

### Your environment
* OS - Windows 10
* Version - `Smallstep CLI/0.15.17-rc1.test-arm6 (windows/amd64)`

### Steps to reproduce

Run [this command](https://smallstep.com/docs/step-cli/reference/crypto/keypair#options), which fails:

```bash
$ step crypto keypair test.pub test.key --no-password
flag '--insecure' requires the '--no-password' flag
```

The docs tell us:

> **`--no-password`** Do not ask for a password to encrypt a private key. Sensitive key material will be written to disk unencrypted. This is not recommended. Requires `--insecure` flag.

So I guess the error message should be more like: `flag '--no-password' requires the '--insecure' flag`, aka the other way around.

dopey:
Fixes #480

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -45,7 +45,8 @@ func createCommand() cli.Command {
 [**--password-file**=<path>] [**--ca**=<issuer-cert>]
 [**--ca-key**=<issuer-key>] [**--ca-password-file**=<path>]
 [**--san**=<SAN>] [**--bundle**] [**--key**=<path>]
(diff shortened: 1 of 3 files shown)


maraino (approved):
lgtm

PR #494, opened 2021-06-16
Make empty STDIN work for `crypto jwt sign` when no input is specified

jgiannuzzi:
### Description
Some CI platforms will feed us an empty pipe as STDIN (i.e. GitHub Actions). This means that the following command — which would work fine locally — will fail on the CI:
```
$ step crypto jwt sign --key key.pem --subtle
error decoding JSON from STDIN: EOF
```

This PR ensures that if no payload is specified and STDIN is empty, we treat it as a valid empty JSON. This is consistent with the current behaviour when STDIN is not a pipe.

### DIFF
--- a/command/crypto/jwt/sign.go
+++ b/command/crypto/jwt/sign.go
@@ -476,6 +476,11 @@ func readPayload(filename string) (interface{}, error) {
 
 	v := make(map[string]interface{})
 	if err := json.NewDecoder(r).Decode(&v); err != nil {
+		// Some CI platforms will feed an empty pipe as STDIN.
+		// In that case we should treat it as a valid empty JSON.
+		if filename == "" && errors.Is(err, io.EOF) {
+			return v, nil
+		}
 		if filename == "" || filename == "-" {
 			return nil, errors.Wrap(err, "error decoding JSON from STDIN")
 		}


maraino (changes requested):
The change should also take the `filename == "-"` into account, so it should be something like:
```diff
diff --git a/command/crypto/jwt/sign.go b/command/crypto/jwt/sign.go
index 6d61bde..1d9913d 100644
--- a/command/crypto/jwt/sign.go
+++ b/command/crypto/jwt/sign.go
@@ -477,6 +477,11 @@ func readPayload(filename string) (interface{}, error) {
        v := make(map[string]interface{})
        if err := json.NewDecoder(r).Decode(&v); err != nil {
                if filename == "" || filename == "-" {
+                       // Some CI platforms will feed an empty pipe as STDIN. In that case
+                       // we should treat it as a valid empty JSON.
+                       if errors.Is(err, io.EOF) {
+                               return v, nil
+                       }
                        return nil, errors.Wrap(err, "error decoding JSON from STDIN")
                }
                return nil, errors.Wrapf(err, "error decoding JSON from %s", filename)
```

jgiannuzzi:
Hey @maraino, thanks a lot for your review!

My rationale was to keep the same behaviour differences that we have today between specifying `-` for the payload and not specifying anything. See the snippet below: https://github.com/smallstep/cli/blob/97f2ed457d66711d0cdeb7be2ccb04fe7f6df2f4/command/crypto/jwt/sign.go#L458-L468

Currently this is the behaviour when using `/dev/null` as STDIN:
```sh
$ step crypto jwt sign --key key.pem --subtle </dev/null
# a JWT gets displayed
$ step crypto jwt sign --key key.pem --subtle - </dev/null
error decoding JSON from STDIN: EOF
```

If you think they should behave in the same way, I would be more than happy to amend my PR!
And food for thought: maybe passing an empty file as the payload should also work in the same way then?

jgiannuzzi:
Hi @maraino, what are your thoughts on the above?

maraino (approved):
@jgiannuzzi Make sense to have the same behavior with `echo -n | step ..` and `step ... < /dev/null`

Issue #398, opened 2020-12-16
Testing certificates for verdancy


tashian:
I'd love to be able to know whether or not a certificate is about to expire, without running something like:
```bash
step certificate inspect $CERT_LOCATION --format json | \
        jq -e "(((.validity.start | fromdate) + \
                 ((.validity.end | fromdate) - (.validity.start | fromdate)) * 0.66) \
                  - now) <= 0'
```

For scripting purposes, I'd love a command such as `step certificate verify --verdancy` that would simply say whether or not the certificate is ready for renewal, and return an exit code:

* 0 GREEN — the leaf is < 66% through its lifetime
* 1 YELLOW — the leaf is >= 66% and < 90% through its lifetime
* 2 RED — the leaf is >= 90% through its lifetime
* 3 BROWN — the leaf is expired
* 255 if there was an error

tashian:
Instead of talking about ripness, one could also say that a certificate is "yellowing" — like a leaf that's about to fall off the tree. But, also like a yellow stoplight.

dopey:
`step certificate expired` and `step certificate verdancy [**--expires-in**=<duration-or-% remaining>] <cert-or-url>`

tashian:
Can we do this for SSH certs too? This would allow the same systemd renewal method for SSH host certs.

tommy-56:
step certificate needs-renewal addresses this issue and was merged in #495

PR #495, opened 2021-06-21
Megan/expiration

cookie-annihilator:
Fixes issue #398 , added flag that allows you to check how much longer until a certificate is about to expire, along with the color coding the outputs based on if it is about to expire/expired (red), close (yellow), or still has under 66% of its life used (green). added to description how to use this function.

### DIFF
--- a/command/certificate/certificate.go
+++ b/command/certificate/certificate.go
@@ -14,7 +14,7 @@ func init() {
 		Description: `**step certificate** command group provides facilities for creating
 certificate signing requests (CSRs), creating self-signed certificates
 (e.g., for use as a root certificate authority), generating leaf or
-intermediate CA certificate by signing a CSR, validating certificates,
+intermediate CA certificate by signing a CSR, validating certificates, check expiration of certificate,
 renewing certificates, generating certificate bundles, and key-wrapping
 of private keys.
 
--- a/command/certificate/certificate.go
+++ b/command/certificate/certificate.go
@@ -14,7 +14,7 @@ func init() {
 		Description: `**step certificate** command group provides facilities for creating
 certificate signing requests (CSRs), creating self-signed certificates
 (e.g., for use as a root certificate authority), generating leaf or
-intermediate CA certificate by signing a CSR, validating certificates, checking certificate verdancy,
+intermediate CA certificate by signing a CSR, validating certificates,
 renewing certificates, generating certificate bundles, and key-wrapping
 of private keys.
 
@@ -87,6 +87,7 @@ $ step certificate uninstall root-ca.crt
 			inspectCommand(),
 			fingerprintCommand(),
 			lintCommand(),
+			needsRenewalCommand(),
 			signCommand(),
 			verifyCommand(),
 			keyCommand(),
--- a/command/certificate/certificate.go
+++ b/command/certificate/certificate.go
@@ -14,7 +14,7 @@ func init() {
 		Description: `**step certificate** command group provides facilities for creating
 certificate signing requests (CSRs), creating self-signed certificates
 (e.g., for use as a root certificate authority), generating leaf or
-intermediate CA certificate by signing a CSR, validating certificates, checking certificate expiration,
+intermediate CA certificate by signing a CSR, validating certificates, checking certificate verdancy,
 renewing certificates, generating certificate bundles, and key-wrapping
 of private keys.
 
--- a/command/certificate/needsRenewal.go
+++ b/command/certificate/needsRenewal.go
@@ -2,14 +2,14 @@ package certificate
 
 import (
 	"crypto/x509"
-	"encoding/pem"
-	"github.com/pkg/errors"
-	"github.com/smallstep/cli/errs"
-	"github.com/urfave/cli"
-	"io/ioutil"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/crypto/pemutil"
+	"github.com/smallstep/cli/errs"
+	"github.com/urfave/cli"
 )
 
 const defaultPercentUsedThreshold = 66
@@ -19,46 +19,55 @@ func needsRenewalCommand() cli.Command {
 		Name:      "needs-renewal",
 		Action:    cli.ActionFunc(needsRenewalAction),
 		Usage:     `Check if a certificate needs to be renewed`,
-		UsageText: `**step certificate needs-renewal** <crt_file or host_name> [**--expires-in**=<duration>]`,
-		Description: `**step certificate needs-renewal** returns '0' if the certificate needs to be renewed based on it's remaining lifetime. 
-		Returns '1' if the certificate is within it's validity lifetime bounds and does not need to be renewed. 
-		Returns '255' for any other error. By default, if a certificate "needs renewal" when it has passed 66% of it's allotted lifetime. 
+		UsageText: `**step certificate needs-renewal** <cert_file or host_name> [**--expires-in**=<duration>]`,
+		Description: `**step certificate needs-renewal** returns '0' if the certificate needs to be renewed based on it's remaining lifetime.
+		Returns '1' if the certificate is within it's validity lifetime bounds and does not need to be renewed.
+		Returns '255' for any other error. By default, a certificate "needs renewal" when it has passed 66% of it's allotted lifetime.
 		This threshold can be adjusted using the '--expires-in' flag.
+
 ## POSITIONAL ARGUMENTS
-<cert_file or hostname> The path to a certificate to validate OR a hostname with protocol prefix.
+
+<cert_file or hostname>
+:  The path to a certificate OR a hostname with protocol prefix.
 
 ## EXIT CODES
 
 This command returns '0' if the certificate needs renewal, '1' if the certificate does not need renewal, and '255' for any error.
 
 ## EXAMPLES
-Check certificate for renewal using custom directory: 
+
+Check certificate for renewal using custom directory:
 '''
-$ step certificate needs-renewal ./certificate.crt 
+$ step certificate needs-renewal ./certificate.crt
 '''
+
 Check certificate for renewal using a hostname:
 $ step certificate needs-renewal https://smallstep.com
 '''
+
 Check if certificate will expire within a given time:
 $ step certificate needs-renewal ./certificate.crt --expires-in 1h15m
 '''
+
 Check if certificate from hostname will expire within a given time:
 $ step certificate needs-renewal https://smallstep.com --expires-in 1h15m
 '''
-Check if certificate has passed a percentage of its lifetime: 
+
+Check if certificate has passed 75 percent of it's lifetime:
 $ step certificate needs-renewal ./certificate.crt --expires-in 75%
 '''
-Check if certificate from a hostname has passed a percentage of its lifetime:
+
+Check if certificate from a hostname has passed 75 percent of it's lifetime:
 $ step certificate needs-renewal https://smallstep.com --expires-in 75%
 `,
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name: "expires-in",
-				Usage: `Check if the certificate expires in given time duration
-				using <percent|duration>. With <percent>, must be followed by "%".
-				With <duration>, it is a sequence of decimal numbers, each with optional
-				fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid
-				time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
+				Usage: `Check if the certificate expires within the given time window
+using <percent|duration>. If using <percent>, the input must be followed by a "%"
+character. If using <duration>, the input must be a sequence of decimal numbers,
+each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
+Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
 			},
 		},
 	}
@@ -68,91 +77,73 @@ func needsRenewalAction(ctx *cli.Context) error {
 	if err := errs.NumberOfArguments(ctx, 1); err != nil {
 		return err
 	}
+
 	var (
+		err        error
 		crtFile    = ctx.Args().Get(0)
 		expiresIn  = ctx.String("expires-in")
 		roots      = ctx.String("roots")
 		serverName = ctx.String("servername")
 	)
 
-	var blocks []*pem.Block
-	var block *pem.Block
+	var certs []*x509.Certificate
 	if addr, isURL, err := trimURL(crtFile); err != nil {
 		return errs.NewExitError(err, 255)
 	} else if isURL {
-		peerCertificates, err := getPeerCertificates(addr, serverName, roots, false)
+		certs, err = getPeerCertificates(addr, serverName, roots, false)
 		if err != nil {
 			return errs.NewExitError(err, 255)
 		}
-		for _, crt := range peerCertificates {
-			blocks = append(blocks, &pem.Block{
-				Type:  "CERTIFICATE",
-				Bytes: crt.Raw,
-			})
-		}
-
 	} else {
-		crtBytes, err := ioutil.ReadFile(crtFile)
+		certs, err = pemutil.ReadCertificateBundle(crtFile)
 		if err != nil {
 			return errs.NewExitError(err, 255)
 		}
-
-		// The first certificate PEM in the file is our leaf Certificate.
-		// Any certificate after the first is added to the list of Intermediate
-		// certificates used for path validation.
-		for len(crtBytes) > 0 {
-			block, crtBytes = pem.Decode(crtBytes)
-			if block == nil {
-				return errs.NewExitError(errors.Errorf("%s contains an invalid PEM block", crtFile), 255)
-			}
-			if block.Type != "CERTIFICATE" {
-				continue
-			}
-
-			blocks = append(blocks, block)
-		}
-		if block == nil {
-			return errs.NewExitError(errors.Errorf("%s contains no PEM certificate blocks", crtFile), 255)
-		}
-
 	}
 
-	for _, block := range blocks {
-		cert, err := x509.ParseCertificate(block.Bytes)
-		if err != nil {
-			return errs.NewExitError(errors.WithStack(err), 255)
+	var (
+		percentThreshold int
+		duration         time.Duration
+		isPercent        = expiresIn == "" || strings.HasSuffix(expiresIn, "%")
+	)
+
+	if isPercent {
+		var percentThreshold int
+		if expiresIn == "" {
+			percentThreshold = defaultPercentUsedThreshold
+		} else {
+			percentThreshold, err = strconv.Atoi(strings.TrimSuffix(expiresIn, "%"))
+			if err != nil {
+				return errs.NewExitError(err, 255)
+			}
 		}
-		var remainingValidity = time.Until(cert.NotAfter)
-		var totalValidity = cert.NotAfter.Sub(cert.NotBefore)
-		var percentUsed = (1 - remainingValidity.Minutes()/totalValidity.Minutes()) * 100
+		if percentThreshold > 100 || percentThreshold < 0 {
+			return errs.NewExitError(errors.Errorf("Percentage must be in range 0-100"), 255)
+		}
+	} else {
+		duration, err = time.ParseDuration(expiresIn)
+		if err != nil {
+			return errs.NewExitError(err, 255)
+		}
+	}
 
-		if expiresIn == "" || strings.Contains(expiresIn, "%") {
-			var threshold int
-			if expiresIn == "" {
-				threshold = defaultPercentUsedThreshold
-			} else {
-				threshold, err = strconv.Atoi(strings.TrimSuffix(expiresIn, "%"))
-				if err != nil {
-					return errs.NewExitError(err, 255)
-				}
-			}
-			if threshold > 100 || threshold < 0 {
-				return errs.NewExitError(errors.Errorf("Percentage must be in range 0-100"), 255)
-			}
+	for _, cert := range certs {
 
-			if int(percentUsed) >= threshold {
+		remainingValidity := time.Until(cert.NotAfter)
+
+		if isPercent {
+			totalValidity := cert.NotAfter.Sub(cert.NotBefore)
+			percentUsed := (1 - remainingValidity.Minutes()/totalValidity.Minutes()) * 100
+
+			if int(percentUsed) >= percentThreshold {
 				return nil
 			}
 		} else {
-			duration, err := time.ParseDuration(expiresIn)
-
-			if err != nil {
-				return errs.NewExitError(err, 255)
-			} else if duration.Minutes() > remainingValidity.Minutes() {
+			if duration >= remainingValidity {
 				return nil
 			}
 		}
 	}
 
-	return errs.NewExitError(errors.Errorf("Certificate does not need renewal"), 1)
+	return errs.NewExitError(errors.Errorf("certificate does not need renewal"), 1)
 }
--- a/command/certificate/needsRenewal.go
+++ b/command/certificate/needsRenewal.go
@@ -1,0 +1,163 @@
+package certificate
+
+import (
+	"crypto/x509"
+	"encoding/pem"
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/errs"
+	"github.com/urfave/cli"
+	"io/ioutil"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+)
+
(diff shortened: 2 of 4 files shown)


command/certificate/certificate.go
```diff
@@ -14,7 +14,7 @@ func init() {
 		Description: `**step certificate** command group provides facilities for creating
 certificate signing requests (CSRs), creating self-signed certificates
 (e.g., for use as a root certificate authority), generating leaf or
-intermediate CA certificate by signing a CSR, validating certificates,
+intermediate CA certificate by signing a CSR, validating certificates, check expiration of certificate,
```
dopey, line 17: `checking certificate expiration` 

The tense on the other verbs all seem to end with `ing`, so we should be consistent.

command/certificate/verify.go
```diff
@@ -164,6 +177,30 @@ func verifyAction(ctx *cli.Context) error {
...
+		} else if percentIntoLifeTime > 90 {
+			fmt.Println("\033[31m", "Leaf is", int(percentIntoLifeTime), "% through its lifetime.", "\033[0m")
+		} else if percentIntoLifeTime > 66 && percentIntoLifeTime < 90 {
+			fmt.Println("\033[33m", "Leaf is", int(percentIntoLifeTime), "% through its lifetime.", "\033[0m")
+		} else if percentIntoLifeTime < 66 && percentIntoLifeTime > 1 {
+			fmt.Println("\033[32m", "Leaf is", int(percentIntoLifeTime), "% through its lifetime.", "\033[0m")
+		} else if percentIntoLifeTime < 1 {
+			fmt.Println("\033[32m", "Leaf is less than 1% through its lifetime.", "\033[0m")
+		} else {
+			fmt.Println("Error")
```
dopey, line 198: return an `errors.Errorf` error here that described the issue. See errors returned in above code for examples.

command/certificate/verify.go
```diff
@@ -164,6 +177,30 @@ func verifyAction(ctx *cli.Context) error {
...
 
+	if expire {
+
+		NowTillEndOfCert := time.Until(cert.NotAfter)
+		totalLifeTimeOfCert := cert.NotAfter.Sub(cert.NotBefore)
+
+		percentIntoLifeTime := ((totalLifeTimeOfCert.Hours() - NowTillEndOfCert.Hours()) / totalLifeTimeOfCert.Hours()) * 100
+
+		if percentIntoLifeTime >= 100 {
+			fmt.Println("\033[31m", "This certificate has already expired.", "\033[0m") //"\033[__m" are color codes
```
dopey, line 189: Just gonna list some bullet points here for improvements to this section: 

```
if percent >= 100 {
}
else if percent >= 90 {
}
else if percent >= 66 {
}
else if percent >= 1 {
}
else if precent >= 0 {
} else {
}
```
* Use `fmt.Printf` instead of `fmt.Println` if possible.
* parameterize commonly used strings and colors so you can reuse them. e.g `var colorExpired = "31m"`.

command/certificate/verify.go
```diff
@@ -164,6 +177,30 @@ func verifyAction(ctx *cli.Context) error {
 		}
 	}
 
+	if expire {
+
+		NowTillEndOfCert := time.Until(cert.NotAfter)
```
dopey, line 183: variable names with capital first letters are exported. Use a lower case first letter unless you need to export the variable beyond the current package.

command/certificate/verify.go
```diff
@@ -164,6 +177,30 @@ func verifyAction(ctx *cli.Context) error {
...
+		totalLifeTimeOfCert := cert.NotAfter.Sub(cert.NotBefore)
+
+		percentIntoLifeTime := ((totalLifeTimeOfCert.Hours() - NowTillEndOfCert.Hours()) / totalLifeTimeOfCert.Hours()) * 100
```
dopey, line 186: Why are we specifying `.Hours()` here rather than using the complete duration?

command/certificate/verify.go
```diff
@@ -164,6 +177,30 @@ func verifyAction(ctx *cli.Context) error {
 		}
 	}
 
+	if expire {
+
+		NowTillEndOfCert := time.Until(cert.NotAfter)
+		totalLifeTimeOfCert := cert.NotAfter.Sub(cert.NotBefore)
```
dopey, line 184: I would check first if `time.Now().Before(cert.NotBefore)` because if so, then the cert is not valid yet, and that's a different message than any of the ones you have below. At the same time I would check if `time.Now().After(cert.NotAfter)` just to rule out both of those special cases. Then we know that the results should be within our bounds.

I would also do `(1 - remainder/totalValidity) = `

command/certificate/verify.go
dopey: use `remainingValidity` and `totalValidity`. The term validity window is already associated with certificates and we already frequently use the work `validity` in our code when doing similar calculations.

command/certificate/verify.go
dopey: You can just call this variable percent.

Generally speaking, the rule for variable names is that the length of the name should be proportionate to how far away from the instantiation of the variable it is being used. So if it's super far away, e.g. different functions, files or packages, then you want a really descriptive name like you've chosen here. But in this case, the variable and it's use are super close. So we can opt for shorter, less descriptive, more context dependent names. Lmk if that does/doesn't make sense.

command/certificate/verify.go
```diff
@@ -65,12 +68,22 @@ Verify a certificate using a custom directory of root certificates for path vali
 '''
 $ step certificate verify ./certificate.crt --roots ./root-certificates/
 '''
+
+Verify the expiration time left of a certificate using a custom root certificate and host for path validation:
```
dopey, line 72: Verify the remaining validity of a certificate ...

command/certificate/verify.go
```diff
@@ -65,12 +68,22 @@ Verify a certificate using a custom directory of root certificates for path vali
...
+'''
 `,
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "host",
 				Usage: `Check whether the certificate is for the specified host.`,
 			},
+			cli.BoolFlag{
+				Name:  "expire",
+				Usage: `Checks the certificate time till expiration`,
```
dopey, line 85: Checks the remaining certificate validity till expiration

command/certificate/verify.go
```diff
@@ -164,6 +178,35 @@ func verifyAction(ctx *cli.Context) error {
...
+			fmt.Printf(green,"Leaf is", int(percentUsed), "% through its lifetime.", reset)
+		} else if percentUsed < 1{
+			fmt.Printf(green,"Leaf is less than 1% through its lifetime.", reset)
+		} else {
+			return errors.Errorf("failure to determine expiration time for certificate")
+		}
+
+		return nil
+	}
+
```
dopey, line 209: I think we should verify the certificate first `cert.VerifyOpts` below, and then if it's valid we can analyze how much of the lifespan has been used. The benefit of this is that we won't have to check if now < cert.NotBefore or if the cert is already expired. The `VerifyOpts` method already does that for us. So only if the certificate is valid do we print expiry information.

command/certificate/verify.go
```diff
@@ -164,6 +178,35 @@ func verifyAction(ctx *cli.Context) error {
...
+		red := "\033[31m"
+		green := "\033[32m"
+		yellow := "\033[33m"
+		reset := "\033[0m"
+		
+		if percentUsed >= 100 {
+			fmt.Println(red, "This certificate has already expired.", reset)
+		} else if percentUsed > 90 {
+			fmt.Printf(red,"Leaf is", int(percentUsed), "% through its lifetime.", reset)
```
dopey, line 196: When you use `fmt.Printf` you can do things like: `fmt.Printf("%sLeaf is %d% through it's lifetime.", red, percentUsed)`. The variables  at the end replace the `%s` and `%d`.

command/certificate/verify.go
```diff
@@ -164,6 +178,35 @@ func verifyAction(ctx *cli.Context) error {
...
+
+		percentUsed := ((totalValidity.Hours() - remainingValidity.Hours()) / totalValidity.Hours()) * 100
+
+		red := "\033[31m"
+		green := "\033[32m"
+		yellow := "\033[33m"
+		reset := "\033[0m"
+		
+		if percentUsed >= 100 {
+			fmt.Println(red, "This certificate has already expired.", reset)
```
dopey, line 194: We're using the word `certificate` in this output, but the word `leaf` in future outputs to refer to the same thing. I think we should consistently use the term `certificate`.

dopey, line 194: @tashian How would you like this output to look? Your plan was to use this as input from a pipe. I recall you saying that you wanted these to be errorCodes rather than printed statements.

command/certificate/verify.go
```diff
@@ -65,12 +68,22 @@ Verify a certificate using a custom directory of root certificates for path vali
 '''
 $ step certificate verify ./certificate.crt --roots ./root-certificates/
 '''
+
+Verify the remaining validity of a certificate using a custom root certificate and host for path validation:
+
+'''
+$ step certificate verify ./certificate.crt --host smallstep.com --expire
```
dopey, line 75: @mike @tashian what did we want this flag to be called? `expire`, `validity`, `verdancy`?

tashian: `verdancy`

command/certificate/verify.go
tashian: Yes, the error codes are described in #398

tashian:
hi @cookie-annihilator @tommy-56 I'm excited to see this PR!
I want to explain why `--verdancy` is the word we landed on for this flag during design discussions.
The idea of verdancy is, it's a matter of degree: What the command is asking is "Just how green is this certificate?"
And, like a leaf on a tree, as the certificate ages it changes color.
Whereas, `--expire` implies a yes/no answer: Is it expired or not?
Verdancy provides that answer and more.

The use cases are:
- Verdancy aids automated renewal, and once this is merged I will immediately put it into use in our docs. We recommend people renew certificates when they are around 2/3 through their validity period, so we want an easy way to confirm that.
- Expired (yes/no) could also be used in a script, if there is a chance a certificate may have expired. Because with an unexpired cert you can run `step ca renew` to get a new certificate, but with an expired cert you'd need to run `step ca certificate`.
 - We also discussed an `expires-in` flag that would allow you to pass a % or duration, and you'd get back a yes/no.

Happy to answer any other questions y'all have about this feature...

mmalone:
New proposal: `step certificate verdancy` with an `--expires-in` flag that returns `0/1`.

The `--expires-in` flag should take a percentage of the certificate's lifetime or something parseable by [`time.ParseDuration`](https://golang.org/pkg/time/#ParseDuration).

Let's have a separate `step certificate expired` subcommand to test whether the certificate has already expired. We may want a second flag on `step certificate verdancy` to control the behavior when the certificate is expired or if the current time is < not-before. But it's not clear that's needed since `step certificate expired` would allow you to distinguish between a certificate that's "nearing expiry but not yet expired" and one that's already expired.

This seems better / easier / more flexible than a command with fixed error codes for fixed percentages-to-expiry.

dopey:
* return 0 if it's a certificate that needs to be renewed
* return 1 if it's a certificate that is valid (timewise) but doesn't need to be renewed
* return 255 for any error 
* default 66% expires-in
* In the docs mention that this is designed for use in systemd (or timers/chrons in general)
* What if the certificate is expired?
  * We want to allow renew-after-expiry down the road.
  * If we also add `step certificate expired` then you could customize behavior by composing commands. 
* Command Name? `needs-renewal`, `renewal-ready`.
  * `needs-renewal`

maraino:
> return 0 if it's a certificate that needs to be renewed
> return 0 if the certificate is expired
> return 1 if it's a certificate that is valid (timewise) but doesn't need to be renewed
> return 255 for any error

Currently, 0 means success, and 1 error. In my opinion, to be more consistent I think we should do:
* 0 if it's a certificate that needs to be renewed (or is expired?).
* 1 if there's an error, wrapping the error and returning an error (instead of doing os.Exit(1) 
* And return 2 if the certificate is still valid (os.Exit(2)).

tommy-56:
My understanding was @tashian was using it for systemd which he was saying takes in only 255 exit code as an error happening which was why we went with 255 exit code and changed the 1/0 meaning to fit the systemd requirements

maraino:
@tashian systemd seems to define 1 as the default failure code https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Exit%20Codes

tashian:
@mariano yes, for `ExecStart` it does. But for `ExecCondition` statements, where I'm expecting to use this feature, exit codes are treated a little differently:
* 0 means continue (eg. renew the cert)
* 1-254 means stop
* 255 is a hard failure (is logged at a higher log level)

tashian:
btw I'm not averse to a `--systemd` flag just because this is non-standard

maraino:
@tashian if that's the case I'd like to have a custom error type that is detected here
https://github.com/smallstep/cli/blob/3dbf944429d9234fc854a7075725f434ff991c8c/cmd/step/main.go#L97-L113

Exiting with 255, and always returns errors in the command methods instead of doing os.Exit(255)

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+	"strings"
+	"time"
+)
+
+func needsRenewalCommand() cli.Command {
+	return cli.Command{
+		Name:      "needs-renewal",
+		Action:    cli.ActionFunc(needsRenewalAction),
+		Usage:     `Check if a certificate needs to be renewed`,
+		UsageText: `**step certificate needs-renewal** <crt_file or host_name> [**--expires-in <duration>]`,
```
dopey, line 21: Should be `[**--expires-in**=<duration>]` -- see here: https://github.com/smallstep/cli/blob/master/command/ca/certificate.go#L24.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+		Description: `**step certificate needs-renewal** Checks certificate expiration from file or from a host if
```
dopey, line 22: **step certificate needs-renewal** returns '0' if the certificate needs to be renewed based on it's remaining lifetime. Returns '1' if the certificate is within it's validity lifetime bounds and does not need to be renewed. Returns '255' for any other error. By default, if a certificate "needs renewal" when it has passed 66% of it's allotted lifetime. This threshold can be adjusted using the '--expires-in' flag.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+		Action:    cli.ActionFunc(needsRenewalAction),
+		Usage:     `Check if a certificate needs to be renewed`,
+		UsageText: `**step certificate needs-renewal** <crt_file or host_name> [**--expires-in <duration>]`,
+		Description: `**step certificate needs-renewal** Checks certificate expiration from file or from a host if 
+		the certificate is over 66% of its lifetime. If the certificate needs renewal this command will return '0'.
+		If the certificate is not far enough into its life, it will return '1'. If validation fails, or if an error occurs, 
+		this command will produce a non-zero return value.
+## POSITIONAL ARGUMENTS
+
+<crt_file>
```
dopey, line 28: Combine the positional arguments into one, as we do with the UsageText above.

`<cert_file or hostname> The path to a certificate to validate OR a hostname with protocol prefix.`

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+: The path to a certificate to validate.
+
+<host_name>
+: Address of remote host 
+
+## EXIT CODES
+
+This command returns 0 if needing renewal, or returns 1 if the certificate does not need renewal. It will return 255 if any errors occurred
```
dopey, line 36: `return '0' if the certificate needs renewal, '1' if the certificate does not need renewal, and '255' for any other error.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+
+## EXAMPLES
+Check certificate for renewal using custom directory 
+'''
+$ step certificate needs-renewal ./certificate.crt 
+'''
+Check certificate for renewal using a host
```
dopey, line 43: using a hostname:

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+: The path to a certificate to validate.
+
+<host_name>
+: Address of remote host 
+
+## EXIT CODES
+
+This command returns 0 if needing renewal, or returns 1 if the certificate does not need renewal. It will return 255 if any errors occurred
+
+## EXAMPLES
```
dopey, line 38: Put a colon `:` after each sentence prefacing the example command, like so: https://github.com/smallstep/cli/blob/master/command/certificate/verify.go#L38

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+'''
+Check if certificate from a host has passed a percentage of its lifetime
+$ step certificate needs-renewal https://smallstep.com --expires-in 75%
+`,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name: "expires-in",
+				Usage: `Check if the certificate expires in given time duration
+				using <percent|duration>. For <percent>, must be followed by "%".
+				With <duration>, it is a sequence of decimal numbers, each with optional
```
dopey, line 63: With -> For

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+	var (
+		crtFile    = ctx.Args().Get(0)
+		expiresIn  = ctx.String("expires-in")
+		roots      = ctx.String("roots")
+		serverName = ctx.String("servername")
+		cert       *x509.Certificate
+	)
+
+	if addr, isURL, err := trimURL(crtFile); err != nil {
+		return err
```
dopey, line 84: this should return 255 as well

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+	} else if isURL {
+		peerCertificates, err := getPeerCertificates(addr, serverName, roots, false)
```
dopey, line 86: Since we're changing the operation of this command to now apply to all certificates in the bundle, you can probably just use the full list of peer certificates rather than grabbing the first index.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+		crtBytes, err := ioutil.ReadFile(crtFile)
+		if err != nil {
+			return errs.NewExitError(err, 255)
+		}
+
+		var block *pem.Block
+		// The first certificate PEM in the file is our leaf Certificate.
+		// Any certificate after the first is added to the list of Intermediate
+		// certificates used for path validation.
+		for len(crtBytes) > 0 {
```
dopey, line 102: Since we're changing the behavior of this command to apply to all certs in the file, what we want to do is loop over each pem block in the file (just like you're doing) and then convert it to a certificate and `append` it to an array of x509 certifcates.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+					return errs.NewExitError(errors.WithStack(err), 255)
+				}
+			}
+		}
+		if cert == nil {
+			return errs.NewExitError(errors.Errorf("%s contains no PEM certificate blocks", crtFile), 255)
+		}
+
+	}
+	var remainingValidity = time.Until(cert.NotAfter)
```
dopey, line 122: This will happen in a for loop over all the certificates in the list

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+	var totalValidity = cert.NotAfter.Sub(cert.NotBefore)
+	var percentUsed = (1 - remainingValidity.Minutes()/totalValidity.Minutes()) * 100
+
+	if expiresIn != "" {
+		if strings.Contains(expiresIn, "%") {
+			percentageInput, err := strconv.Atoi(strings.ReplaceAll(expiresIn, "%", ""))
```
dopey, line 128: I would say just trimSuffix instead of replace all. We want people entering 50% instead of %3%5

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+
+			if err != nil {
+				return errs.NewExitError(err, 255)
+			} else if duration.Minutes() > remainingValidity.Minutes() {
+				return nil
+			}
+			os.Exit(1)
+		}
+	} else {
+		if percentUsed >= 66 {
```
dopey, line 153: Let's parameterize this value, meaning let's make it a constant defined at the top of the file. 

```
const defaultPercentUsedThreshold = 66
```

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+
+			if err != nil {
+				return errs.NewExitError(err, 255)
+			}
+			if percentageInput > 100 || percentageInput < 0 {
+				return errs.NewExitError(errors.Errorf("Percentage must be in range 0-100"), 255)
+			}
+
+			if percentageInput > int(percentUsed) {
+				return nil
```
dopey, line 138: I think this is the case where we want to return 1.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,163 @@
...
+			if err != nil {
+				return errs.NewExitError(err, 255)
+			} else if duration.Minutes() > remainingValidity.Minutes() {
+				return nil
+			}
+			os.Exit(1)
+		}
+	} else {
+		if percentUsed >= 66 {
+			return nil
```
dopey, line 154: It would be nice to combine this with the other percentage case so we don't copy the same conditional code.

Something like: 

```
for _, cert := range certs {
if expiredIn == "" || strings.Contains(expiresIn, "%") {
    var threshold int
    if expiresIn == "" {
        threshold = defaultPercentThreshold
    } else {
        threshold, err = strconv.Atoi(strings.TrimSuffix(expiresIn, "%"))
        if err ....
    }
    
    if threshold > 100 || threshold < 0 ...

    if int(percentUsed) >= threshold ...
        return nil
} else {
    			duration, err := time.ParseDuration(expiresIn)

			if err != nil {
				return errs.NewExitError(err, 255)
			} else if duration > remainingValidity {
				return nil
			}
}
}
// We only get here if the certs are not expiring, so return an error to that effect.
return errors.New("certificate is verdant")

**Review by dopey — CHANGES_REQUESTED** [review]
left a bunch of comments.

File: command/certificate/needsRenewal.go [hunk 28]
```diff
@@ -7,59 +7,55 @@ import (
...
 	return cli.Command{
 		Name:      "needs-renewal",
 		Action:    cli.ActionFunc(needsRenewalAction),
 		Usage:     `Check if a certificate needs to be renewed`,
-		UsageText: `**step certificate needs-renewal** <crt_file or host_name> [**--expires-in <duration>]`,
-		Description: `**step certificate needs-renewal** Checks certificate expiration from file or from a host if 
-		the certificate is over 66% of its lifetime. If the certificate needs renewal this command will return '0'.
-		If the certificate is not far enough into its life, it will return '1'. If validation fails, or if an error occurs, 
-		this command will produce a non-zero return value.
+		UsageText: `**step certificate needs-renewal** <crt_file or host_name> [**--expires-in**=<duration>]`,
```
dopey, line 22: Use `cert_file` here and in other places where you refer to this input. Over time we came to the custom of using `cert` if it's in documentation that the user sees because that's the standard other popular tools use (e.g. openssl, curl), and `crt` if we're just naming variables internally in our code.

command/certificate/needsRenewal.go
```diff
@@ -7,59 +7,55 @@ import (
...
+		Description: `**step certificate needs-renewal** returns '0' if the certificate needs to be renewed based on it's remaining lifetime. 
+		Returns '1' if the certificate is within it's validity lifetime bounds and does not need to be renewed. 
+		Returns '255' for any other error. By default, if a certificate "needs renewal" when it has passed 66% of it's allotted lifetime.
```
dopey, line 25: remove the `if` in this line.

command/certificate/needsRenewal.go
```diff
@@ -7,59 +7,55 @@ import (
...
+		Returns '255' for any other error. By default, if a certificate "needs renewal" when it has passed 66% of it's allotted lifetime. 
+		This threshold can be adjusted using the '--expires-in' flag.
 ## POSITIONAL ARGUMENTS
-
-<crt_file>
-: The path to a certificate to validate.
-
-<host_name>
-: Address of remote host 
+<cert_file or hostname> The path to a certificate to validate OR a hostname with protocol prefix.
```
dopey, line 28: remove `to validate`

tashian:
Thank you @tommy-56 @cookie-annihilator for this!

I've updated our systemd stuff to use `needs-renewal` (see smallstep/certificates#653 and smallstep/docs#105). It's so much cleaner, and it removes the dependency on `jq`.

PR #496, opened 2021-06-24
Needs-Renewal

tommy-56:
Referencing issue #398 
step certificate needs-renewal subcommand using exit codes 
And having optional flag --expires-in 
Example:
step certificate needs-renewal foo.crt 
step certificate needs-renewal foo.crt --expires-in 15m

step certificate needs-renewal https://smallstep.com
step certificate needs-renewal  https://smallstep.com --expires-in 15m


command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+## EXIT CODES
+
+This command returns 0 if needing renewal, or returns 1 if the certificate does not need renewal. It will return 255 if any errors occurred
+
+## EXAMPLES
+Check certificate for renewal using custom directory 
+'''
+$ step certificate needs-renewal ./certificate.crt 
+'''
+Check if certificate will expire within a given time
```
dopey, line 40: We should add an example using a host and one using percentage (rather than duration)

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+	"os"
+	"time"
+)
+
+func needsRenewalCommand() cli.Command {
+	return cli.Command{
+		Name:      "needs-renewal",
+		Action:    cli.ActionFunc(needsRenewalAction),
+		Usage:     `Check if a certificate needs to be renewed`,
+		UsageText: `**step certificate needs-renewal** <crt_file> or <host_name> [**--expires-in <duration>]`,
```
dopey, line 18: We don't really do this `<crt_file> or <host_name>` in other places. Which is not to say that we shouldn't be doing that. 

But I think instead let's try `<crt_file or host_name>` to make it clear that's it’s one argument.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+$ step certificate needs-renewal ./certificate.crt 
+'''
+Check if certificate will expire within a given time 
+'''
+$ step certificate needs-renewal ./certificate.crt --expires-in 1h15m
+'''
+`,
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "expires-in",
```
dopey, line 47: See the description of the `not-before` flag in `command/certificate/create.go`. Use the second part of that description that talks about duration for this flag as well. And additionally, I think we want the option to use a percent for this flag. So if I pass `70%` it should be parsed as a percent. Let's try to make that clear in the description.

So for example -- the beginning of `not-before` description starts with `<time|duration>`. This description could start with `<percent|duration>`.

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+		serverName = ctx.String("servername")
+		cert       *x509.Certificate
+	)
+
+	if addr, isURL, err := trimURL(crtFile); err != nil {
+		return err
+	} else if isURL {
+		peerCertificates, err := getPeerCertificates(addr, serverName, roots, false)
+		if err != nil {
+			os.Exit(255)
```
dopey, line 71: Returning `255` is technically correct. But we want a bit more than that. What we really want is to return an error type that has an attribute where we can store what the return code should be. And then we need to be able to parse that error type wherever this error is handled.

Think of yourself as the user of the program. You run it and all you see is exit code `255`. Well, a bunch of things probably cause that error code, so how do you figure out what went wrong. That's why we return error types. So there's some context for the user as to what went wrong.

Unfortunately, I'm not sure if we have an error type that you could use for this. We may have to write a new one. @mariano do we return non zero or 1 exit status from anywhere else? If not, should we create a new error type and then check for it in main.go?

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+		if cert == nil {
+			os.Exit(255)
+		}
+
+	}
+	var remainingValidity = time.Until(cert.NotAfter)
+	var totalValidity = cert.NotAfter.Sub(cert.NotBefore)
+	var percentUsed = (1 - remainingValidity.Minutes()/totalValidity.Minutes()) * 100
+
+	if expiresIn != "" {
```
dopey, line 111: I believe we want `--expires-in` to be either a percent or a duration (https://github.com/smallstep/cli/pull/495#issuecomment-867046184). In order to do that we can try the following logic:

1) Check if input string has a `%` sign as it's last index. If it does, then treat the input as a percent. The prefix of the input should be a number in a string. Golang has utilites to parse a int from a string. Parse it and then verify that it is between 0 and 100. If not we'll return an error.
2) The input does not have a `%` sign and therefore is a duration. Your code already looks correct for this case.

Question for @mmalone: How exactly is the % option supposed to work (it's not as easy to understand as the duration option)? E.g. 

```
step certificate needs-renewal --expires-in 15m
```
Answers the question: Will this certificate be expired in 15 minutes? 

```
step certificate needs-renewal --expires-in 75%
```
Do we want to know if the certificate will expire if 75% of it's lifetime passes? If it's already lived past 75% of it's lifetime. It's not very clear.

command/certificate/needsRenewal.go
maraino: @dopey are we naming this needs-renewal instead of verdancy?

command/certificate/needsRenewal.go
maraino: We usually just return a wrapped error in this cases. Commands returning an error will print the error and exit with 1. And we can get the trace using `STEPDEBUG=1`

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+		peerCertificates, err := getPeerCertificates(addr, serverName, roots, false)
+		if err != nil {
+			os.Exit(255)
+		}
+		cert = peerCertificates[0]
+
+	} else {
+		crtBytes, err := ioutil.ReadFile(crtFile)
+		if err != nil {
+			os.Exit(255)
```
maraino, line 78: We should return an error here too

maraino: We should return wrapped errors here and in all the exit conditions. For example this should be like:
```
crtBytes, err := ioutil.ReadFile(crtFile)
if err != nil {
    return errors.Wrapf(err, "error reading %s", crtFile)
}
```
Or use the helper `errs.FileError(err, crtFile)`

command/certificate/needsRenewal.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		var (
+			block *pem.Block
```
maraino, line 82: To define one variable use just `var block *pem.Block`. It looks more clean to me.

command/certificate/needsRenewal.go
tommy-56: We were going for this functionality I believe:

- return 0 if it's a certificate that needs to be renewed
- return 0 if the certificate is expired
- return 1 if it's a certificate that is valid (timewise) but doesn't need to be renewed
- return 255 for any error

Which is why we needed a different error type that would give 255 as an exit code
os.Exit(255) was there in the mean time to make sure the logic worked right until we knew if there was an error type built like that. 
For cases where exit code 1 is needed should we continue to use os.Exit(1) or should we change it to some kind of error type which doesn't seem right if its not actually an error?

maraino: See https://github.com/smallstep/cli/pull/495#issuecomment-870048089, I think we should be consistent and keep `1` as exit code for an error. 

In bash scripts is very easy to differentiate exit errors. Unless systemd is an issue we should keep `1` as an error and use `2` for example to say that the certificate is valid.

Issue #470, opened 2021-04-28
`step crypto key public` should be able to extract the public key from a cert.


dopey:
You can already use `step certificate key`, but I feel that `crypto key public` should work on higher order crypto structures as well. Especially since we already have the working code.

cookie-annihilator:
issue fixed and merged with #497

PR #497, opened 2021-07-02
add functionality to check a certificates public key

cookie-annihilator:
fixes issue #470 - now can check the public key from a certificate or a key, flag function works with it as well.
ex:
step crypto key public foo.crt
step crypto key public --out foo.pem foo.crt

### DIFF
--- a/command/crypto/key/public.go
+++ b/command/crypto/key/public.go
@@ -1,29 +1,29 @@
 package key
 
 import (
-        "bytes"
-        "crypto"
-        "encoding/pem"
-        "fmt"
-        "os"
+	"bytes"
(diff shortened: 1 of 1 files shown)


command/crypto/key/public.go
```diff
@@ -23,19 +25,18 @@ func publicCommand() cli.Command {
...
 :  Path to a private key.
-
 ## EXAMPLES
-
 Print the corresponding public key:
 '''
 $ step crypto key public priv.pem
 '''
-
+Print the public key of certificate:
```
dopey, line 36: of an x509 certificate

Issue #360, opened 2020-09-28
`step certificate inspect` fails with "Invalid PEM block"


polarathene:
### Subject of the issue

I used `step certificate inspect /path/to/cert.pem` but received an error `<cert-path> contains an invalid PEM block`.

### Your environment
* OS - Docker (Alpine) - Official smallstep/step-ca image
* Version - Smallstep CLI/0.15.2-dev (linux/amd64) - Release Date: 2020-08-31 21:38 UTC

### Steps to reproduce

The error was being triggered because the certificate produced (third-party acme client) a full chain with some comment/label to identify each blocks cert in a human friendly way for inspection. `openssl verify` throws no errors about this, and it appears to be a common practice, but not an official standard.

`step certificate inspect` is actually ok with this too, **provided** there isn't anything before the initial `-----BEGIN CERTIFICATE-----` line. That seems to be what throws it off, so reporting it here as potential bug.

```pem
example.test - leaf certificate - timestamp etc
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----

Smallstep Intermediate CA
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----

Smallstep Root CA
-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
```

### Expected behaviour

It should not complain about an invalid PEM block when the certificate blocks themselves are perfectly fine to parse.

### Actual behaviour

The error is due to text content outside of them and only prior to the first certificate block. It doesn't seem to matter what the contents is, even a blank line is enough to cause the error.

polarathene:
Still experiencing this issue, 15.3 on Arch Linux:

```
Smallstep CLI/0.15.3 (linux/amd64)
Release Date: 2020-10-21 23:46 UTC
```

Is there anything else I can provide that would be helpful?

maraino:
The problem is this else condition https://github.com/smallstep/cli/blob/da65f7c1b773e69ccdd1c1cb65b30a71380d63da/command/certificate/inspect.go#L235-L256

We need to do a logic similar to Go's `x509.CertPool.AppendCertsFromPEM()` where data that is not in a PEM block is skipped.

cookie-annihilator:
Issue fixed with merged #503 to only read the PEM blocks

PR #500, opened 2021-07-08
only read PEM block

cookie-annihilator:
fixes issue #360 by trimming out anything extra so that it will only read the PEM block


command/certificate/inspect.go
```diff
@@ -31,106 +31,72 @@ or CSR in a human readable format. Output from the inspect command is printed to
 STDERR instead of STDOUT. This is an intentional barrier to accidental
 misuse: scripts should never rely on the contents of an unvalidated certificate.
 For scripting purposes, use **step certificate verify**.
-
```
dopey, line 34: Why are we deleting all these newlines?

cookie-annihilator:
latest commit fixes the code back to before the accidental merge

PR #503, opened 2021-07-12
Only read PEM blocks

cookie-annihilator:
fixes issue #360 by trimming out anything extra so that it will only read the PEM block


command/certificate/inspect.go
```diff
@@ -48,89 +47,75 @@ This command returns 0 on success and \>0 if any error occurs.
...
 '''
 
 Inspect a remote certificate chain in PEM format:
-
 '''
 $ step certificate inspect https://smallstep.com --format pem --bundle
 '''
 
 Inspect a local CSR in text format (default):
-
```
dopey, line 127: These are the only whitespaces that we should remove. Whitespaces between any other types of documentation lines should remain.

PR #504, opened 2021-07-13
Allow more url variations in step certificate inspect

### LINKED ISSUE #275: Allow more url variations in "certificate inspect"
Allow more url variations in "step certificate inspect".
e.g.
```
// this works
step certificate inspect https://www.google.com

// these don't
step certificate inspect https://www.google.com/
step certificate inspect www.google.com:443
step certificate inspect www.google.com
```

### Why this is needed
More convenience

tommy-56:
Closes #275. The inspect command now allows www.example.com and www.example.com:443 to be given as a URL to inspect.

### DIFF
--- a/command/certificate/inspect.go
+++ b/command/certificate/inspect.go
@@ -196,7 +196,7 @@ func inspectAction(ctx *cli.Context) error {
 	var block *pem.Block
 	var blocks []*pem.Block
 	//check if address is www.example.com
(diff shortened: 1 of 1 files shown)


dopey:
My only question here is, what if you have a file named `www.foo.bar.crt` that you want to inspect? @mmalone?
I thought that's why we avoided making this change when we addressed this issue originally.

PR #511, opened 2021-07-21
Clarify the use of the --redirect-url flag

### LINKED ISSUE #510: redirect-url doesnt work
If I run a command like this:

~~~
step oauth --redirect-url=urn:ietf:wg:oauth:2.0:oob
~~~

I still get the localhost url:

~~~
https://accounts.google.com/o/oauth2/v2/auth?
client_id=1087160488420-8qt7bavg3qesdhs6it824mhnfgcfe8il.apps.googleusercontent.com&
code_challenge=Uk_xqM911XAjUuG4kIAFFRfNRnYi2ciYdFU7rlo2NoI&
code_challenge_method=S256&
nonce=9ab7adfa13a2e8bfd6a0c58fe554549d1a084c3fd333f6c0c63924327727006d&
redirect_uri=http%3A%2F%2F127.0.0.1%3A58340&response_type=code&
scope=openid+email&state=DvmigvUDZDziYcInEEwRL1x9mXwZLPON
~~~

https://smallstep.com/docs/step-cli/reference/oauth

maraino:
### Description 

This PR changes the usage of the --redirect-url. This flag is used in the `step oauth` command and can also be used in the `step ca bootstrap` command to write it in the defaults.json file.

This flag won't have any effect on oob flows, but we cannot fail if both are used because it will cause errors on specific configurations where errors are not expected.

Fixes #510

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -272,10 +272,11 @@ be stored in the 'sshpop' header.`,
 the **--team** option. If the url contains <\<\>> placeholders, they are replaced with the team ID.`,
 	}
 
-	// RedirectURL is a cli.Flag used to pass the OAuth redirect URL.
+	// RedirectURL is a cli.Flag used to pass a url to redirect after an OAuth
+	// flow finishes..
 	RedirectURL = cli.StringFlag{
 		Name:  "redirect-url",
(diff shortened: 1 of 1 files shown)


dopey (approved):
lgtm

PR #512, opened 2021-07-22
Add ACME External Account Binding commands

hslatman:
### Description

This PR adds (beta) ca commands for managing ACME External Account Binding key material. It's related to the ACME EAB support [PR](https://github.com/smallstep/certificates/pull/650).

Currently it supports adding new keys. Some more commands may be added. Being able to list and remove keys are probably important to implement. Also have to rewrite this into using Protobuf.

Let me know if the `eak` namespace is OK to use for this.


command/ca/eak/eak.go
```diff
@@ -0,0 +1,49 @@
...
+
+func toCLI(ctx *cli.Context, client *ca.AdminClient, eak *adminAPI.CreateExternalAccountKeyResponse) (*cliEAK, error) {
+	// TODO: more fields for other purposes, like including the createdat/boundat/account for listing?
+	return &cliEAK{id: eak.KeyID, name: eak.Name, key: base64.StdEncoding.EncodeToString(eak.Key)}, nil
+}
+
+// Command returns the jwk subcommand.
+func Command() cli.Command {
+	return cli.Command{
+		Name:      "eak",
```
maraino, line 25: eak is not a common name. Yes, it means external account key. EAB is a more common name (google `acme eak` vs `acme eab`), and it's probably best understood.

So I think we should probably add commands under `step ca eab` or `step ca acme-eab` or even `step ca acme eab`

The same can be applied to APIs in step-ca.

hslatman: I've added an `acme` command and `eab` subcommand below that.

command/ca/acme/eab/add.go
```diff
@@ -0,0 +1,87 @@
...
+		},
+		Description: `**step beta ca acme eab add** adds ACME External Account Binding Key.
+
+## POSITIONAL ARGUMENTS
+
+<provisioner_name>
+: Name of the provisioner to add an ACME EAB key to 
+
+<name_or_reference>
+:  Name or (external) reference for the key to be created
```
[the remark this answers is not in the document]
hslatman: I think you meant a single whitespace? That's what I saw in the other commands.

I've made this one consistent with the rest in 6bd3a20f23a6a326db143bccc6241e6c608abdfb.

command/ca/acme/eab/add.go
```diff
@@ -0,0 +1,87 @@
...
+		return err
+	}
+
+	eak, err := client.CreateExternalAccountKey(&adminAPI.CreateExternalAccountKeyRequest{
+		ProvisionerName: provisionerName,
+		Name:            credentialName,
+	})
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
hslatman: Discussed this one with @dopey, as it could go both ways. The conclusion was that it would make sense to allow users to search by the credentialName, so this should indeed become unique.

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,81 @@
...
+			flags.PasswordFile,
+			flags.CaURL,
+			flags.Root,
+		},
+		Description: `**step beta ca acme eab list** lists all ACME External Account Binding Keys.
+
+## POSITIONAL ARGUMENTS
+
+<provisioner_name>
+: Name of the provisioner to add an ACME EAB key to
```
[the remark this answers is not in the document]
hslatman: Seems to work OK

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,81 @@
...
+
+	args := ctx.Args()
+	provisionerName := args.Get(0)
+
+	client, err := cautils.NewAdminClient(ctx)
+	if err != nil {
+		return err
+	}
+
+	eaks, err := client.GetExternalAccountKeys(provisionerName)
```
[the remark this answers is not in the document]
hslatman: Indeed. The API returns an empty byte slice. The CLI output shows the key as masked, but that's faked; the actual contents are never returned as part of the list command; only shown once when adding a new key.

command/ca/provisionerbeta/update.go
```diff
@@ -231,7 +235,7 @@ step beta ca provisioner update x5c --x5c-root x5c_ca.crt
 
 Update an ACME provisioner:
 '''
-step beta ca provisioner update acme --force-cn
+step beta ca provisioner update acme --force-cn --require-eab --disable-eab
```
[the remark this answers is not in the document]
hslatman: You're right; it's a wrong example to combine both. Fixed in 6bd3a20f23a6a326db143bccc6241e6c608abdfb.

command/ca/provisionerbeta/update.go
```diff
@@ -573,6 +577,17 @@ func updateACMEDetails(ctx *cli.Context, p *linkedca.Provisioner) error {
 	if ctx.IsSet("force-cn") {
 		details.ForceCn = ctx.Bool("force-cn")
 	}
+	requireEABSet := ctx.IsSet("require-eab")
+	disableEABSet := ctx.IsSet("disable-eab")
+	if requireEABSet && disableEABSet {
+		return errors.New("cannot provide both --require-eab and --disable-eab")
```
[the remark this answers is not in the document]
hslatman: Fixed in 6bd3a20f23a6a326db143bccc6241e6c608abdfb.

command/ca/provisionerbeta/update.go
```diff
@@ -573,6 +577,17 @@ func updateACMEDetails(ctx *cli.Context, p *linkedca.Provisioner) error {
...
+	}
+	if requireEABSet {
+		details.RequireEab = ctx.Bool("require-eab")
+	}
+	if disableEABSet {
+		details.RequireEab = false
+	}
```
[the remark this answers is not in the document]
hslatman: I tried this and it works as expected.

command/ca/provisionerbeta/add.go
```diff
@@ -69,7 +69,7 @@ func addCommand() cli.Command {
 [**--admin-provisioner**=<string>] [**--admin-subject**=<string>]
 [**--password-file**=<file>] [**--ca-url**=<uri>] [**--root**=<file>]
 
-**step beta ca provisioner add** <name> **--type**=ACME [**--force-cn**]
+**step beta ca provisioner add** <name> **--type**=ACME [**--force-cn**] [**--require-eab**]
```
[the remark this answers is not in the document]
hslatman: I omitted it intentionally, because it's the default when creating a provisioner.

command/ca/provisionerbeta/provisioner.go
```diff
@@ -169,6 +169,14 @@ var (
...
 		Usage: `Always set the common name in provisioned certificates.`,
 	}
+	requireEABFlag = cli.BoolFlag{
+		Name:  "require-eab",
+		Usage: `Require (and enable) External Account Binding for Account creation.`,
+	}
+	disableEABFlag = cli.BoolFlag{
+		Name:  "disable-eab",
+		Usage: `Disable External Account Binding for Account creation.`,
+	}
```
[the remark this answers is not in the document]
hslatman: Transformation of the provisioner seems to work as expected. EAB can be enabled and disabled and the values are reflected properly in the provisioner listing afterwards. 

I have added additional error paths for when EAB is disabled on a provisioner in https://github.com/hslatman/certificates/commit/02cd3b6b3bf763662b24bef35279b07dd000b107.

command/ca/acme/eab/add.go
```diff
@@ -0,0 +1,87 @@
...
+
+	cliEAK, err := toCLI(ctx, client, eak)
+	if err != nil {
+		return err
+	}
+
+	// TODO: JSON output, so that executing this command can be more easily automated?
```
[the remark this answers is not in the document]
hslatman: Created https://github.com/smallstep/cli/issues/550. I hope this captures what you were thinking, too?

hslatman:
Will fix the `go.mod` before merging.

maraino (changes requested):
The pager loop does not work as I was expecting.

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,160 @@
...
+		eaksResponse, err := client.GetExternalAccountKeysPaginate(provisioner, reference, options...)
+		if err != nil {
+			return errors.Wrap(err, "error retrieving ACME EAB keys")
+		}
+		eaks = append(eaks, eaksResponse.EAKs...)
+		if eaksResponse.NextCursor == "" {
+			break
+		}
+		cursor = eaksResponse.NextCursor
+	}
```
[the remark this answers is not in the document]
maraino: We might need to manage SIGPIPE too to stop the loop. We might need to ask somebody to test that example managing SIGPIPE in windows.

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,160 @@
...
+	client, err := cautils.NewAdminClient(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error creating admin client")
+	}
+
+	// default to API paging per 100 entities
+	limit := 100
+	if ctx.IsSet("limit") {
+		limit = ctx.Int("limit")
+	}
```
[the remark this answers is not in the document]
maraino: Leave the limit to 0 by default, and let the server decide.

maraino (approved):
LGTM.

FYI: in the signal handler the defer (signals, and pager) won't be executed, but that should be ok.

command/ca/acme/acme.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/smallstep/cli/command/ca/acme/eab"
+	"github.com/urfave/cli"
+)
+
+// Command returns the acme subcommand.
+func Command() cli.Command {
+	return cli.Command{
+		Name:        "acme",
+		Usage:       "manage ACME EAB",
+		UsageText:   "**step beta ca acme** <subcommand> [arguments] [global-flags] [subcommand-flags]",
```
dopey, line 13: These aren't under `beta` are they?

dopey: oh, nvm. They are under `beta`. Is that a good idea? Why `beta` as opposed to normal? Are we still figuring out the API?

command/ca/acme/acme.go
```diff
@@ -0,0 +1,19 @@
...
+		Description: `**step beta ca acme** command group provides facilities for managing ACME EAB.`,
```
dopey, line 14: The subcommand `acme` is for EAB related configuration, right? whereas, `acme` is more generally for ACME configuration. Or am I misunderstanding?

command/ca/acme/eab/add.go
```diff
@@ -0,0 +1,94 @@
...
+			flags.CaURL,
+			flags.Root,
+			flags.Context,
+		},
+		Description: `**step beta ca acme eab add** adds ACME External Account Binding Key.
+
+## POSITIONAL ARGUMENTS
+
+<provisioner>
+: Name of the provisioner to add an ACME EAB key to
```
dopey, line 40: Name of the provisioner to which the ACME EAB key will be added.

command/ca/acme/eab/add.go
```diff
@@ -0,0 +1,94 @@
...
+			flags.Context,
+		},
+		Description: `**step beta ca acme eab add** adds ACME External Account Binding Key.
+
+## POSITIONAL ARGUMENTS
+
+<provisioner>
+: Name of the provisioner to add an ACME EAB key to 
+
+<reference>
```
dopey, line 42: `for the key that will be added.`

command/ca/acme/acme.go
hslatman: I just followed your example ;-)

More seriously: my thinking back then was that new commands would first be `beta` and somewhat later `beta` would be removed. Another thing that was on my mind back then was the fact that the DB backed provisioners were something new and this seemed to also (always) needed to be DB backed. 

I'd say the API, commands, etc. are most probably done, so in that sense they're not really `beta`.

command/ca/acme/acme.go
hslatman: Yes, that's the idea. There's only one subcommand under `acme` now though, which is why the usage now says `manage ACME EAB`. I'm not sure if there'll ever be more subcommands. Given that it's somewhat of a special kind of provisioner already, I don't think it's unlikely that there will be more subcommands in the future. WDYT?

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,178 @@
...
+	var out io.WriteCloser
+	var cmd *exec.Cmd
+
+	usePager := true
+	if ctx.IsSet("no-pager") {
+		usePager = !ctx.Bool("no-pager")
+	}
+
+	// the pipeSignalHandler goroutine ensures that the parent process is closed
+	// whenever one of its childs is killed.
```
dopey, line 96: children are killed

command/ca/acme/eab/list.go
```diff
@@ -0,0 +1,178 @@
...
+	if ctx.IsSet("limit") {
+		limit = ctx.Uint("limit")
+	}
+
+	cursor := ""
+	format := "%-36s%-28s%-16s%-30s%-30s%-36s%s\n"
+	firstIteration := true
+	startedPager := false
+
+	for {
```
dopey, line 125: I like the way you did this iteration. Gonna use this style elsewhere.

command/ca/acme/acme.go
dopey: We can leave it in beta for now if you like. Should be easy to migrate later since there's no conflic.

command/ca/acme/acme.go
dopey: Let's keep the usage for ACME more general on the chance that we come back and add more stuff in here. There's a good chance we'd forget to update it if we do it the other way around.

dopey (approved):
lgtm but I recommended a few small changes. mostly documentation / grammar.

command/ca/acme/eab/list.go
hslatman: 😅 Fixed in 686e78f8613be934418156b4c357eacb32e06f80.

command/ca/acme/eab/add.go
hslatman: Fixed in 686e78f8613be934418156b4c357eacb32e06f80.

command/ca/acme/acme.go
hslatman: Yes, let's leave it in beta for a bit. I'll be working on EAB management in the product too, so I'll revisit this soon.

command/ca/acme/eab/list.go
hslatman: First time implementing `$PAGER` and I like it too! There may be rough edges with how various terminals / $PAGERs work. I've only tested it with `less`. `--no-pager` probably unblocks those cases. Got that one from `git`. Also was inspired by @tashian's (a.o.) https://clig.dev/ 😃. 

I thought about making this a bit more generic, so that it could be in `cli-utils` sometime, but I'd need to do some more concrete instances first to see what works best. 

@maraino suggested performing the API calls in the same loop as printing (as opposed to loading all results first and then printing), which results in now only performing API calls when the user actively pages down. Simple, but effective! 🙂

PR #514, opened 2021-07-23
Step ca rekey Command

### LINKED ISSUE #316: Add `step ca rekey` command to integrate with new `Rekey` API in CA

tommy-56:
Closes #316 by adding `step ca rekey` as a command. 
Example Uses:

Rekey the certificate and write the certificate and newly generated private key to specified files:
`step ca rekey internal.crt internal.key --out-crt foo.crt --out-key foo.key
`
Rekey the certificate using a given private key:
`step ca rekey internal.crt internal.key --private-key foo.key`
--offline flag added on also works

### DIFF
--- a/command/ca/rekey.go
+++ b/command/ca/rekey.go
@@ -1,7 +1,6 @@
 package ca
 
 import (
-	"github.com/smallstep/cli/crypto/pemutil"
 	"go.step.sm/crypto/keyutil"
 	"io/ioutil"
 	"math/rand"
@@ -26,7 +25,7 @@ func rekeyCertificateCommand() cli.Command {
 		Action: command.ActionFunc(rekeyCertificateAction),
 		Usage:  "rekey a valid certificate",
 		UsageText: `**step ca rekey** <crt-file> <key-file>
-[**--out-crt**=<file>] [**--out-key**=<file>] [**--private-key**=<file>]
+[**--out-cert**=<file>] [**--out-key**=<file>] [**--private-key**=<file>]
 [**--ca-url**=<uri>] [**--root**=<file>] [**--password-file**=<file>]
 [**--out**=<file>] [**--expires-in**=<duration>] [**--force**]
 [**--expires-in**=<duration>] [**--pid**=<int>] [**--pid-file**=<file>]
@@ -203,24 +202,28 @@ if nothing is passed, default goes to generating a random pair.`,
 }
 
 func rekeyCertificateAction(ctx *cli.Context) error {
-	err := errs.NumberOfArguments(ctx, 2)
+	err := errs.MinMaxNumberOfArguments(ctx, 2, 3)
 	if err != nil {
 		return err
 	}
 
 	args := ctx.Args()
 	certFile := args.Get(0)
-	keyFile := ctx.String("private-key")
+	keyFile := args.Get(1)
+	privateKey := args.Get(2)
 	passFile := ctx.String("password-file")
-	outKey := ctx.String("out-key")
 	isDaemon := ctx.Bool("daemon")
 	execCmd := ctx.String("exec")
+	givenPrivate := ctx.String("private-key")
 
+	outCert := ctx.String("out-cert")
+	if len(outCert) == 0 {
+		outCert = certFile
+	}
 
-
-	outCrt := ctx.String("out-cert")
-	if len(outCrt) == 0 {
-		outCrt = certFile
+	outKey := ctx.String("out-key")
+	if len(outKey) == 0 {
+		outKey = keyFile
 	}
 
 	rootFile := ctx.String("root")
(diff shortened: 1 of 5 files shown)


command/ca/rekey.go
```diff
@@ -0,0 +1,340 @@
...
+[**--expires-in**=<duration>] [**--pid**=<int>] [**--pid-file**=<file>]
+[**--signal**=<int>] [**--exec**=<string>] [**--renew-period**=<duration>]`,
+		Description: `
+**step ca rekey** command rekeys the given certificate (with a request to the
+certificate authority) and writes the new certificate and private key 
+to disk - either overwriting <crt-file> <key-file> or using a new file when
+the **--out-cert**=<file> and **--out-key**=<file> flags are used.
+
+With the **--daemon** flag the command will periodically update the given
+certificate. By default, it will renew the certificate before 2/3 of the validity
```
dopey, line 40: renew -> rekey

command/ca/rekey.go
```diff
@@ -0,0 +1,340 @@
...
+		UsageText: `**step ca rekey** <crt-file> <key-file>
+[**--out-cert**=<file>] [**--out-key**=<file>] [**--private-key**=<file>]
+[**--ca-url**=<uri>] [**--root**=<file>] [**--password-file**=<file>]
+[**--expires-in**=<duration>] [**--force**] [**--exec**=<string>] [**--daemon**]
+[**--expires-in**=<duration>] [**--pid**=<int>] [**--pid-file**=<file>]
+[**--signal**=<int>] [**--exec**=<string>] [**--renew-period**=<duration>]`,
+		Description: `
+**step ca rekey** command rekeys the given certificate (with a request to the
+certificate authority) and writes the new certificate and private key 
+to disk - either overwriting <crt-file> <key-file> or using a new file when
```
dopey, line 36: either overwriting <crt-file> and <key-file> or writing to new files when the ...

command/ca/rekey.go
```diff
@@ -0,0 +1,340 @@
...
+			},
+			cli.BoolFlag{
+				Name: "daemon",
+				Usage: `Run the renew command as a daemon, renewing and overwriting the certificate
+periodically. By default the daemon will renew a certificate before 2/3 of the
+time to expiration has elapsed. The period can be configured using the
+**--renew-period** or **--expires-in** flags.`,
+			},
+			cli.StringFlag{
+				Name: "renew-period",
```
cookie-annihilator, line 178: @dopey should the flag change to rekey-period instead?

dopey: yep! I'm actually posting a few changed right now, so I can add that to my changes.

tommy-56:
Had to add some functions inside of renew.go to get the daemon wired up and working with rekey. After getting all that to work everything both rekey and renew work with all the different flags and options.

command/ca/rekey.go
```diff
@@ -335,23 +325,32 @@ func rekeyCertificateAction(ctx *cli.Context) error {
 		}
 	}
 
-	var flag = false
+	var writeKey = false
 	//want to write for these cases
 	if (givenPrivate != "" && ctx.IsSet("out-key")) || (ctx.IsSet("out-key")) || (givenPrivate == "" && !(ctx.IsSet("out-key"))) {
```
dopey, line 330: I *think* this can be summed up with 
```
if ctx .IsSet("out-key") || givenPrivate == "" {
...
}
```
Because the only way it would get to the second case is if ctx.IsSet("out-key") was not set, so we don't need to check that again. Can you check my logic here?

command/ca/renew.go
```diff
@@ -514,17 +514,17 @@ func (r *renewer) RenewAndPrepareNext(outFile string, expiresIn, renewPeriod tim
 
 	// Get next renew duration
 	next := nextRenewDuration(resp.ServerPEM.Certificate, expiresIn, renewPeriod)
-	Info.Printf("%s certificate rekeyed, next in %s", resp.ServerPEM.Certificate.Subject.CommonName, next.Round(time.Second))
+	Info.Printf("%s certificate renewed, next in %s", resp.ServerPEM.Certificate.Subject.CommonName, next.Round(time.Second))
```
dopey: shouldn't this be `rekeyed`?

command/ca/rekey.go
cookie-annihilator: I just tested that out with the different cases I was using and it works

command/ca/renew.go
tommy-56: Getting the daemon mode to work we made a new function for rekey since we had to pass in different parameters instead of bloating up `RenewAndPrepareNext()` we separated it into a new `RekeyAndPrepareNext()`  So there is still the renew functions there that print out has been renewed and the new ones for rekey that print out rekey.

dopey (approved):
lgtm

PR #519, opened 2021-08-10
error for invalid expires-in and fix renew daemon

dopey:


dopey:
@cookie-annihilator @tommy-56

command/ca/renew.go
```diff
@@ -504,7 +513,7 @@ func (r *renewer) Daemon(outFile string, next, expiresIn, renewPeriod time.Durat
 				return nil
 			}
 		case <-time.After(next):
-			if _, err := r.RenewAndPrepareNext(outFile, expiresIn, renewPeriod); err != nil {
+			if next, err = r.RenewAndPrepareNext(outFile, expiresIn, renewPeriod); err != nil {
```
maraino, line 516: We're not using the next here for anything.

maraino (approved):
Looks good, but we can ignore the next inside the for loop.

command/ca/renew.go
maraino: Ignore this, we are using in the channel.

PR #522, opened 2021-08-18
Add support to format CSRs to DER and back to PEM.

maraino:
### Description

This PR adds support of CSRs in `step certificate format`.

### DIFF
--- a/command/certificate/format.go
+++ b/command/certificate/format.go
@@ -21,16 +21,15 @@ func formatCommand() cli.Command {
 		Action:    command.ActionFunc(formatAction),
 		Usage:     `reformat certificate`,
 		UsageText: `**step certificate format** <crt_file> [**--out**=<file>]`,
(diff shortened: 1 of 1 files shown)


command/certificate/format.go
```diff
@@ -21,34 +21,34 @@ func formatCommand() cli.Command {
 		Action:    command.ActionFunc(formatAction),
 		Usage:     `reformat certificate`,
 		UsageText: `**step certificate format** <crt_file> [**--out**=<file>]`,
-		Description: `**step certificate format** prints the certificate in
-a different format.
+		Description: `**step certificate format** prints the certificate or certificate signing
+request in a different format.
```
dopey, line 25: nitpicking, but I would use `csr` here since that's what you use in the documentation below.

dopey (approved):
lgtm

command/certificate/format.go
maraino: Fixed in ad3e20a

PR #528, opened 2021-08-26
Github Actions cosign over all artifacts generated by goreleaser

dopey:


maraino (approved):
lgtm, we should add docs on how to use cosign to verify images.

PR #532, opened 2021-09-01
[action] sign multiarch docker images with cosign

dopey:


maraino (approved):
As long as cosign signs the images for all the platforms, it looks ok.

PR #534, opened 2021-09-03
Add checksum for debian

### LINKED ISSUE #526: Add checksums for Debian package
Add checksums for step-cli Debian packages.
Currently a checksum.txt file is provided for all artifacts except for the deb package. 

### Why this is needed
Although the security benefit is minimal to none (as both files are provided over https), it can be useful for scripts to verify that the correct file was downloaded.

dopey:
Fixes #526


maraino (changes requested):
I would try to append it to checksums.txt

maraino (approved):
Looks good to me, but can we remove `debian/` from the .gitignore?

PR #538, opened 2021-09-04
Fix `make bootstrap` failing to install `golangci-lint` using GOPATH

### LINKED ISSUE #535: When building step cli locally, `make bootstrap` tries to install golangci-lint in `/bin/` (and fails)
### Subject of the issue
When building step cli locally, `make bootstrap` tries to install golangci-lint in `/bin/` (and fails)

### Your environment
* OS - **Debian**
* Version - **11 (Bullseye)**

### Steps to reproduce
* Install Debian Bullseye (e.g. in a VM). Make sure `git` and `golang` are installed (using `apt` if necessary).
* Clone the step cli repository under go/src/github/smallstep/cli as indicated in `docs/local-development.md`
* Run `make bootstrap` to install `golangci-lint` (normally in `${HOME}/go/bin/`)

### Expected behaviour
`make bootstrap` should install `golangci-lint` in the `${HOME}/go//bin/` (user) directory (built on GOPATH) 

### Actual behaviour
`make bootstrap` attempts to install `golangci-lint` in the `/bin/` (system) directory (and fails, of course)

### Additional context
Analysis of make/common.mk shows a call to `$(go env GOPATH)`, which is interpreted as a *Make* construct rather than the *Bash* construct it is probably intended to be. If a `$` sign is prepended (i.e., `$$(go env GOPATH)`), then `make bootstrap` correctly installs `golangci-lint` under `${HOME}/go/bin/`.

aaribaud:
Fixes #535

### Description
To get the Go path, make/common.mk used `$(go env GOPATH)` which was thus evaluated by Make (wrongly, to an empty string). To get it evaluated (properly) by the shell, the expression must be `$$(go env GOPATH)`.

This PR adds the missing `$`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Deprecated
 ### Removed
 ### Fixed
+- Fix `make bootstrap` failing to get GOPATH and install `golangci-lint`.
 ### Security
 - Use cosign to sign and upload signatures for multi-arch Docker container.
 
--- a/make/common.mk
+++ b/make/common.mk
@@ -17,7 +17,7 @@ OUTPUT_ROOT=output/
 
 bootstra%:
 	# Using a released version of golangci-lint to take into account custom replacements in their go.mod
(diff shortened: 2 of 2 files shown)


dopey (approved):
lgtm

PR #543, opened 2021-09-08
needs-renewal: return 2 if file does not exist and ...

### LINKED ISSUE #541: Add explicit return code when file not found for `step certificate needs-renewal`
I think it would be useful to have an explicit return code for the case of a file not found for the `step certificate needs-renewal` command.

### Why this is needed
I was using this command in some ansible I am writing to renew certificates that have gone past their expiration date. I am reusing the same ansible role I used to install the cert in the first place. It would be useful for me to distinguish the state of the cert without having to parse the error message from the command. If the file not found condition resulted in a return code of, say, `2`, it would be easier for me to take specific action.

* Return code of `0`: Renew cert.
* Return code of `1`: Don't renew cert.
* Return code of `2` (certificate file not found): Provision and install cert from CA.
* Return code of `255`: Stop executing the playbook because something broke.

Thanks!

EDIT: I forgot to include the version of the cli I'm using:

```
Smallstep CLI/0.17.2 (linux/arm64)
Release Date: 2021-08-31T04:02:01Z
```

dopey:
- return 255 in some cases where we were not.

Fixes #541

### DIFF
--- a/command/certificate/needsRenewal.go
+++ b/command/certificate/needsRenewal.go
@@ -2,6 +2,7 @@ package certificate
 
 import (
 	"crypto/x509"
+	"os"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


maraino (approved):
👍

PR #544, opened 2021-09-08
Thread CA password decryption into offline CA bootstrap

### LINKED ISSUE #539: Use `--password-file` for CA/SSH key password in `ca token --offline`
 The `--password-file` flag is an alias for `--provisioner-password-file` in the `ca token` command. When `ca token` is invoked with the `--offline` flag, that means the intermediate CA key password must be provided interactively.

I request/suggest that the `--password-file` flag should not be an alias for `--provisioner-password-file`, but should allow the intermediate CA key password to be passed in so `ca token --offline` can be used non-interactively.

### Why this is needed
The `ca token --offline` command cannot be scripted as currently implemented.

The desired functionality would allow the following command to be performed without user input.

```bash
step ca token foo.example.org \
  --offline \
  --not-after 30m \
  --root $(step path)/certs/root_ca.crt \
  --password-file <(echo -n 'ca/ssh password') \
  --provisioner jwk \
  --provisioner-password-file <(echo -n 'jwk password') \
  --ssh --host
# => Currently, prints: "Cannot use two forms of the same flag: password-file provisioner-password-file"
```

Removing `--password-file <(echo -n 'ca/ssh password')` creates the token, once you enter the intermediate CA/SSH key password three times.

This is an instance of the general feature request in #502, but I haven't found any commands that it's not possible to script other than `ca token --offline`.

dopey:
Fixes #539.

### DIFF
--- a/utils/cautils/offline.go
+++ b/utils/cautils/offline.go
@@ -36,7 +36,7 @@ type OfflineCA struct {
 var offlineInstance *OfflineCA
 
 // NewOfflineCA initializes an offlineCA.
-func NewOfflineCA(configFile string) (*OfflineCA, error) {
+func NewOfflineCA(ctx *cli.Context, configFile string) (*OfflineCA, error) {
 	if offlineInstance != nil {
 		return offlineInstance, nil
 	}
@@ -55,10 +55,20 @@ func NewOfflineCA(configFile string) (*OfflineCA, error) {
 		return nil, errors.Errorf("error parsing %s: no provisioners found", configFile)
 	}
 
+	if ctx.IsSet("password-file") {
+		passFile := ctx.String("password-file")
+		pass, err := utils.ReadPasswordFromFile(passFile)
+		if err != nil {
+			return nil, errors.Wrapf(err, "error reading %s", passFile)
+		}
+		cfg.Password = string(pass)
+	}
+
 	auth, err := authority.New(&cfg)
 	if err != nil {
 		return nil, err
 	}
+	fmt.Printf("stop")
 
 	offlineInstance = &OfflineCA{
 		authority:  auth,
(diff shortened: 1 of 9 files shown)


utils/cautils/offline.go
```diff
@@ -55,10 +55,20 @@ func NewOfflineCA(configFile string) (*OfflineCA, error) {
...
+			return nil, errors.Wrapf(err, "error reading %s", passFile)
+		}
+		cfg.Password = string(pass)
+	}
+
 	auth, err := authority.New(&cfg)
 	if err != nil {
 		return nil, err
 	}
+	fmt.Printf("stop")
```
maraino, line 71: remove debug statement

maraino (changes requested):
And I still see x5c and sshpop using `password-file` instead of `provisioner-password-file`
They should prefer that one, and fallback to `password-file`.

See:
https://github.com/smallstep/cli/blob/cce2ea8ef22610e3f13cb7391d9cad8afd62d9b7/utils/cautils/token_generator.go#L170
https://github.com/smallstep/cli/blob/cce2ea8ef22610e3f13cb7391d9cad8afd62d9b7/utils/cautils/token_generator.go#L208
https://github.com/smallstep/cli/blob/cce2ea8ef22610e3f13cb7391d9cad8afd62d9b7/utils/cautils/token_generator.go#L239-L244

maraino (approved):
It looks good to me, but I think we should extract that switch logic to one private method so we can use it in the 3 cases:

It can be something like:
```go
func getProvisionerPasswordFile(ctx *cli.Context) (string, bool) {
	switch {
	case ctx.String("provisioner-password-file") != "":
		return ctx.String("provisioner-password-file"), true
	case ctx.String("password-file") != "":
		return ctx.String("password-file"), true
	default:
		return "", false
	}
}
```

Or if in all cases we are using the jose.WithPasswordFile, we can do `getProvisionerPasswordOptions(ctx *cli.Context) jose.Option`

labichn:
Works on my machine, thanks!

PR #545, opened 2021-09-09
Change dns to string slice in `ca init` but backwards compatible csv

### LINKED ISSUE #537: `ca init --dns=csv` validated as single DNS name, not a CSV of them
### Subject of the issue
The value passed to the `--dns` flag of `step ca init` is validated incorrectly.

The issue seems to be [here][init.go#L372]. The `ui.DNS()` [validator][validators.go#L46] tests for a single DNS name, not a CSV of them.

### Your environment
* OS - Arch Linux
* Version - Smallstep CLI/0.16.1 (linux/amd64) Release Date: 2021-07-12

### Steps to reproduce
```bash
$ step ca init --dns=[::1] --address=:443 --name=ca --provisioner=jwk --password-file=<(echo foo)
# Works fine
$ step ca init --dns=localhost,127.0.0.1 ...
# Works because 'localhost,127.0.0.1' is a valid host per https://pkg.go.dev/net#SplitHostPort
$ step ca init --dns=localhost,127.0.0.1,[::1] ...
localhost,127.0.0.1,[::1] is not a valid DNS name
```

### Expected behaviour
The third command should continue without error and lead to the expected `ca.json:dnsNames`:
 ```bash
$ step ca init --dns=localhost,127.0.0.1,[::1] --address=:443 --name=ca --provisioner=jwk --password-file=<(echo foo)
$ jq .dnsNames $STEPPATH/config/ca.json
[
    "localhost",
    "127.0.0.1",
    "[::1]"
]
```

### Actual behaviour
The validation error `localhost,127.0.0.1,[::1] is not a valid DNS name` is printed to stderr.

[init.go#L372]: https://github.com/smallstep/cli/blob/master/command/ca/init.go#L372
[validators.go#L46]: https://github.com/smallstep/cli/blob/master/ui/validators.go#L46

### PR DESCRIPTION
**Pull request by dopey**
Fixes #537

### DIFF
--- a/command/ca/init.go
+++ b/command/ca/init.go
@@ -82,9 +82,10 @@ func initCommand() cli.Command {
 				Name:  "name",
 				Usage: "The <name> of the new PKI.",
 			},
-			cli.StringFlag{
-				Name:  "dns",
-				Usage: "The comma separated DNS <names> or IP addresses of the new CA.",
+			cli.StringSliceFlag{
+				Name: "dns",
+				Usage: `The DNS <name> or IP address of the new CA.
+Use the '--dns' flag multiple times to configure multiple DNS names.`,
 			},
 			cli.StringFlag{
 				Name:  "address",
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: command/ca/init.go [hunk 1]
```diff
@@ -366,23 +367,28 @@ func initAction(ctx *cli.Context) (err error) {
...
 			if len(name) == 0 {
 				continue
 			}
+			if err := dnsValidator(name); err != nil {
+				return err
+			}
 			dnsNames = append(dnsNames, strings.TrimSpace(name))
 		}
 
+		fmt.Printf("dnsNames = %+v\n", dnsNames)
```
maraino, line 391: Remove debug statement.

maraino (changes requested):
Looks good to me, but can you remove the debug statement

PR #551, opened 2021-09-17
Remove checks for OIDC provided certificates.

### LINKED ISSUE #340: Loose requirements for step ca certificate

`step ca certificate` and related commands are sometimes too strict, and create certificate requests with unwanted data. This is specifically noticeable creating host certificates using the OIDC provisioner with an admin user, where the user email is added to the CSR, and ends up in the certificate. See smallstep/certificates#339

### Notes to take into account
1. We originally wanted the subject to be specified in `step ca certificate <subject>` and for it to match the subject in the token to close a potential security hole: if the OTT is used to bootstrap and get a certificate an adversary could send a fake OTT and trick you into getting a cert from the wrong CA with the wrong name. This is a pretty obscure hole though and probably not critical.
2. More importantly, the check(s) here do help with debugging. Without checking this you’ll probably get a pretty opaque authn/authz failure error from the CA. I can think of two workarounds: we could serve templates via API and the CLI could try to figure out what the error might be (probably too hard / lots of issues to contend with). Alternatively (and easier) we could keep some of the logic we currently have but only run it if we get an authn/authz error back from the CA. At that point we could use that logic to include a “suggestion” in the CLI error like “Authn/authz failed. This may be because the requested subject foo did not match email bar@example.com. Check CA logs for details”.

maraino:
### Description

This PR removes the checks in `step` when certificates are signed using OIDC provisioners.

With OIDC tokens the CSR will be built based on the `--san` flags, subject argument, token.sub, token.email and token.iss based on the following rules:

* If at least one `--san` is provided: CN=subject SANs=splitSANs(sans)
* If no sans and subject == token.email: CN=token.sub SANs=token.email, token.iss#token.sub
* if no sans and subject != token.email: CN=subject SANs=splitSANs(subject)

Fixes #340

Notes: 
* If a non-admin uses a subject that does not match his email the CSR sent to the CA will be build using the rules above, but he will get a certificate with CN=token.sub, SANs=email, token.iss#token.sub
* Extra checks that we want to add in the CA are not part of the scope of this PR, and we should add a new issue to validate this if we want to show an error.

### DIFF
--- a/utils/cautils/certificate_flow.go
+++ b/utils/cautils/certificate_flow.go
@@ -287,7 +287,7 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
 		}
 	case token.OIDC:
 		// If no sans are given using the --san flag, and the subject argument
-		// matches the email then CN=token.sub SANs=email.
+		// matches the email then CN=token.sub SANs=email, token.iss#token.sub
 		//
 		// If no sans are given and the subject argument does not match the
 		// email then CN=subject SANs=splitSANs(subject)
@@ -297,6 +297,10 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
 			if jwt.Payload.Email != "" && strings.EqualFold(subject, jwt.Payload.Email) {
 				subject = jwt.Payload.Subject
 				emails = append(emails, jwt.Payload.Email)
+				if iss, err := url.Parse(jwt.Payload.Issuer); err == nil && iss.Scheme != "" {
+					iss.Fragment = jwt.Payload.Subject
+					uris = append(uris, iss)
+				}
 			} else {
 				dnsNames, ips, emails, uris = splitSANs([]string{subject})
 			}
--- a/utils/cautils/certificate_flow.go
+++ b/utils/cautils/certificate_flow.go
@@ -293,6 +293,10 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
 		// email then CN=subject SANs=splitSANs(subject)
 		//
 		// If sans are provided CN=subject SANs=splitSANs(sans)
+		//
+		// Note that with the way token types are identified, an OIDC token with
+		// `sans` claim will never reach this code. We will leave the condition
+		// as it is in case we want it to support it later.
 		if len(sans) == 0 && len(jwt.Payload.SANs) == 0 {
 			if jwt.Payload.Email != "" && strings.EqualFold(subject, jwt.Payload.Email) {
 				subject = jwt.Payload.Subject
--- a/utils/cautils/certificate_flow.go
+++ b/utils/cautils/certificate_flow.go
@@ -286,10 +286,21 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
 			dnsNames, ips, emails, uris = splitSANs(defaultSANs)
 		}
 	case token.OIDC:
-		if jwt.Payload.Email != "" {
-			emails = append(emails, jwt.Payload.Email)
+		// If no sans are given using the --san flag, and the subject argument
+		// matches the email then CN=token.sub SANs=email.
+		//
+		// If no sans are given and the subject argument does not match the
+		// email then CN=subject SANs=splitSANs(subject)
+		//
+		// If sans are provided CN=subject SANs=splitSANs(sans)
+		if len(sans) == 0 {
+			if jwt.Payload.Email != "" && strings.EqualFold(subject, jwt.Payload.Email) {
+				subject = jwt.Payload.Subject
+				emails = append(emails, jwt.Payload.Email)
+			} else {
+				dnsNames, ips, emails, uris = splitSANs([]string{subject})
+			}
 		}
-		subject = jwt.Payload.Subject
 	case token.K8sSA:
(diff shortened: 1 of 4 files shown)


mmalone:
These rules look good to me. I do think the non-admin + subject mismatch scenario should probably result in an error. But, yea, that error should be detected server-side and it makes sense to deal with that as a separate issue / PR.

mmalone (approved):

utils/cautils/certificate_flow.go
```diff
@@ -286,10 +286,25 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
...
-		if jwt.Payload.Email != "" {
-			emails = append(emails, jwt.Payload.Email)
+		// If no sans are given using the --san flag, and the subject argument
+		// matches the email then CN=token.sub SANs=email, token.iss#token.sub
+		//
+		// If no sans are given and the subject argument does not match the
+		// email then CN=subject SANs=splitSANs(subject)
+		//
+		// If sans are provided CN=subject SANs=splitSANs(sans)
+		if len(sans) == 0 {
```
dopey, line 296: Do we need to be checking jwt.Payload.SANs here as well?

utils/cautils/certificate_flow.go
```diff
@@ -286,10 +286,25 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
...
+			if jwt.Payload.Email != "" && strings.EqualFold(subject, jwt.Payload.Email) {
+				subject = jwt.Payload.Subject
+				emails = append(emails, jwt.Payload.Email)
+				if iss, err := url.Parse(jwt.Payload.Issuer); err == nil && iss.Scheme != "" {
+					iss.Fragment = jwt.Payload.Subject
+					uris = append(uris, iss)
+				}
+			} else {
+				dnsNames, ips, emails, uris = splitSANs([]string{subject})
```
dopey, line 305: what about the SANs in jwt.Payload.SANs?

dopey: https://github.com/smallstep/cli/pull/551/files#diff-d45e314b2cc37cd26b0c92007c7a6dabc13a7006602b51bf9f5e9b4729b8c1ebR252

utils/cautils/certificate_flow.go
```diff
@@ -286,10 +286,25 @@ func (f *CertificateFlow) CreateSignRequest(ctx *cli.Context, tok, subject strin
...
+		// If no sans are given using the --san flag, and the subject argument
+		// matches the email then CN=token.sub SANs=email, token.iss#token.sub
+		//
+		// If no sans are given and the subject argument does not match the
+		// email then CN=subject SANs=splitSANs(subject)
+		//
+		// If sans are provided CN=subject SANs=splitSANs(sans)
+		if len(sans) == 0 {
+			if jwt.Payload.Email != "" && strings.EqualFold(subject, jwt.Payload.Email) {
+				subject = jwt.Payload.Subject
```
dopey, line 298: In this conditional we're appending SANs to emails and URIs, but in the case below we're creating new splitSANs, ignoring existing SANs. I'm confused by the inconsistency.

dopey (changes requested):
I have some questions. Left Comments.

utils/cautils/certificate_flow.go
maraino: I had it like that before, but I wanted to make it explicit when the `--san` flag was passed, and having a custom claim "sans" in an OIDC token would be quite weird.

utils/cautils/certificate_flow.go
maraino: I've looked at this again, and if you add the "sans" claim the logic thinks that is a JWT provisioner, so it won't go here. So right now we don't support the custom claim "sans" in an OIDC token.

utils/cautils/certificate_flow.go
maraino: What are the existing SANs? Due to how tokens work, "sans" is not currently supported on OIDC tokens.

So the only sans are from `--san` or from the `subject` positional argument. 

Without `--san` arguments:
* If the email == subject, then we create the CSR using the same way the CA creates the template, so admins will have the same SANs and CN a non-admin will have.
* If the email != subject, then we create the CSR with the appropriate SAN (dns, email, IP or uri).

But if at least one `--san` is passed, then we create the CSR according to the splitSANs, an admin will get the appropriate certificate, and non-admins will get a certificate with its email in it.

With this logic, admins will always send the CSR that they want, and non-admins too, but they will get a different certificate, the one with the email because that's what the CA does on the OIDC provisioner. The CSR is not checked at all. And this is something that we will have to change in the CA, but as I said in the description this is a different story, and we'll have to change the CA to return an error for non-admins if the CSR has a SAN that does not match the email or uri, or some similar logic.

utils/cautils/certificate_flow.go
dopey: At the very least let's document that invariant. But I'd rather explicitly set dnsNames, ips, emails, uris to empty lists. If that invariant ever changes, and we're using append to existing lists then the behavior will silently change.

maraino: Fixed in 245ddf1

utils/cautils/certificate_flow.go
maraino: Fixed in 245ddf1

dopey (approved):
lgtm

PR #553, opened 2021-09-27
Remove vestigial ssh certificate command password hack

dopey:


maraino (approved):
Looks good to me, but we might want to do `if ctx.String("password-file") != ""` in offline.go

PR #556, opened 2021-09-30
Flexible config and OAuth redirect_uri

maraino:
### Description

This PR allows to use `--set` and `--team` together in the `step ssh config`. There's no need to make them incompatible and can be useful for some templates.

This PR also allows overriding a fixed listen-address coming from the CA with the value on an environment variable. This combined with a new `--listen-url` flag allows to do the OAuth flow in restricted environments like containers if the IDP doesn't support the out of band flow.

These changes allow the following flow with an OIDC provisioner:
```
$ docker run -it -p 10000:10000 smallstep/step-cli:latest /bin/bash
bash-5.1$ step ca bootstrap --ca-url ...  --fingerprint ...
bash-5.1$ export STEP_LISTEN=0.0.0.0:10000
bash-5.1$ export STEP_LISTEN_URL=http://127.0.0.1:10000
bash-5.1$ step ca certificate mariano@smallstep.com mariano.crt mariano.key
```

If we just use `STEP_LISTEN=:10000`, the host machine won't be able to able to send the request back to the server as docker won't allow port redirections going to 127.0.0.1. And we cannot use `STEP_LISTEN=0.0.0.0:10000` because some IDPs like Google won't allow this. For example, Google shows this:

```
Authorization Error
Error 400: invalid_request
Invalid parameter value for redirect_uri: Raw IP addresses not allowed: http://0.0.0.0:10000
```

Combining both we can have the redirect_uri set to http://127.0.0.1:10000 and we can use the port mapping properly.

I've used the name `--listen-url` instead of `--redirect-uri` to avoid confusion with the current `--redirect-url` parameter and because this new flag will be generally used together with `--listen`, although this is not a requirement.

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -116,6 +116,11 @@ Redirect to a fixed port instead of random one:
 $ step oauth --listen :10000
 '''
 
+Redirect to a fixed url but listen in all the interfaces:
+'''
+$ step oauth --listen 0.0.0.0:10000 --listen-url http://127.0.0.1:10000
+'''
+
 Get just the access token:
 '''
 $ step oauth --bare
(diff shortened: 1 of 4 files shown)


command/oauth/cmd.go
```diff
@@ -116,6 +116,11 @@ Redirect to a fixed port instead of random one:
 $ step oauth --listen :10000
 '''
 
+Redirect to a fixed url but listen in all the interfaces:
```
dopey, line 119: listen on

dopey (approved):
lgtm

command/oauth/cmd.go
maraino: Fixed on 8ed7c9f

PR #565, opened 2021-10-16
feat: utils/cautils/acmeutils.go: Treat all public CAs the same

### LINKED ISSUE #564: Use public CA infrastructure even for non-letsencrypt CA
### What would you like to be removed
Remove the special case with http transporter for letsencrypt. Check if --root is specified instead, if not, use public CA.

### Why this is needed
Assume that the clients already have their CA installed in the trust, there should be no reason to specify the --root param since it's already available.
Even so, you still require --root right now if it's not letsencrypt, so why not use it in https://github.com/smallstep/cli/blob/c0f884caf607cbcec7dce1f48f0d03bd3a371971/utils/cautils/acmeutils.go#L381

isodude:
If --root is given as a parameter, not-after/not-before is allowed.
If not, then treat it as any public CA.

This does not change any functionality, but rather allow skipping --root
if the private CA is already in the systems trust.

Signed-off-by: Josef Johansson <josef@oderland.se>

resolves #564 

💔Thank you!

### DIFF
--- a/utils/cautils/acmeutils.go
+++ b/utils/cautils/acmeutils.go
@@ -3,11 +3,13 @@ package cautils
 import (
 	"context"
 	"crypto/rand"
+	"crypto/tls"
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -361,6 +363,43 @@ func newACMEFlow(ctx *cli.Context, ops ...acmeFlowOp) (*acmeFlow, error) {
 	return af, nil
 }
 
+func (af *acmeFlow) getRootCAs(mergeRootCAs bool) (ca.ClientOption, error) {
+	root := ""
+	if af.ctx.IsSet("root") {
+		root = af.ctx.String("root")
+		// If there's an error reading the local root ca, ignore the error and use the system store
+	} else if _, err := os.Stat(pki.GetRootCAPath()); err == nil {
+		root = pki.GetRootCAPath()
+	}
+
+	// 1. Merge local RootCA with system store
+	if mergeRootCAs && len(root) > 0 {
+		rootCAs, err := x509.SystemCertPool()
+		if err != nil || rootCAs == nil {
+			rootCAs = x509.NewCertPool()
+		}
+
+		cert, err := ioutil.ReadFile(root)
+		if err != nil {
+			return ca.WithRootFile(root), errors.Wrap(err, "failed to read local root ca")
+		}
+
+		if ok := rootCAs.AppendCertsFromPEM(cert); !ok {
+			return ca.WithRootFile(root), errors.New("failed to append local root ca to system cert pool")
+		}
+
+		return ca.WithTransport(&http.Transport{TLSClientConfig: &tls.Config{RootCAs: rootCAs}}), nil
+	}
+
+	// Use local Root CA only
+	if len(root) > 0 {
+		return ca.WithRootFile(root), nil
+	}
+
+	// Use system store only
+	return ca.WithTransport(http.DefaultTransport), nil
+}
+
 func (af *acmeFlow) GetCertificate() ([]*x509.Certificate, error) {
 	dnsNames, ips, err := validateSANsForACME(af.sans)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


isodude:
CLA signed.

utils/cautils/acmeutils.go
```diff
@@ -361,32 +360,45 @@ func newACMEFlow(ctx *cli.Context, ops ...acmeFlowOp) (*acmeFlow, error) {
...
 
 func (af *acmeFlow) GetCertificate() ([]*x509.Certificate, error) {
+	var (
+		orderPayload []byte
+		clientOps    []ca.ClientOption
+		idents       []acme.Identifier
+		nor          interface{}
+	)
+
+	// Public CA does not support NotBefore and NotAfter attributes in orders.
```
dopey, line 370: I'm not sure we want to error here. Let's Encrypt is not the only public ACME CA and don't think there's anything in the spec that would prevent `not-after` and `not-before` being sent. Instead, we probably want to make these values pointers, so that they can be set to nil if the values are empty - rather than the zero time value.

utils/cautils/acmeutils.go
```diff
@@ -402,37 +414,23 @@ func (af *acmeFlow) GetCertificate() ([]*x509.Certificate, error) {
...
-		orderPayload, err = json.Marshal(struct {
-			Identifiers []acme.Identifier
-		}{Identifiers: idents})
-		if err != nil {
-			return nil, errors.Wrap(err, "error marshaling new letsencrypt order request")
-		}
+		nor = struct{ Identifiers []acme.Identifier }{Identifiers: idents}
+	}
+
+	if root := af.ctx.String("root"); root == "" {
```
dopey, line 420: I'm confused, because this actually will change the functionality. Before, if you didn't provide a root flag we would still try to load the root from the default location (it didn't need to be in your truststore). But this implementation forces you to either use `--root` or have it in your truststore if you're using a private CA. I'm wary of changing this behavior. Are there options that avoid changing existing behavior? If we are going to change behavior here, I'll need to discuss with the team again.

utils/cautils/acmeutils.go
isodude: Make sense.

utils/cautils/acmeutils.go
isodude: How about
If --root is detected, use that.
If --root is "" but the default root exists, use that.
If not use truststore.

dopey: The problem that I see here is that if `--root` is not provided, and I do have a root in the default step location, then I won't use the default transport to call a public CA - so this will fail. Which means the behavior will have changed. Let me bring this back up with the team during our next triage.

isodude: Couldn't you include rather than only use the default step location, thus combining truststore and default step location root cert?

dopey: That's definitely an option. Need to consider the security implications as well as whether this may cause unexpected behavior for anyone. We've got a triage meeting on wednesdays - will circle back after that.

dopey:
Hey @isodude, apologies for the long turn around here. We had a chance to discuss the logic and here's how we think it should work:

Case 1: ctx.IsSet("acme") or ctx.String("acme") != "" (one of those two, not both).

If the acme flag is used then if `ctx.String("root") > 0` then add that value to the system truststores and use the combination as the truststore for the TLS connection.

Case 2: `acme` flag is not set.

If `ctx.String("root") > 0` then use this as the only root cert in the TLS connection. Otherwise use the default truststore. I think in this case, the code may fail elsewhere if `ctx.String("root")` doesn't exist (not sure).

Does that logic make sense? Let me know if you have any questions.

On an unrelated note, we'd love to send you some swag for the contribution and for bearing with us through the comments and delays. Shoot me an email (max@smallstep.com) and we'll set it up.

Cheers!

isodude:
I'll update the PR with you suggestion and see how it looks :)

>  On an unrelated note, we'd love to send you some swag for the contribution and for bearing with us through the comments and delays. Shoot me an email (max@smallstep.com) and we'll set it up.

You can send some to the address on the bottom of oderland.se if you want to :-)

dopey:
Hey Josef, according to our CLA bot above it still doesn't think you've signed the CLA. Seems like you may have signed it as Josef Johansson rather than using your github account. However, it seems like you have commits from both accounts on this PR so you may need to sign it with both accounts 🤷 . 

Also, let me know if you have any lingering questions about the feedback. Cheers 🍻

isodude:
@dopey I have signed the CLA, but I think the bot still thinks that I am josjoh@oderland.se.

I got a lot on my plate right now, sorry for not updating the issue. Next week will probably be better!

dopey:
@isodude no worries at all. We're not in any rush. 

Reach out if you're looking for any assistance or a review. Good luck knocking stuff out in the mean time, and happy holidays!

isodude:
> Hey @isodude, apologies for the long turn around here. We had a chance to discuss the logic and here's how we think it should work:
> 
> Case 1: ctx.IsSet("acme") or ctx.String("acme") != "" (one of those two, not both).
> 
> If the acme flag is used then if `ctx.String("root") > 0` then add that value to the system truststores and use the combination as the truststore for the TLS connection.
> 
> Case 2: `acme` flag is not set.
> 
> If `ctx.String("root") > 0` then use this as the only root cert in the TLS connection. Otherwise use the default truststore. I think in this case, the code may fail elsewhere if `ctx.String("root")` doesn't exist (not sure).
> 
> Does that logic make sense? Let me know if you have any questions.
> 

Coming to think about it, you don't mention the default root CA (.step/certs/root_ca.pem), should we ignore it completely?

isodude:
Sorry for the delay, the PR is updated now, the commit message should reflect the changes properly.

PR #586, opened 2021-11-23
Update autocomplete scripts from urfave/cli

elyscape:
The [bash] and [zsh] autocomplete scripts from upstream [urfave/cli] now handle autocompletion of files and also correctly handle command flags.

[bash]: https://github.com/urfave/cli/blob/master/autocomplete/bash_autocomplete
[zsh]: https://github.com/urfave/cli/blob/master/autocomplete/zsh_autocomplete
[urfave/cli]: https://github.com/urfave/cli

### DIFF
--- a/autocomplete/bash_autocomplete
+++ b/autocomplete/bash_autocomplete
@@ -1,21 +1,16 @@
 #! /bin/bash
 
-: ${PROG:=$(basename ${BASH_SOURCE})}
-
-_cli_bash_autocomplete() {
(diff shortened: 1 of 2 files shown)


dopey:
I'll be honest I don't know how any of this works. But I tested it, and it's definitely better than before.

dopey (approved):
lgtm. and working.

dopey:
Thanks @elyscape!

PR #587, opened 2021-12-01
Build .deb and .rpm packages via GoReleaser

tashian:
GoReleaser includes [support for building Linux packages](https://goreleaser.com/customization/nfpm/).

With this config, I was able to build a `.deb` & `.rpm` that is similar to what we have now. Some subtle differences:
- The archive doesn't contain a `changelog.gz` file. As far as I can tell though, this file never had anything in it.
- The version number uses `~` instead of `.` for rc version numbers, eg. `0.18.0~rc3`

TODOs:
- [x] Limit supported architectures. Right now it's building rpms and debs for 386, mips, etc.
- [x] Clean out `debian` directory, keeping only `copyright`
- [x] Remove `changelog` make rule?
- [x] Test asset links and apt .deb install (Ubuntu) WORKS
- [x] Get yum .rpm install working properly (Rocky Linux)
  - [x] There is no `step` command / alternative in the .rpm package; we may need another nFPM profile for .rpm that installs the binary as `step` or does some fancy pre/post install footwork
 - [ ] We should consider using conventional .rpm and .deb filenames instead of the ones we're using now. If we were to set the nfpm file name template to `"{{ .ConventionalFileName }}"`, would we be able to generate the right file names in the release notes Markdown? (Because that template field isn't available there)
- [x] Multi-line package description (first line is a summary, like a git commit message)


.goreleaser.yml
```diff
@@ -124,7 +147,7 @@ release:
     #### Linux
 
     - 📦 [step_linux_{{ .Version }}_amd64.tar.gz](https://dl.step.sm/cli/{{ .Tag }}/step_linux_{{ .Version }}_amd64.tar.gz)
-    - 📦 [step-cli_{{ .Env.DEB_VERSION }}_amd64.deb](https://dl.step.sm/cli/{{ .Tag }}/step-cli_{{ .Env.DEB_VERSION }}_amd64.deb)
```
dopey, line 127: Why are we changing this? Originally I used DEB_VERSION instead of VERSION because our release candidate versions have a dash which is generally not supported in debian package versions. So DEB_VERSION would be the version but with the dashes replaced by another character. 

If we no longer need to worry about dashes then we should remove the presence of `DEB_VERSION` entirely. If we do need to worry about them, then we should switch back to using `DEB_VERSION`. We should test with a release candidate.

.github/workflows/release.yml
```diff
@@ -112,14 +112,15 @@ jobs:
         id: aptInstall
         run: sudo apt-get -y install build-essential debhelper fakeroot
       -
```
dopey, line 114: Why are we commenting instead of removing this whole section?

dopey:
I left a couple comments / questions.

Another question: you mentioned the changelog.gz doesn't exist. Doesn't the archive contain the changelog file updated by `make changelog`?

dopey (changes requested):
left a few comments.

tashian:
@dopey wrote:
> Another question: you mentioned the changelog.gz doesn't exist. Doesn't the archive contain the changelog file updated by `make changelog`?

It does, but the resulting changelog file isn't a changelog. It never changes, so I'm not sure why we need it?

dopey:
I think `make debian` fails without a changelog file. One of the steps we are currently using requires it. Seems like maybe goreleaser doesn't?

tashian:
ok @dopey I think this is ready for an rc test!

.goreleaser.yml
tashian: So, ultimately the way this worked out with GoReleaser is, the metadata will have `~` instead of `-`, but the filename will have the `-`. And I think that's OK, but if we need something different we can specify a custom filename in the the `nfpms` Linux packager config above

.github/workflows/release.yml
```diff
@@ -119,14 +112,13 @@ jobs:
           echo "::set-output name=RELEASE_DATE::${RELEASE_DATE}"
       -
         name: Run GoReleaser
-        uses: goreleaser/goreleaser-action@5a54d7e660bda43b405e8463261b3d25631ffe86 # v2.7.0
```
tashian, line 122: @dopey were you pinning to v2.7.0 of the action for a reason?
just noticed that v2.8.0 was released in October, wanted to be sure we were up to date

tashian:
Ok @dopey I think this is ready to merge.

Here's where we ended up:
- The `.deb` is basically the same as before, in terms of contents. Tested it on Ubuntu.
- The `.rpm` may need some adjustments, but I was able to install it properly on Rocky Linux using `yum`. It installs a symlink for the `step` command instead of Debian alternatives.

.github/workflows/release.yml
```diff
@@ -92,17 +89,12 @@ jobs:
       -
         name: APT Install
         id: aptInstall
-        run: sudo apt-get -y install build-essential debhelper fakeroot
+        run: sudo apt-get -y install build-essential fakeroot
```
dopey, line 92: Do you know if we still need build-essential and fakeroot? Those were only necessary for the `make debian` step.

dopey (approved):
lgtm

PR #594, opened 2021-12-08
Describe revocation via JWK vs mTLS

tashian:


command/ca/revoke.go
```diff
@@ -65,16 +65,29 @@ certificate lifetimes.
...
+client certificate.
+
+When you supply a serial number, you're prompted to choose a JWK provisioner,
+and a provisioner token is transparently generated. Any JWK provisioner
+can revoke any certificate.
+
+When you supply a certificate and private key (with --crt and --key),
+mTLS is used to authorize the revocation.
+
+Certificates generated using the OIDC provisioner cannot be revoked
```
dopey, line 78: This statement isn't true. It's just that I can' revoke any serial number using an OIDC provisioner (or possibly any provisioner that is not JWK).

command/ca/revoke.go
```diff
@@ -65,16 +65,29 @@ certificate lifetimes.
 **step ca revoke** currently only supports passive revocation. Active revocation
 is on our roadmap.
 
+A revocation request can be authorized using a JWK provisioner token, or using a
+client certificate.
+
+When you supply a serial number, you're prompted to choose a JWK provisioner,
```
dopey, line 71: Currently, you're prompted to choose any one of your provisioners. But you should choose a JWK provisioner. Until we fix the code.

dopey (changes requested):
left one comment where the documentation wasn't quite right. Otherwise lgtm

Issue #190, opened 2019-12-02
Add support for RHEL/CentOS


ghost:
### What would you like to be added

Please add support for RHEL/CentOS. Ideally, it would be great to update the documentation to include instructions for CentOS 7/8 and to also generate RPM packages.

Instead of creating and maintaining your own RPM repository, you may submit your packages to the EPEL repository.

### Why this is needed

RHEL is a major server Linux distro by RedHat and CentOS is a community-driven distro based on RHEL.

By generating RHEL/CentOS rpm packages, _step_ becomes available to millions of systems.

Thank you.

dopey:
Hey @robo2bobo, thanks for opening an issue / feature request! We've actually gotten this request a couple of times from our [gitter community](https://gitter.im/smallstep/community) as well, so there's definitely a need for it. 

The last few months for us have been heads-down working towards a release. Once that release is ready to go we'll definitely prioritize making a RHEL build. 

Having said that, if you (or anyone with experience) have any guidance on CI/CD for submitting packages to the EPEL repository, we'd definitely be interested to see or hear about examples. The team doesn't have much experience with RHEL/CentOS which is the only reason we hadn't prioritized a package before. 

We had a request for an ArchLinux package (which we also have little to no experience with) and luckily one of the maintainers was able to help us cobble a package together.

robo2bobo:
Unfortunately I have no experience with rpm packages, but you could ask in the #epel irc channel in freenode for more details, I'm sure they have helped many people with how to create rpm packages.

robo2bobo:
no interest, closing

dopey:
Hey @robo2bobo we're still planning on doing this - probably in the near term. Sorry for the radio silence, but we've been heads down working on product stuff towards an upcoming release that focuses on ssh usability. Once that's released, making the `cli` and `ca` easily accessible on different operating systems is on our short list.

Would you mind if I reopen, just in case other folks are searching for the same thing?

mkkeffeler:
Can confirm this is very much needed!

maraino:
Reopening this issue, as this is something we plan to do.

sourishkrout:
Related to https://github.com/smallstep/cli/issues/163.

rabin-io:
If anyone need rpm packages for step-cli or step-ca I currently package them manually from the release binaries, 
you can find them here: 

- https://build.opensuse.org/package/show/home:rabin-io/step-cli
[...]

PR #600, opened 2021-12-27
Use alternatives system for RPMs

elyscape:
This unifies the package scriptlets. Additionally, it changes the preremove scriptlet to a postremove script, as that's the more typical place to remove alternatives entries.

[As promised](https://github.com/smallstep/cli/issues/190#issuecomment-999421026).


tashian:
Thanks Eli!
And I'm glad we have one set of scripts for all platforms.
That really simplifies things.
Just one question on this.
From the [RPM docs](http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html) it appears that the defined %ghost files are removed  by the package manager when the rpm is uninstalled.
For the `step` symlink, won't `update-alternatives --remove` also remove the symlink?
If so, maybe we don't need the ghost file for `step` itself?

elyscape:
> From the [RPM docs](http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html) it appears that the defined %ghost files are removed by the package manager when the rpm is uninstalled.

Ghost files are treated as owned by a package but without specifying contents. As a result, multiple packages can own a file provided that all of them (with up to one exception) specify it as a ghost file. A package's ghost files are removed unless some other installed package also provides those files (as ghosts or otherwise). As an aside, that property also applies to directories.

> For the `step` symlink, won't `update-alternatives --remove` also remove the symlink?
> If so, maybe we don't need the ghost file for `step` itself?

The [Fedora packaging guidelines][fedora] recommend ghost files for alternatives:

> Putting the alternativized files in the file list ensures that they are owned by respective packages, which means that commands like:
> - rpm -qf /usr/bin/foo
> - dnf install /usr/bin/foo
> - repoquery --whatprovides /usr/bin/foo
> 
> all work properly.

While obviously we aren't packaging for the official Fedora repos, this seems like it's probably worthwhile.

[fedora]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Alternatives/#_how_to_use_alternatives

PR #607, opened 2022-01-11
Update OIDC add error for duplicate ID

dopey:
- oidc uses only the clientID as the unique identifier.

### DIFF
--- a/command/ca/provisioner/add.go
+++ b/command/ca/provisioner/add.go
@@ -503,7 +503,7 @@ func addOIDCProvisioner(ctx *cli.Context, name string, provMap map[string]bool) 
 	if _, ok := provMap[p.GetID()]; !ok {
 		provMap[p.GetID()] = true
 	} else {
-		return nil, errors.Errorf("duplicated provisioner: CA config already contains a provisioner with name=%s and client-id=%s", p.GetName(), p.GetID())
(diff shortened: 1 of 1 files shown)


command/ca/provisioner/add.go
```diff
@@ -503,7 +503,7 @@ func addOIDCProvisioner(ctx *cli.Context, name string, provMap map[string]bool)
 	if _, ok := provMap[p.GetID()]; !ok {
 		provMap[p.GetID()] = true
 	} else {
-		return nil, errors.Errorf("duplicated provisioner: CA config already contains a provisioner with name=%s and client-id=%s", p.GetName(), p.GetID())
+		return nil, errors.Errorf("duplicated provisioner: CA config already contains a provisioner with client-id=%s", p.GetID())
```
maraino, line 506: I'm not sure if this is the right fix, on a db-backed step-ca GetID will be a random uuid:
```go
// GetID returns the provisioner unique identifier, the OIDC provisioner the
// uses the clientID for this.
func (o *OIDC) GetID() string {
	if o.ID != "" {
		return o.ID
	}
	return o.GetIDForToken()
}

// GetIDForToken returns an identifier that will be used to load the provisioner
// from a token.
func (o *OIDC) GetIDForToken() string {
	return o.ClientID
}
```

maraino (approved):
lgtm

PR #610, opened 2022-01-19
Normalize IPv6 addresses as IPs instead of hostnames

### LINKED ISSUE #605: ca init: does not accept IPv6 addresses
### Subject of the issue

`step ca init` does not accept IPv6 addresses as valid.

### Your environment

* Version - 0.18.0

### Steps to reproduce

```
step ca init
✔ Deployment Type: Standalone
What would you like to name your new PKI?
✔ (e.g. Smallstep): Test
What DNS names or IP addresses would you like to add to your new CA?
✔ (e.g. ca.smallstep.com[,1.1.1.1,etc.]): 1.1.1.1
What IP and port will your new CA bind to?
✔ (e.g. :443 or 127.0.0.1:443): :443
What would you like to name the CA's first provisioner?
✔ (e.g. you@smallstep.com): test@example.com
Choose a password for your CA keys and first provisioner.
✔ [leave empty and we'll generate one]: 

Generating root certificate... done!
Generating intermediate certificate... done!
...
```

IPv6 is not recognized as valid (taking the [example one from Wikipedia](https://en.wikipedia.org/wiki/IPv6_address#Representation)):

```
$ step ca init
✔ Deployment Type: Standalone
What would you like to name your new PKI?
✔ (e.g. Smallstep): Test
What DNS names or IP addresses would you like to add to your new CA?
✔ (e.g. ca.smallstep.com[,1.1.1.1,etc.]): 2001:0db8:85a3:0000:0000:8a2e:0370:7334
2001:0db8:85a3:0000:0000:8a2e:0370:7334 is not a valid DNS name
```

hslatman:
This PR normalizes IPv6 addresses internally as an IP instead of the hostname notation using square brackets. When a request is made to an IPv6 host, the required square brackets are added again.

Because the CA URL is also used in the token audience, the CA also needs to ensure that the IPv6 address is properly encoded in the audiences. The PR for this is here: https://github.com/smallstep/certificates/pull/804.

Fixes #605.


dopey (approved):
lgtm

hslatman:
@dopey: made some additional improvements to make the conversions work more reliably. Also applied it to the offline mode.

flags/flags.go
```diff
@@ -477,9 +477,34 @@ func parseCaURL(ctx *cli.Context, caURL string) (string, error) {
...
+		// these cases, then the input dns is not changed.
+		lastIndex := strings.LastIndex(host, ":")
+		hostPart, portPart := host[:lastIndex], host[lastIndex+1:]
+		if ip := net.ParseIP(hostPart); ip != nil {
+			hostname = "[" + hostPart + "]:" + portPart
+		} else if ip := net.ParseIP(host); ip != nil {
+			hostname = "[" + host + "]"
+		}
+		u.Host = hostname
+	} else if ip := net.ParseIP(hostname); ip != nil && ip.To4() == nil {
```
dopey, line 501: Can you give an example of a IPv6 address that would not fall into the first case, but would fall into the second case. Just trying to understand how we'd fall into the second case.

dopey: Is the second case only for IPv6 addresses that were bracketed? I guess, would you mind adding a comment on the second case to just make it abundantly clear what we're checking and why. 

Currently I'm thinking we do that because if the original address had brackets, url.Parse() will remove them. So we want to add them back.

dopey (approved):
lgtm but asking for an additional comment in one spot (for slow brained folks like myself).

flags/flags.go
hslatman: I must've overlooked this case when checking the test coverage, because it doesn't seem to be covered by the test cases and I can't currently think of a case that would end up here. They would already fail at parsing the URL. I think I must've changed the first case to cover all weird IPv6 addresses; at least the ones that could already be parsed as an URL.

hslatman:
> lgtm but asking for an additional comment in one spot (for slow brained folks like myself).

Apparently I'm part of that same group 😅

PR #612, opened 2022-01-19
Change authority and profile to default to context name when set

### LINKED ISSUE #606: Store context-enabled CA configuration in a context-specific directory
Currently the logic for determining the directories for storing all configuration and files when providing `context=some-example-context` uses the first DNS provided (i.e. `ca.localhost`) as the root directory name for a CA configuration. This results in prompting to overwrite files in directories when creating a new CA using the same (first) DNS entry, even though the `context` is different, which may be surprising. This also makes it harder to run multiple CAs on the same hostname(s), but using different ports. Finally, it requires the user to find the right CA directory belonging to a context if there's a need to inspect the configuration manually.

I think it makes sense to store the configuration for a context-enabled CA in a directory structure that relates to the context. So, for a CA with context `local-ca` and DNS `ca.localhost`, instead of:

```
...
authorities/...
authorities/ca.localhost
authorities/ca.localhost/...
profiles/...
profiles/ca.localhost
profiles/ca.localhost/...
...
```

It would result in the directory structure as follows:

```
...
authorities/...
authorities/local-ca
authorities/local-ca/...
profiles/...
profiles/local-ca
profiles/local-ca/...
...
```

Everything that belongs to a certain CA context will then by default be in a directory structure that starts with the name of the context, unless overruled by providing the `--authority` or `--profile` flag.

WDYT, @dopey?

hslatman:
Before this change the authority and profile would default to
the first DNS name, potentially resulting in multiple CAs using
the same directories and configuration or configurations being
overwritten.

After this change the name of the context will be used as the
authority and profile name by default. They can still be overridden
using the `--authority` and `--profile` flags. When no context
name is provided, we still default to the first DNS name provided.

Fixes #606


hslatman:
@dopey: let me know if it makes sense to do something similar for bootstrapping. I think it's good to be consistent, but I saw that the bootstrap sets some more meaningful defaults.

dopey (approved):
lgtm

dopey:
@hslatman I agree that we should do the same thing (where possible) for `ca bootstrap` and for `ssh config`. And like you said we need to be careful not to break existing functionality. Luckily there's no tests for any of this :)

hslatman:
Lucky us! 😝 

I'll add the logic for `bootstrap`. And good that you mentioned `ssh config`; I didn't have that one on my mind. Will do the integration test manually.

dopey:
https://github.com/smallstep/cli/blob/b2b8f52035246d4387c201128200ced1fafd1735/utils/cautils/bootstrap.go#L251

Currently results in: ```
"ssh.beta": {
        "profile": "beta",
        "authority": "ssh.beta.ca.smallstep.com"
    },
```

If we're ok with that being `ssh.beta` for everything then we may as well remove the unused attributes in the BootstrapOption.

I think we used the team name for the profile because we thought a team might just have a single profile. But you can always configure that yourself by using the flag.

PR #613, opened 2022-01-19
Clarify listen port message in `step ca init --helm`

maraino:
### Description
Add clarification for the listen port in `step ca init --helm`

Fixes smallstep/helm-charts#82


tashian (approved):
lgtm

PR #615, opened 2022-01-21
Add --auth-param flag to oauth command

dopey:


command/oauth/cmd.go
```diff
@@ -67,21 +67,25 @@ func init() {
 		Usage: "authorization and single sign-on using OAuth & OIDC",
 		UsageText: `**step oauth**
 [**--provider**=<provider>] [**--client-id**=<client-id> **--client-secret**=<client-secret>]
-[**--scope**=<scope> ...] [**--bare** [**--oidc**]] [**--header** [**--oidc**]] [**--prompt**=<prompt>]
+[**--scope**=<scope> ...] [**--bare** [**--oidc**]] [**--header** [**--oidc**]]
+[**--prompt**=<prompt>] [**--auth-param**=<key-value>]
```
maraino, line 71: The help should display `[**--auth-param**=<key=value>]` to make it more clear.

command/oauth/cmd.go
```diff
@@ -336,7 +352,20 @@ func oauthCmd(c *cli.Context) error {
 		prompt = c.String("prompt")
 	}
 
-	o, err := newOauth(opts.Provider, clientID, clientSecret, authzEp, tokenEp, scope, prompt, opts)
+	authParams := map[string]string{}
+	for _, keyval := range c.StringSlice("auth-param") {
+		parts := strings.Split(keyval, "=")
+		if len(parts) != 2 {
+			return errs.InvalidFlagValue(c, "auth-param", keyval, "")
+		}
```
maraino, line 360: It is not common, but we can be a little more permissive and use SplitN, so weird parameters like `--auth-param foo=bar=zar` will work, key will be `foo` and the value `bar=zar`

command/oauth/cmd.go
```diff
@@ -336,7 +352,20 @@ func oauthCmd(c *cli.Context) error {
...
+		k, v := parts[0], parts[1]
+		if k == "" || v == "" {
+			return errs.InvalidFlagValue(c, "auth-param", keyval, "")
+		}
```
maraino, line 364: I would say it's ok to have `v == ""`, key is not.

command/oauth/cmd.go
```diff
@@ -438,11 +467,12 @@ type oauth struct {
 	CallbackPath        string
 	terminalRedirect    string
 	browser             string
+	authParams          map[string]string
```
maraino, line 470: You should use [url.Values](https://pkg.go.dev/net/url#Values) or at what is the same `map[string][]string`

maraino (changes requested):
Being a little fussy

command/oauth/cmd.go
```diff
@@ -902,7 +940,7 @@ func (o *oauth) Auth() (string, error) {
 	if o.loginHint != "" {
 		q.Add("login_hint", o.loginHint)
 	}
-	u.RawQuery = q.Encode()
+	u.RawQuery = fmt.Sprintf("%s&%s", q.Encode(), o.authParams.Encode())
```
maraino, line 943: I would say is cleaner to merge those two url.Values rather than appending them like that.

maraino (changes requested):
Looks ok, but I would merge the url.Values instead of doing what you do.

maraino (approved):
LGTM

PR #616, opened 2022-01-21
Add SCEP (beta) provisioner support

hslatman:
This PR adds support for adding and updating SCEP provisioners.

It also includes some small typo fixes.

### DIFF
--- a/command/ca/provisionerbeta/add.go
+++ b/command/ca/provisionerbeta/add.go
@@ -73,7 +73,13 @@ func addCommand() cli.Command {
 **step beta ca provisioner add** <name> **--type**=ACME [**--force-cn**]
 [**--admin-cert**=<file>] [**--admin-key**=<file>] [**--admin-provisioner**=<name>]
 [**--admin-subject**=<subject>] [**--password-file**=<file>] [**--ca-url**=<uri>]
-[**--root**=<file>] [**--context**=<name>]`,
+[**--root**=<file>] [**--context**=<name>]
+
+**step beta ca provisioner add** <name> **--type**=SCEP [**--force-cn**] [**--challenge**=<challenge>] 
+[**--capabilities**=<capabilities>] [**--include-root**] [**--minimum-public-key-length**=<length>] 
+[**--encryption-algorithm-identifier**=<id>] [**--admin-cert**=<file>] [**--admin-key**=<file>] 
+[**--admin-provisioner**=<string>] [**--admin-subject**=<string>] [**--password-file**=<file>] 
+[**--ca-url**=<uri>] [**--root**=<file>] [**--context**=<name>]`,
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "type",
@@ -107,7 +113,10 @@ func addCommand() cli.Command {
     : Uses Kubernetes Service Account tokens.
 
     **SSHPOP**
-    : Uses an SSH Certificate / private key pair to sign provisioning tokens.`},
+    : Uses an SSH Certificate / private key pair to sign provisioning tokens.
+	
+    **SCEP**
+	: Uses the SCEP protocol to create certificates.`},
 			x509TemplateFlag,
 			x509TemplateDataFlag,
 			sshTemplateFlag,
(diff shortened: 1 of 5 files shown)


command/ca/provisionerbeta/add.go
```diff
@@ -73,7 +73,13 @@ func addCommand() cli.Command {
 **step beta ca provisioner add** <name> **--type**=ACME [**--force-cn**]
 [**--admin-cert**=<file>] [**--admin-key**=<file>] [**--admin-provisioner**=<name>]
 [**--admin-subject**=<subject>] [**--password-file**=<file>] [**--ca-url**=<uri>]
-[**--root**=<file>] [**--context**=<name>]`,
+[**--root**=<file>] [**--context**=<name>]
+
+**step beta ca provisioner add** <name> **--type**=SCEP [**--force-cn**] [**--challenge**=<challenge>] 
+[**--capabilities**=<capabilities>] [**--include-root**] [**--minimum-public-key-length**=<length>]
```
dopey, line 79: It think we can use `min-public-key-length`.

command/ca/provisionerbeta/provisioner.go
```diff
@@ -170,6 +170,35 @@ var (
 		Usage: `Always set the common name in provisioned certificates.`,
 	}
 
+	// SCEP provisioner flags
```
dopey, line 173: When defining flags that have a value try to us the type in the description inside `<>`, like so: https://github.com/smallstep/cli/blob/6b23538520fc8c720b067bec2fb7fe630f8825f4/flags/flags.go#L251-L257. That's become sort of a standard.

dopey: According to the usage at the top, the type for the challenge flag is `<challenge>`.

command/ca/provisionerbeta/update.go
```diff
@@ -794,3 +815,32 @@ func updateGCPDetails(ctx *cli.Context, p *linkedca.Provisioner) error {
...
+	}
+	details := data.SCEP
+
+	if ctx.IsSet("force-cn") {
+		details.ForceCn = ctx.Bool("force-cn")
+	}
+	if ctx.IsSet("challenge") {
+		details.Challenge = ctx.String("challenge")
+	}
+	if ctx.IsSet("capabilities") {
```
dopey, line 832: Are you intending to replace the entire list with the new one? For a few of the other provisioners I had to define a `remove-<flag>` to have a list of the values to remove and then do the logic in the update method before post-ing. Might no be necessary for this, just letting you know.

command/ca/provisionerbeta/update.go
```diff
@@ -794,3 +815,32 @@ func updateGCPDetails(ctx *cli.Context, p *linkedca.Provisioner) error {
...
+		details.Capabilities = ctx.StringSlice("capabilities")
+	}
+	if ctx.IsSet("minimum-public-key-length") {
```
dopey, line 835: don't forget to change this to `min` if you change the flag name

dopey (changes requested):
see comments

command/ca/provisionerbeta/add.go
hslatman: Indeed; more consistent with other minimums. Fixed in 8606fdcf0560e56688309948f5c81d23f8588022.

command/ca/provisionerbeta/provisioner.go
hslatman: 👍 

Will try to keep this in mind for future additions.

command/ca/provisionerbeta/update.go
hslatman: I saw your previous work and thought about adding those, but decided against it (for now). There'll likely be more work in the future related to SCEP and that might also touch the capabilities. Making this part a bit more user friendly can be done then, I think.

command/ca/provisionerbeta/update.go
hslatman: Fixed in 8606fdcf0560e56688309948f5c81d23f8588022.

dopey (approved):
lgtm

PR #621, opened 2022-02-01
Improvements step ssh login and logout

maraino:
### Description

This PR makes the command `step ssh login` and `step ssh logout` more intuitive. Allowing the use of them without any positional argument.

### DIFF
--- a/command/ssh/login.go
+++ b/command/ssh/login.go
@@ -200,6 +200,18 @@ func loginAction(ctx *cli.Context) error {
 		return err
 	}
 
+	// NOTE: For OIDC tokens the subject should always be the email. The
+	// provisioner is responsible for loading and setting the principals with
+	// the application of an Identity function.
+	if email, ok := tokenHasEmail(token); ok {
+		subject = email
+	} else if subject == "" {
+		// For non-oidc tokens we will use the token subject.
+		if sub, ok := tokenSubject(token); ok {
+			subject = sub
+		}
+	}
+
 	// Generate identity certificate (x509) if necessary
 	var identityCSR api.CertificateRequest
 	var identityKey crypto.PrivateKey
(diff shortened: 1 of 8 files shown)


command/ssh/login.go
```diff
@@ -24,9 +24,9 @@ func loginCommand() cli.Command {
 		Name:   "login",
 		Action: command.ActionFunc(loginAction),
 		Usage:  "adds a SSH certificate into the authentication agent",
-		UsageText: `**step ssh login** <identity>
+		UsageText: `**step ssh login** [<identity>]
 [**--token**=<token>] [**--provisioner**=<name>] [**--provisioner-password-file**=<file>]
```
dopey, line 28: Can we add an identity flag to the usage text

command/ssh/login.go
```diff
@@ -187,6 +200,18 @@ func loginAction(ctx *cli.Context) error {
 		return err
 	}
 
+	// NOTE: For OIDC tokens the subject should always be the email. The
+	// provisioner is responsible for loading and setting the principals with
+	// the application of an Identity function.
+	if email, ok := tokenHasEmail(token); ok {
```
dopey, line 206: it's strange to have `tokenHasEmail` and `tokenSubject` when they return the same thing (`(val, bool)`). Let's use the same naming convention. I like `tokenSubject` and `tokenEmail`.

crypto/sshutil/agent.go
```diff
@@ -50,6 +50,16 @@ func WithSignatureKey(keys []ssh.PublicKey) AgentOption {
 	}
 }
 
+// WithCertsOnly filter by only certificates.
```
dopey, line 53: filter only those keys accompanied by a certificate.

utils/cautils/token_flow.go
```diff
@@ -100,6 +100,21 @@ func NewTokenFlow(ctx *cli.Context, tokType int, subject string, sans []string,
 		return "", err
 	}
 
+	if subject == "" {
+		// For OIDC provisioners the CA automatically generate the principals
```
dopey, line 104: automatically generates

command/ssh/login.go
maraino: The identity flag is only there so it can be defined in the defaults.json, you're not supposed to use it.

command/ssh/login.go
maraino: Fixed with 2c2a8b5

crypto/sshutil/agent.go
maraino: Fixed with e1f1e3a

utils/cautils/token_flow.go
maraino: Fixed with e99c2f0

PR #626, opened 2022-02-03
Pass missing template data in X509 sign offline mode

### LINKED ISSUE #624: .Insecure.User variables are not set in templates when using --set "var=val"
### Subject of the issue
I've set up an offline CA and want to specify the certificate's DN with values that are supplied on the command line.
Now the values I give to the "step ca" command with the --set option are not available in the template / generated certificate, they show as "&lt;no value&gt;". I'm referring to the documentation at https://smallstep.com/docs/step-ca/templates .

### Your environment
* OS - Linux (Ubuntu 20.04)
* Step Version - 0.18.0

### Steps to reproduce
Set up a simple CA as per the docs. Then configure a JWK provisioner that has a template configured:
Mine is like this:

```

   ...
   "authority": {
   		"provisioners": [
			{
				"type": "JWK",
				"name": "web",
				"key": {
                                    ...
                                 },
				"options": {
					"x509": {
						"templateFile": "/home/stepca/web.tpl"
					}
				}
			}
         ]
    ....
```
The file "/home/stepca/web.tpl" looks like this:

```
{
    "subject": {
        "country": "CH",
        "organization": "{{ .Insecure.User.org }}",
        "commonName": "{{ .Subject.CommonName }}",
    },
    "keyUsage": ["keyEncipherment", "digitalSignature"],
    "extKeyUsage": ["serverAuth", "clientAuth"],
    "sans": {{ toJson .SANs }}
}
```

I then execute this command line:

```
step ca certificate theCN new.crt new.key --provisioner web --set "org=MYORG" --offline
```
The resulting certificate does not have the organization field set:
 ```

openssl x509 -text -in new.crt | grep Subject
        Subject: C = CH, O = "<no value>", CN = theCN
        Subject Public Key Info:
            X509v3 Subject Key Identifier: 
            X509v3 Subject Alternative Name: 
```

### Expected behaviour
I would expect the generated certificate to have a Subject like this:

```
Subject: C = CH, O = "MYORG", CN = theCN
```

### Actual behaviour
The value that I pass in via the command line's "--set" parameter is not available in the template, apparently.

maraino:
### Description

When templates are used in offline mode, the `TemplateData` was missing in X.509 signing requests, SSH signing was working fine.

Fixes #624

### DIFF
--- a/utils/cautils/offline.go
+++ b/utils/cautils/offline.go
@@ -190,8 +190,9 @@ func (c *OfflineCA) Sign(req *api.SignRequest) (*api.SignResponse, error) {
 		return nil, err
 	}
 	signOpts := provisioner.SignOptions{
-		NotBefore: req.NotBefore,
-		NotAfter:  req.NotAfter,
+		NotBefore:    req.NotBefore,
+		NotAfter:     req.NotAfter,
+		TemplateData: req.TemplateData,
 	}
 	certChain, err := c.authority.Sign(req.CsrPEM.CertificateRequest, signOpts, opts...)
 	if err != nil {


dopey (approved):
lgtm

PR #628, opened 2022-02-08
Add completion command

### LINKED ISSUE #622: step completion sub-command
It's common to see the sub-command `completion` followed by the shell's name to generate the completion script to standard out. The user can redirect the output to the appropriate file for their shell.

The output of `step completion bash` would be the same as:
[https://github.com/smallstep/cli/blob/master/autocomplete/bash_autocomplete](https://github.com/smallstep/cli/blob/master/autocomplete/bash_autocomplete)
But without the shebang.

### Why this is needed
I think it's better UX for the user. When I try to install a completion for any tool, the first place I would look for is the `completion` sub-command.

And for people who like to automate the installations of their favorite CLIs, it's very convenient in scripts.

ooraini:
Fixes #622

### Description
Add shell completion command. #622

I removed the `-o nospace` from `autocomplete/bash_autocomplete`. I think it's far better.

### DIFF
--- a/command/completion/completion.go
+++ b/command/completion/completion.go
@@ -1,0 +1,110 @@
+package completion
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/cli/flags"
+	"github.com/urfave/cli"
+	"go.step.sm/cli-utils/command"
+	"go.step.sm/cli-utils/errs"
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:      "completion",
+		Usage:     "print the shell completion script",
+		UsageText: "**step completion** <shell>",
+		Description: `**step completion** command prints the shell completion script.
+
+## POSITIONAL ARGUMENTS
+
+<shell>
+: The shell program. Either bash or zsh.
+
+## EXAMPLES
+
+Add bash completion for the current user.
+'''
+$ step completion bash >> ~/.bash_completion
+'''
+`,
+		Flags: []cli.Flag{
+			flags.HiddenNoContext,
+		},
+		Action: Completion,
+		BashComplete: func(c *cli.Context) {
+			if c.NArg() > 0 {
+				return
+			}
+			fmt.Println("bash")
+			fmt.Println("zsh")
+		},
+	}
+
+	command.Register(cmd)
+}
+
+var bash = `# bash completion for step
+_step_cli_bash_autocomplete() {
+	local cur opts base
+	COMPREPLY=()
+	cur="${COMP_WORDS[COMP_CWORD]}"
+	if [[ "$cur" == "-"* ]]; then
+		opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} ${cur} --generate-bash-completion )
+	else
+		opts=$( ${COMP_WORDS[@]:0:$COMP_CWORD} --generate-bash-completion )
+	fi
+	COMPREPLY=( $(compgen -W "${opts}" -- ${cur}) )
+	return 0
+}
+
+complete -o bashdefault -o default -F _step_cli_bash_autocomplete step
+
+`
+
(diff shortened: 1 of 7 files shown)


command/completion/completion.go
```diff
@@ -0,0 +1,107 @@
...
+Add bash completion for the current user.
+'''
+$ step completion bash >> ~/.bash_completion
+'''
+`,
+		Flags: []cli.Flag{
+			flags.HiddenNoContext,
+		},
+		Action: Completion,
+		BashComplete: func(c *cli.Context) {
```
dopey, line 35: I can't figure out what this method is doing. If I run `step completion` I just get the error which is already in the action about now having enough command line args.

command/completion/completion.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+			fmt.Println("bash")
+			fmt.Println("zsh")
+		},
+	}
+
+	command.Register(cmd)
+}
+
+var bash = `# bash completion for step
```
dopey, line 47: Can we read these values from the `./autocomplete/xxx` files? Otherwise we'll have to keep these updated in two places.

command/completion/completion.go
```diff
@@ -0,0 +1,107 @@
...
+
+`
+
+func Completion(ctx *cli.Context) error {
+	if err := errs.NumberOfArguments(ctx, 1); err != nil {
+		return err
+	}
+
+	shell := ctx.Args().First()
+	if shell == "bash" {
```
dopey, line 98: Our linter is complaining about this if-else condition. Any if-else with more than 2 branches and it wants us to use `switch` instead.

dopey (changes requested):
see comments

command/completion/completion.go
ooraini: This function gets invoked when you write `step completion` and press tab, it will basically check if you haven't already passed an argument, or else output the `bash` and `zsh`. I will test it out again.

command/completion/completion.go
ooraini: There two options here:

1. Phase out the files in autocomplete and change all installation scripts to use the command instead.
2. Symlink the files inside the package and use something like embed to access them.

I'm leaning towards option 1. What do you think?

command/completion/completion.go
ooraini: Sure thing.

command/completion/completion.go
dopey: Oh that's pretty cool! Just tested it out and it works for me.

command/completion/completion.go
dopey: I do like option 1 but I wonder if that will break other people's automation. I really don't know how people are installing `step` but if they expect those files to be present (as we do in our debian files and our Makefile) then they'll run into issues.

ooraini: We could keep them for a while and transition everything towards the command, and at some point later remove them. The completion scripts are very unlikely to change, so we wouldn't worry about them diverging.

dopey: I like that idea. 

The places that I think we need to update:
* .goreleaser.yml
* debian/step-cli.bash-completion
* make/bundle.sh

dopey: We can add a README in the autocomplete dir that says those files are deprecated and the source of truth in the future will be `step completion <shell>`.

ooraini: Sounds good. Should I do them in this PR?

dopey: Please, if you wouldn't mind. I know they're not "directly related" to your initial issue.

ooraini: It's ready now. @dopey

dopey:
Trying out the debian build in a tagged release candidate. Assuming that works, will merge.

PR #630, opened 2022-02-09
Add IPs as SANs when performing ACME flow with existing CSR

hslatman:
When performing an ACME flow with an existing CSR, not all SANs would be in the ACME order. This PR fixes that.

It's related to https://github.com/smallstep/certificates/discussions/819.


dopey (approved):
lgtm

PR #629, opened 2022-02-09
`step ca renew` in offline mode should not require `--ca-url` and `--root` flags.

henschkowski:
Fixes #627.

I'm new to Go - this is my attempt to fix the issue.

💔Thank you!

PR #631, opened 2022-02-09
Use parseCaURLIfExists for renew and rekey in case of offline

### LINKED ISSUE #627: Offline "step ca renew" insists on --ca-url and --root parameters
### Subject of the issue
It is irritating that the command
```step ca renew renewed.crt renewed.key --offline --ca-config ./config/ca.json ``` requires the --ca-url and --root parameters.
The documentation at https://smallstep.com/docs/step-cli/reference/ca/renew has a sample invocation that does not work at all for me:
```step ca renew --offline internal.crt internal.key```

### Your environment
* OS - Linux (Ubuntu 20.04)
* Version - master (84e2b87739a581ea0e83efa0861fc671a9d51654)

### Steps to reproduce

- Create an offline CA with CA config at config/ca.json (not the default path)
- Issue a certificate with ```step ca certificate c c.crt c.key --offline --ca-config config/ca.json```
- Renew that certificate with ```step ca renew c.crt c.key --offline --ca-config config/ca.json```

### Expected behaviour
The certificate can be renewed with the command line above and/or a correct command line for this command is supplied at https://smallstep.com/docs/step-cli/reference/ca/renew

### Actual behaviour
The command aborts with 
```'step ca renew' requires the '--ca-url' flag```
When a dummy URL is provided, the command 
```step ca renew c.crt c.key --offline --ca-config config/ca.json --ca-url https://dummy.url``` 

complains:

```open /home/user/.step/certs/root_ca.crt failed: no such file or directory```
The "root" property in config/ca.json is not honored.

When giving the --root and --ca-url options, the certificate is renewed as expected.

dopey:
Fixes #627 

@henschkowski we've already got a method that will parse the CAURL iff it is not empty. My thought is we just use that, but getting a second opinion from @maraino.

### DIFF
--- a/command/ca/rekey.go
+++ b/command/ca/rekey.go
@@ -234,7 +234,7 @@ func rekeyCertificateAction(ctx *cli.Context) error {
 		rootFile = pki.GetRootCAPath()
 	}
 
-	caURL, err := flags.ParseCaURL(ctx)
+	caURL, err := flags.ParseCaURLIfExists(ctx)
 	if err != nil {
 		return err
 	}
--- a/command/ca/renew.go
+++ b/command/ca/renew.go
@@ -214,7 +214,7 @@ func renewCertificateAction(ctx *cli.Context) error {
 		rootFile = pki.GetRootCAPath()
 	}
 
-	caURL, err := flags.ParseCaURL(ctx)
+	caURL, err := flags.ParseCaURLIfExists(ctx)
 	if err != nil {
 		return err
 	}


maraino (changes requested):
As it is the error if `--ca-url` is not passed and you're using an online CA is not clear.

dopey:
If that's the case, I think we can use @henschkowski 's original PR https://github.com/smallstep/cli/pull/629.

PR #632, opened 2022-02-09
Improvements to CRL inspect

tashian:
Allow verification of a CRL with empty Authority Key ID.
Add a Reason if CRL is invalid.


maraino (approved):
LGTM, it's a little bit weird to use Reason in text and invalid_reason in json.

PR #634, opened 2022-02-10
Improve ACME Admin API error message 

hslatman:


command/ca/acme/eab/eab.go
```diff
@@ -67,3 +72,16 @@ $ step beta ca acme eab remove key_id
 `,
 	}
 }
+
+// notImplemented checks if an error indicates that the operation is not implemented
+// in the CA and adds additional information to the error if that's the case. Other
+// types of errors pass through without changes.
+func notImplemented(err error) error {
+	var adminErr *ca.AdminClientError
+	if errors.As(err, &adminErr) && adminErr.Type == admin.ErrorNotImplementedType.String() {
```
dopey, line 81: Won't `adminErr` end up being `**ca.AdminClientError`?

hslatman: The syntax indeed looks a bit weird and I had to try a few times myself too to get it right, but the types are correct. I've tested the output, and that was OK.

I can add a test for this.

dopey: I guess it makes sense. The description of the parameters in the documentation had me confused: https://pkg.go.dev/github.com/pkg/errors#As. In order to store the value in `target` it would have to be a pointer, though.

dopey (approved):
lgtm

PR #635, opened 2022-02-10
Fix principals not including local part of an email

jhult:
So `step ssh login` used to automatically fill in `principals` with both `username@email.com` and `username`. However, it no longer does this. It looks like this may have changed around this commit: https://github.com/smallstep/cli/commit/3fdcde8caf85bf9539dbfb697f564579b9276316

I fixed this by adding back in `createPrincipalsFromSubject`.

Related PRs:

- https://github.com/smallstep/cli/pull/370
- https://github.com/smallstep/cli/pull/621

### DIFF
--- a/command/ssh/login.go
+++ b/command/ssh/login.go
@@ -99,7 +99,7 @@ func loginAction(ctx *cli.Context) error {
 
 	principals := ctx.StringSlice("principal")
 	if subject != "" && len(principals) == 0 {
-		principals = []string{subject}
+		principals = createPrincipalsFromSubject(subject)
 	}
 
 	// Flags


mannp:
> * --principal foo my@email.org

The last two of those do not work for me at all, they both give the same error,

step ssh login --principal local admin@mydomain.org

The request was forbidden by the certificate authority: ssh certificate principals does not match - got [local], want [admin admin@mydomain.org].

maraino:
> step ssh login --principal local admin@mydomain.org

@mannp which provisioner are you using? For example, a JWK provisioner will you allow doing that, an OIDC provisioner will not.

PR #647, opened 2022-03-09
Renew after expiry

maraino:
### Description

This PR adds the necessary changes to step to support the renewal of certificates after expiry.

Requires:
- [x] smallstep/certificates#849


command/ca/provisionerbeta/add.go
```diff
@@ -404,7 +405,8 @@ func addAction(ctx *cli.Context) (err error) {
 			},
 			Enabled: !(ctx.IsSet("ssh") && !ctx.Bool("ssh")),
 		},
-		DisableRenewal: ctx.Bool("disable-renewal"),
+		DisableRenewal:        ctx.Bool("disable-renewal"),
```
dopey, line 408: if we're using `disable-renewal` then shouldn't it be `allow-renewal-xxx`?

command/ca/renew.go
```diff
@@ -558,6 +576,44 @@ func (r *renewer) Daemon(outFile string, next, expiresIn, renewPeriod time.Durat
 	}
 }
 
+// RenewAfterExpiry creates an authorization token with the given certificate
+// and attempts to renew the expired certificate.
+func (r *renewer) RenewAfterExpiry(cert tls.Certificate) (*api.SignResponse, error) {
+	var issuer string
+	if ext, ok := provisioner.GetProvisionerExtension(cert.Leaf); ok {
```
dopey, line 583: how does this work for certificates authenticated by RA but signed by a different CA (e.g., Vault RA)?

command/ca/renew.go
```diff
@@ -558,6 +576,44 @@ func (r *renewer) Daemon(outFile string, next, expiresIn, renewPeriod time.Durat
...
+		x5c = append(x5c, base64.StdEncoding.EncodeToString(b))
+	}
+	if claims.ExtraHeaders == nil {
+		claims.ExtraHeaders = make(map[string]interface{})
+	}
+	claims.ExtraHeaders[jose.X5cInsecureKey] = x5c
+
+	tok, err := claims.Sign("", cert.PrivateKey)
+	if err != nil {
+		return nil, errors.Wrap(err, "error creating authorization token")
```
dopey, line 605: maybe 'error signing authorization token' since the previous error is already for creating token.

command/ca/token.go
```diff
@@ -137,6 +137,12 @@ $ step ca token max@smallstep.com --ssh
 Get a new token for an SSH host certificate:
 '''
 $ step ca token my-remote.hostname --ssh --host
+'''
+
+Generate a renew token and use it in a renew after expiry request:
+'''
+$ TOKEN=$(tep ca token --x5c-cert internal.crt --x5c-key internal.key --renew internal.example.com)
```
dopey, line 144: step (not tep)

flags/flags.go
```diff
@@ -269,6 +269,13 @@ be stored in the 'x5c' header.`,
 be stored in the 'x5c' header.`,
 	}
 
+	// X5cInsecure is a cli.Flag used set the JWT header x5cInsecure instead of
```
dopey, line 272: used to set the

utils/cautils/token_generator.go
```diff
@@ -387,3 +397,56 @@ func generateJWKToken(ctx *cli.Context, p *provisioner.JWK, tokType int, tokAttr
...
+	key, err := pemutil.Read(renewKey)
+	if err != nil {
+		return "", err
+	}
+	if sub != "" && sub != bundle[0].Subject.CommonName {
+		return "", errors.Errorf("positional argument <subject> must match the certificate common name")
+	}
+
+	var issuer string
+	if ext, ok := provisioner.GetProvisionerExtension(bundle[0]); ok {
```
dopey, line 427: What happens if/when the provisioner extension does not exist? (Vault RA)

command/ca/renew.go
```diff
@@ -558,6 +576,44 @@ func (r *renewer) Daemon(outFile string, next, expiresIn, renewPeriod time.Durat
...
 }
 
+// RenewAfterExpiry creates an authorization token with the given certificate
+// and attempts to renew the expired certificate.
+func (r *renewer) RenewAfterExpiry(cert tls.Certificate) (*api.SignResponse, error) {
+	var issuer string
+	if ext, ok := provisioner.GetProvisionerExtension(cert.Leaf); ok {
+		issuer = ext.Name
+	}
+	claims, err := token.NewClaims(
```
dopey, line 586: Why aren't we reusing the code from `generateNewToken` here?

dopey (changes requested):
see comments

command/ca/renew.go
maraino: Renewal is not yet supported on RA mode, but ef532b6 changes this to use a fixed string.

command/ca/renew.go
maraino: Fixed with 30fd159

command/ca/token.go
maraino: Fixed with 71d6b34

flags/flags.go
maraino: Fixed with bfe3c8b

utils/cautils/token_generator.go
maraino: Fixed with 4a1848d

command/ca/renew.go
maraino: Because `generateNewToken` reads the certificate and key from a file pointed by flags. In the `step ca renew` method the certificate and keys are also backed by files, but for example, in a renewal daemon everything is in memory and we write new certificates to a specific file, but after renewal, the user can run a script that might move the file to a different location, so there's no guarantee that the files are going to be there.

command/ca/provisionerbeta/add.go
maraino: Fixed with ff508dc

dopey (approved):
lgtm

PR #650, opened 2022-03-16
Be explicit about the function of the password-file for specific case

dopey:


hslatman (approved):
Approved, but with the note that this seems to be one of very few places where `--password-file` is used like this and not to provide the password to authenticate. At least, that's what I got from a quick scan.

J-Hunter-Hawke (approved):

Issue #611, opened 2022-01-19
Add optional SHA1 fingerprint output for certificates


Delapouite:
Hello

### What would you like to be added

Currently the `step-cli certificate fingerprint foo.pem` command only outputs the fingerprint as a SHA256 hash. It would be convenient to have an option to display a SHA1 fingerprint instead.

### Why this is needed

Azure Iot Hub services display the fingerprints as SHA1 hashes on the Azure portal, so it's not easy to compare both types of fingerprints.

Thanks.

dopey:
Note for self (`step certificate format internal.crt | sha1sum
627780a72c976c84a0c0ddd973e3e4cabe318456  -`)

dopey:
Hey @Delapouite thanks for opening the issue! This is a great idea and I think we'd like to see this implemented with a `--hash` (or maybe `--alg`) flag that would default to sha256 but allow other algs like sha1sum.

We'll open this one up to the community as it should be an easy addition. If anyone is looking to pick this up and needs some guidance, definitely let us know.

Cheers 🍻

Dev4522v:
Hey @dopey! I'd like to pick this issue up. Can you help me out in getting started?

dopey:
Hey @Dev4522v, I'd be happy to. We'll be looking at making edits to this file: https://github.com/smallstep/cli/blob/master/command/certificate/fingerprint.go. What we want to do is add a flag to change the hashing algorithm - we can call the flag `hash`. This is a good flag to base it on: https://github.com/smallstep/cli/blob/6b23538520fc8c720b067bec2fb7fe630f8825f4/flags/flags.go#L20-L37. The default will be sha256. And then just change the sha algorithm used in the action depending on the flag. 

If you have questions about getting started developing in golang - look up `setting up a golang dev environment`. The steps should be pretty simple. I develop on MacOS and I've used `brew` to set up the few things I need.

Let me know if you have additional questions :) Good luck!

Dev4522v:
Thanks for the reply @dopey! I have the golang development environment setup on my mac too. I needed some help understanding the fingerprint.go file. I understand go syntax but I am having difficulty figuring out what the file is exactly doing. Can you explain that or maybe setup a brief call (if convenient)?

Thanks. Looking forward to contributing.

dopey:
[...]

PR #652, opened 2022-03-20
Add optional SHA1 fingerprint output for certificates

1984gobrr:
Add optional SHA1 fingerprint output for certificates has requested in https://github.com/smallstep/cli/issues/611

### DIFF
--- a/command/certificate/fingerprint.go
+++ b/command/certificate/fingerprint.go
@@ -18,7 +18,8 @@ func fingerprintCommand() cli.Command {
 		Action: cli.ActionFunc(fingerprintAction),
 		Usage:  "print the fingerprint of a certificate",
 		UsageText: `**step certificate fingerprint** <crt-file>
-[**--bundle**] [**--roots**=<root-bundle>] [**--servername**=<servername>] [**--format**=<format>]`,
+[**--bundle**] [**--roots**=<root-bundle>] [**--servername**=<servername>] 
+[**--format**=<format>] [**--sha1**] [**--insecure**]`,
 		Description: `**step certificate fingerprint** reads a certificate and prints to STDOUT the
 certificate SHA256 of the raw certificate.
 
@@ -82,7 +83,7 @@ debugging invalid certificates remotely.`,
 			command.FingerprintFormatFlag("hex"),
 			cli.BoolFlag{
 				Name:  "sha1",
-				Usage: `Use the SHA-1 hash algorithm to hash the certificate. Require '--insecure' flag.`,
+				Usage: `Use the SHA-1 hash algorithm to hash the certificate. Requires **--insecure** flag.`,
 			},
 		},
 	}
@@ -101,7 +102,7 @@ func fingerprintAction(ctx *cli.Context) error {
 		insecure   = ctx.Bool("insecure")
 		crtFile    = ctx.Args().First()
 		format     = ctx.String("format")
-		sha1       = ctx.Bool("sha1")
+		useSHA1    = ctx.Bool("sha1")
 	)
 
 	encoding, err := command.GetFingerprintEncoding(format)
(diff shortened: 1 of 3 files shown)


hslatman:
Hi @1984gobrr,

Thank you for your contribution!

We've discussed adding this functionality to the CLI with the team and concluded we'd like to gate usage of `SHA1` in behind the `--insecure` flag, as `SHA1` is cryptographically broken and people generally shouldn't be relying on it. Using `--insecure` still allows one to use the CLI for legacy interoperability cases, but is also a small nudge in the direction of using more secure hashing algorithms. Could you add usage of `--insecure` to your PR?

One other thing I noted is that there are a couple more functionalities in the CLI that also allow providing/switching hash algorithms. The way this PR makes it possible to set the `SHA1` algorithm to use is slightly different than how it's done in the other commands. At this time I don't think we'll need to support a complete spectrum of hash algorithms, so I would propose to change the `--alg=<hash>` flag to a `--sha1` bool flag, similar to what's done in `command/crypto/key/fingerprint.go`. Providing `--sha1` would then override the default SHA-256 algorithm. The additional benefit of this is that users don't have to get the algorithm name right nor do they have to use the right case (lower vs. upper). I realize this is slightly different from what was suggested to be done in the issue (e.g. this way of doing it exists in `step crypto hash`), but I think it's sensible to do for this use case and very similar to what `step crypto key fingerprint` does. Any potential future changes to this may result in doing the hash selection slightly differently, but in that case we'd probably want to make it a bit more reusable and also make it more user friendly in terms of casing and algorithm naming.

command/certificate/fingerprint.go
```diff
@@ -80,6 +80,18 @@ debugging invalid certificates remotely.`,
...
+				Value: "SHA-256",
+				Usage: `The <alg> to hash the certificate. 
+ If unset, default is SHA-256. 
+  
+: <alg> is a case-sensitive string and must be one of: 
+  
+     **SHA-256**
+
+     **SHA-1** `,
+			},
```
maraino, line 94: In `step crypto hash digest` we use `sha1` and `sha256`, we should be consitent

1984gobrr: > Hi @1984gobrr,
> 
> Thank you for your contribution!
> 
> We've discussed adding this functionality to the CLI with the team and concluded we'd like to gate usage of `SHA1` behind the `--insecure` flag, as `SHA1` is cryptographically broken and people generally shouldn't be relying on it. Using `--insecure` still allows one to use the CLI for legacy interoperability cases, but is also a small nudge in the direction of using more secure hashing algorithms. Could you add usage of `--insecure` to your PR?
> 
> One other thing I noted is that there are a couple more functionalities in the CLI that also allow providing/switching hash algorithms. The way this PR makes it possible to set the `SHA1` algorithm to use is slightly different than how it's done in the other commands. At this time I don't think we'll need to support a complete spectrum of hash algorithms, so I would propose to change the `--alg=<hash>` flag to a `--sha1` bool flag, similar to what's done in `command/crypto/key/fingerprint.go`. Providing `--sha1` would then override the default SHA-256 algorithm. The additional benefit of this is that users don't have to get the algorithm name right nor do they have to use the right case (lower vs. upper). I realize this is slightly different from what was suggested to be done in the issue (e.g. this way of doing it exists in `step crypto hash`), but I think it's sensible to do for this use case and very similar to what `step crypto key fingerprint` does. Any potential future changes to this may result in doing the hash selection slightly differently, but in that case we'd probably want to make it a bit more reusable and also make it more user friendly in terms of casing and algorithm naming.

Ok, I will do that today :smile:

1984gobrr: It is now possible to do `step certificate fingerprint --insecure --sha1 https://google.com`

Let me know if any other changes are needed.

PR #654, opened 2022-03-22
Go 1.18 support 

maraino:
### Description

This PR changes the supported versions of Go to 1.17 and 1.18, using Go 1.18 for releases.

### DIFF
--- a/command/certificate/remote_test.go
+++ b/command/certificate/remote_test.go
@@ -66,7 +66,7 @@ func TestGetPeerCertificateServerName(t *testing.T) {
 	tests := map[string]newTest{
 		"sni-disabled-host": {host, "", nil},
 		"sni-enabled-host":  {host, serverName, nil},
-		"sni-disabled-ip":   {addr, "", fmt.Errorf("failed to connect: x509: cannot validate certificate for %s because it doesn't contain any IP SANs", addr)},
+		"sni-disabled-ip":   {addr, "", fmt.Errorf("failed to connect")},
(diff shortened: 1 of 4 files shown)


command/certificate/remote_test.go
```diff
@@ -66,7 +66,7 @@ func TestGetPeerCertificateServerName(t *testing.T) {
 	tests := map[string]newTest{
 		"sni-disabled-host": {host, "", nil},
 		"sni-enabled-host":  {host, serverName, nil},
-		"sni-disabled-ip":   {addr, "", fmt.Errorf("failed to connect: x509: cannot validate certificate for %s because it doesn't contain any IP SANs", addr)},
+		"sni-disabled-ip":   {addr, "", fmt.Errorf("failed to connect")},
```
azazeal, line 69: not quite sure why this is included in the PR - in case it's not a mistake, maybe use `errors.New` instead of `fmt.Errorf` as there's no interpolation taking place.

azazeal (changes requested):

command/certificate/remote_test.go
maraino: This fixed a problem with the unit tests, I don't generally compare error strings, but this test was doing this, and it was failing on 1.18 because the string is now different, I've just removed the course, so it will match only with the wrapped string.

maraino: I changed to use `errors` with e599b7c

azazeal (approved):
LGTM

PR #657, opened 2022-03-24
Also provide units to renew ssh certificates

darix:
Those are based on the cert-renewer unit files but adapted for ssh certificates

the target file allows you to easily stop/restart/disable all instances.


maraino:
@tashian Can you take a look to this?

systemd/step-ssh-cert-renewer@.timer
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=Timer for Smallstep SSH Certificate Renewal for the %I host certificate
+Documentation=https://smallstep.com/docs/step-ca/certificate-authority-server-production
+PartOf=step-ssh-cert-renewer.target
+
+[Timer]
+Persistent=true
+
+; Run the timer unit every 5 minutes.
+OnCalendar=*:1/5
```
tashian, line 10: One issue I've had with our other `systemd` renewer unit is that it logs a LOT because it runs every 5 minutes. If people are using very short-lived certs, they're going to want that sort of frequency. But, even for a 24 hour cert it may be ok to run this every 15 minutes. So, my feeling is that we should bump the default up to 15 minutes and if people want to lower it, they can.

darix: yeah well i stuck with the defaults from the other units.

though in general I agree and in salt i deploy an unit override file with longer default times.

systemd/step-ssh-cert-renewer@.service
```diff
@@ -0,0 +1,30 @@
...
+; (In systemd <242, you can use ExecStartPre= here.)
+ExecCondition=/usr/bin/step ssh needs-renewal ${CERT_LOCATION}
+
+; ExecStart renews the certificate, if ExecStartPre was successful.
+ExecStart=/usr/bin/step ssh renew --force ${CERT_LOCATION} ${KEY_LOCATION}
+
+; Try to reload or restart the systemd service that relies on this cert-renewer
+; If the relying service doesn't exist, forge ahead.
+; (In systemd <229, use `reload-or-try-restart` instead of `try-reload-or-restart`)
+ExecStartPost=/usr/bin/env sh -c "! systemctl --quiet is-enabled sshd.service || systemctl try-reload-or-restart sshd"
```
tashian, line 27: On some distros (including all Ubuntu) this is called `ssh.service`.
So, let's add another `ExecStartPost` for that name.

systemd/step-ssh-cert-renewer@.timer
tashian: Let's increase it to 15 min here (with 5m jitter) and in the `cert-renewer@.timer` as well

systemd/step-ssh-cert-renewer@.timer
```diff
@@ -0,0 +1,19 @@
+[Unit]
```
tashian, line 1: This is set up as a systemd template, which makes sense for `cert-renewer@.*` because there can be many X.509 certs on a systme. But, I think we should rename these SSH renewal timers to `cert-renewer-ssh.*` because there's typically only one SSH server on a machine.

tashian (changes requested):
Hi @darix thanks for this contribution. I think it's close to being ready to merge.
I've added a couple comments.

systemd/step-ssh-cert-renewer@.timer
darix: yes but it would need to run for 3 certificate types.

tashian: In practice, you really only need an ECDSA or Ed25519 cert.
ECDSA has been supported since OpenSSH 5.7 (2011) and Ed25519 since OpenSSH 6.5 (2014).
You just don't get much in terms of broader compatibility by having more cert types, and it adds to your host management burden.
So, I suggest that it be a single renewer rather than a template, and people can always templatize it by renaming it if they need to.

None of this is in our documentation though, and people would probably benefit from docs around this renewer. I'll make an issue for that in [smallstep/docs](https://github.com/smallstep/docs/)

tashian:
Hi @darix, this is ready to merge. Could you please [sign our CLA](https://cla-assistant.io/smallstep/cli?pullRequest=657) so we can accept this PR?

Thank you for your contribution to the project. :D

tashian (approved):
lgtm

maraino:
@darix Although contributions to an Apache 2 software might not require a CLA, see [this blog post](https://apetro.ghost.io/apache-contributors-no-cla/). We can merge this right away if you do sign it, can you?

@tashian is this still ok? Is there any update you can do to our current scripts that will avoid the need of merging this?

darix:
sorry for the delay. signed now.

darix:
I just noticed that this merge request was modified before the merge. I would like to understand the reasoning to only want to support ed25519 keys.

maraino:
@tashian any reason for hardcoding ed25519 key?
```diff
-            CERT_LOCATION=/etc/ssh/ssh_host_%i_key-cert.pub \
-            KEY_LOCATION=/etc/ssh/ssh_host_%i_key
+            CERT_LOCATION=/etc/ssh/ssh_host_ed25519_key-cert.pub \
+            KEY_LOCATION=/etc/ssh/ssh_host_ed25519_key
```

and pardon my ignorance, how is %i populated?

tashian:
@darix sure, here's the reasoning for my hard coding of the Ed25519 key:

https://github.com/smallstep/cli/pull/657#discussion_r836879508

In short, I don't feel it's worth the added complexity of a systemd template to cover more than one host key type.

PR #659, opened 2022-03-25
golangci-lint v1.45.2 avoids panic on Go 1.18

areed:
### Description
Upgrade golangci-lint to v1.45.2 so it works with Go 1.18.


maraino (approved):
LGTM

PR #662, opened 2022-04-18
rename s3 step binaries to be step_<version>_<os>_<arch>

dopey:


maraino (approved):
I guess sounds good as long as any script that depends on these names is also updated

PR #668, opened 2022-04-25
Add commands for managing policies

hslatman:


hslatman:
Still need to add some examples

command/ca/policy/actions/dns.go
```diff
@@ -0,0 +1,144 @@
...
+		panic("no SSH nor X.509 context set")
+	}
+
+	if clictx.Bool("remove") {
+		for _, domain := range args {
+			dns = remove(domain, dns)
+		}
+	} else {
+		dns = append(dns, args...)
+	}
```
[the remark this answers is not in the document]
maraino: We should try international names with SSH hosts certificates

command/ca/policy/actions/uris.go
```diff
@@ -0,0 +1,122 @@
...
+		panic("no SSH nor X.509 context set")
+	}
+
+	if clictx.Bool("remove") {
+		for _, uri := range args {
+			uris = remove(uri, uris)
+		}
+	} else {
+		uris = append(uris, args...)
+	}
```
maraino, line 94: We should validate the uri, it has to be an *url.URL

command/ca/policy/actions/policy.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return updatedPolicy, nil
+}
+
+func prettyPrint(policy *linkedca.Policy) {
+	b, err := json.MarshalIndent(policy, "", "   ")
```
maraino, line 171: protojson will properly show enums?
```golang
b, err := protojson.Marshal(p)
if err != nil {
        return err
}
var buf bytes.Buffer
if err := json.Indent(&buf, b, "", "   "); err != nil {
        return err
}
fmt.Println(buf.String())
```

maraino (changes requested):

internal/command/inject.go
```diff
@@ -0,0 +1,56 @@
...
+type cliCtxKey struct{}
+
+// withCLIContext adds a cli.Context to the Context.
+func withCLIContext(ctx context.Context, clictx *cli.Context) context.Context {
+	return context.WithValue(ctx, cliCtxKey{}, clictx)
+}
+
+// CLIContextFromContext returns a pointer to a cli.Context
+// It panics when the cli.Context is not set.
+func CLIContextFromContext(ctx context.Context) *cli.Context {
```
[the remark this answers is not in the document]
hslatman: I created this to be able to reuse it for other things. Not sure if there will be more, but I think we should leave it as `CLIContextFromContext` for now.

command/ca/policy/actions/cn.go
```diff
@@ -0,0 +1,125 @@
...
+	default:
+		panic("no SSH nor X.509 context set")
+	}
+
+	if clictx.Bool("remove") {
+		for _, domain := range args {
+			commonNames = remove(domain, commonNames)
+		}
+	} else {
+		commonNames = append(commonNames, args...)
```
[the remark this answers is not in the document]
hslatman: Added this to `updatePolicy` using `policy.Deduplicate()`

command/ca/policy/actions/uris.go
hslatman: At this time the rules for URIs only check the domain part. URI values that contain a scheme, for example, will not be allowed to be POSTed. They will result in an error like this:

```
{
    "type": "badRequest",
    "detail": "bad request",
    "message": "error updating authority policy: cannot parse permitted URI domain constraint \"https://www.example.com\": URI domain constraint \"https://www.example.com\" contains scheme (not supported yet)"
}
```

I've implemented it like this, so that users won't be surprised that the scheme is not taken into account (at this time) when requesting a certificate. Any scheme will thus be allowed right now. If we would parse it as an URL here and only take the domain part of it in the rules engine, that's probably not as transparent as this. Support for URI schemes was discussed, but was left out of scope for now.

command/ca/policy/actions/dns.go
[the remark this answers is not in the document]
hslatman: I agree with you that it's good to ensure they're always properly internationalized. However, I think we shouldn't enforce this in the CLI, but should always validate it in (just) the API. The CLI will always call the API, so IMO there's no need to duplicate the logic for these checks. This way control over the rules will be easier, since we don't have to ensure backwards compatibility between the CLI and CA, since there's only one doing the checks. This will also work for other clients interacting with the API.

As for saving the contents in the DB internationalized: that is an option, but I think it's more user friendly to always show the non-internalized names, so that they're more easily readable. If they're valid UTF8 they should end up in the DB OK. Whenever they're requested from the CLI (or a different system), they won't have to be transformed.

hslatman:
For some reason the CI fails with some unused assignments. They're unrelated to this PR. They also seem to exist in https://github.com/smallstep/cli/pull/673. I'm using Go 1.18 locally with the same version of `golangci-lint` as used in CI, which results in no errors. Not sure yet what the cause is.

command/ca/policy/actions/dns.go
```diff
@@ -0,0 +1,182 @@
...
+		case policycontext.IsAllow(ctx):
+			policy.X509.Allow.Dns = dns
+		case policycontext.IsDeny(ctx):
+			policy.X509.Deny.Dns = dns
+		default:
+			panic("no allow nor deny context set")
+		}
+	default:
+		panic("no SSH nor X.509 context set")
+	}
```
maraino, line 174: This switch and the above one can be converted to one moving the remove/append logic to a method and set these slices with the result of that method:
```golang
policy.xxx.Dns = addRemoveArguments(policy.xxx.Dns, args, ctx.Bool("remove"))
```

You can use the same command in other actions.

command/ca/policy/actions/dns.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func dnsAction(ctx context.Context) (err error) {
+
+	clictx := command.CLIContextFromContext(ctx)
+
+	args := clictx.Args()
+	if len(args) == 0 {
+		return errors.New("please provide at least one domain")
```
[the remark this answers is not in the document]
maraino: I won't mention this issue in other commands.

command/ca/policy/actions/dns.go
```diff
@@ -0,0 +1,182 @@
...
+			flags.AdminProvisioner,
+			flags.AdminSubject,
+			flags.PasswordFile,
+			flags.CaURL,
+			flags.Root,
+			flags.Context,
+			cli.BoolFlag{
+				Name:  "remove",
+				Usage: `removes the provided DNS names from the policy instead of adding them`,
+			},
```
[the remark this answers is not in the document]
maraino: I won't mention this issue in other commands.

command/ca/policy/actions/policy.go
```diff
@@ -0,0 +1,197 @@
...
+			return nil, errs.RequiredOrFlag(clictx, "eab-key-reference", "eab-key-id")
+		}
+		policy, err = client.GetACMEPolicy(provisioner, reference, keyID)
+	default:
+		panic("no context for policy retrieval set")
+	}
+
+	if err != nil {
+		var ae = new(ca.AdminClientError)
+		if errors.As(err, &ae) && ae.Type == "notFound" { // TODO: use constant?
```
maraino, line 60: We don't need a pointer to a pointer here.
```suggestion
		var ae ca.AdminClientError
		if errors.As(err, &ae) && ae.Type == "notFound" { // TODO: use constant?
```

command/ca/policy/policycontext/context.go
```diff
@@ -0,0 +1,186 @@
...
+// WithAllow returns a context.Context with allow policy set.
+func WithAllow(ctx context.Context) context.Context {
+	return context.WithValue(ctx, policyTypeContextKey{}, allowType)
+}
+
+// IsAllow returns if the context.Context has allow set.
+func IsAllow(ctx context.Context) bool {
+	if ctx == nil {
+		return false
+	}
```
maraino, line 130: Perhaps we should avoid sending nil here.

command/ca/policy/policycontext/context.go
```diff
@@ -0,0 +1,186 @@
...
+// WithDeny returns a context.Context with deny set.
+func WithDeny(ctx context.Context) context.Context {
+	return context.WithValue(ctx, policyTypeContextKey{}, denyType)
+}
+
+// IsDeny returns if context.Context has deny set.
+func IsDeny(ctx context.Context) bool {
+	if ctx == nil {
+		return false
+	}
```
maraino, line 144: ditto

command/ca/policy/policycontext/context.go
```diff
@@ -0,0 +1,186 @@
...
+	v, _ := ctx.Value(policyTypeContextKey{}).(policyType)
+	return v == denyType
+}
+
+func GetPrefixedCommandUsage(ctx context.Context, commandName string) string {
+	usage := "step ca policy"
```
maraino, line 150: You should be able to get `step ca policy` using the `cli.Context` with `ctx.App.Name`, and the actual command e.g. `authority` with `ctx.Command.Name`

maraino (changes requested):

command/ca/policy/actions/cn.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func commonNamesAction(ctx context.Context) (err error) {
+
+	clictx := command.CLIContextFromContext(ctx)
+
+	args := clictx.Args()
+	if len(args) == 0 {
+		return errors.New("please provide at least one name")
```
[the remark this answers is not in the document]
hslatman: Simply didn't come to mind

command/ca/policy/actions/cn.go
```diff
@@ -0,0 +1,141 @@
...
+		return errors.New("SSH user policy does not support Common Names")
+	case policycontext.IsX509Policy(ctx):
+		switch {
+		case policycontext.IsAllow(ctx):
+			policy.X509.Allow.CommonNames = commonNames
+		case policycontext.IsDeny(ctx):
+			policy.X509.Deny.CommonNames = commonNames
+		default:
+			panic("no allow nor deny context set")
+		}
```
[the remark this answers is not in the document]
hslatman: Somehow I thought a linter was setup to catch switches without a default. That doesn't seem to be the case. This whole block is now fully removed, though.

command/ca/policy/actions/dns.go
hslatman: Thanks. That cleaned it up 🙂

command/ca/policy/actions/policy.go
hslatman: Had to go for 

```golang
var ae *ca.AdminClientError
```

Using my VS Code linter (`errorsas`)

command/ca/policy/policycontext/context.go
hslatman: The only way to fully prevent that is to check that no-one sending a nil here, either through static analysis or code review. I've chosen to `panic` now, since I consider this a programmer error if it happens.

command/ca/policy/policycontext/context.go
hslatman: i.d.

command/ca/policy/policycontext/context.go
hslatman: Something like that was what I had in mind, but note that we have a normal `context.Context` here; not a `cli.Context`. The `ctx` does contain a `cli.Context`, but only after an action is called (when it's injected just before the command is ran), so not when the help is printed, which is when the `GetPrefixedCommandUsage` is used.

I think the pattern of using `context.Context` can (and should) be used more broadly, the main reason being making the CLI more testable, but we would need to refactor some more parts of the CLI to be able to do so. I don't think what I currently have implemented is perfect, but it gets the job done for now.

command/ca/policy/actions/policy.go
maraino: If it's not initialized a pointer to a pointer makes sense.

maraino (approved):
Looks good to me, but I would remove the new panics in those contexts, `ctx.Value` will panic too if nil is passed.

command/ca/policy/policycontext/context.go
maraino: As I commented, a panic will also happen without that condition, so I would remove it.

hslatman:
> For some reason the CI fails with some unused assignments. They're unrelated to this PR. They also seem to exist in #673. I'm using Go 1.18 locally with the same version of `golangci-lint` as used in CI, which results in no errors. Not sure yet what the cause is.

This seems to have been caused by the replace of `github.com/smallstep/certificates`. After relying on the latest version the CI started to run linters and tests correctly again.

PR #669, opened 2022-04-26
use fingerprint as friendly name for pkcs12 truststore

matthyx:
### Description
When using pkcs12 format to generate a truststore, certificates with the same Subject are colliding, causing overrides.
See https://github.com/SSLMate/go-pkcs12/blob/master/pkcs12.go#L558-L562


hslatman:
@matthyx: thank you for your PR. I'll have a look soon.

hslatman:
@matthyx: I checked your PR using `keytool` on macOS. I was able to reproduce your issue and also found your PR at https://github.com/SSLMate/go-pkcs12/pull/30 and the related issues.

My proposal would be to change it to something like this:

```golang
// Set a (unique) `friendlyName`, preventing keystores and systems reading keystores that identify 
// certificates by (just) their Subject from deduplicating different certificates.
FriendlyName: fmt.Sprintf(
    "%s - %s",
    cert.Subject.String(),
    fingerprint.Fingerprint(cert.Raw, fingerprint.WithHash(crypto.SHA256)),
),
``` 

This mimics the original logic by including the Subject (IMO a more friendly name, depending on where the friendly name is read, of course), but makes the name unique by including the SHA256 hash. The `fingerprint.WithHash(...)` is required, because with your current change the hex representation of the full cert will be the alias.

With the above changes, the output looks like this (tested using leaf certificates, but I don't think that matters in this case):

```console
$ keytool -list -v -keystore store-combined.p12 -storepass "" -storetype PKCS12
...
Alias name: cn=test.localhost - 745dd23ca22727b24d5e0f76dae82d88ca599800bd99d0ff91d809cdef826553
Creation date: 3 May 2022
Entry type: trustedCertEntry
...
```

I saw that the the SHA256 hash is also encoded in the trust store, but using a slightly different format:

```
Certificate fingerprints:
    SHA1: 46:EB:EB:1D:B1:C5:E3:0F:39:FF:9E:6B:7C:A7:FF:F5:B6:AE:E2:E3
    SHA256: 74:5D:D2:3C:A2:27:27:B2:4D:5E:0F:76:DA:E8:2D:88:CA:59:98:00:BD:99:D0:FF:91:D8:09:CD:EF:82:65:53
```

We could add that type of "encoding" to the `fingerprint` package. I've seen the format used before, so it makes sense to be able to create a similar format, IMO. It's not required for this PR, though.

matthyx:
> 

Thanks @hslatman for your review, I have updated the PR.

hslatman:
Thank you for your contribution, @matthyx! 😄

PR #670, opened 2022-04-27
Allow to inspect ssh certificates without the type

maraino:
### Description

SSH certificates are marshaled to disk with the format `type base64-certificate comment`, but we only send through the wire the certificate itself, currently we need to prepend a type (or any fake string) to be able to inspect the certificate itself. This PR solves this and attempts to decode the base64-certificate if the previous method failed.

### DIFF
--- a/command/ssh/inspect.go
+++ b/command/ssh/inspect.go
@@ -1,6 +1,8 @@
 package ssh
 
 import (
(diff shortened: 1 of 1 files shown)


dopey (approved):
lgtm

PR #673, opened 2022-04-30
Combine the beta and regular provisioner commands

### LINKED ISSUE #676: Ability to add database backed JWK provisioners without the private key.

dopey:
#### Name of feature:
Improvement to provisioner CRUD commands

Use the same commands to perform CRUD on provisioners - rather than having a separate `beta` subcommand with different flags.

Uniform implementation and execution. Prevents further drift and provides a unified experience for operators. Also, prevents the need for different sets of documentation that currently confuses users.

Yes, the existing provisioner documentation.

All of them.

N/A

Fixes https://github.com/smallstep/certificates/issues/886
Fixes #676 

💔Thank you!


command/ca/provisioner/provisioner.go
```diff
@@ -72,3 +88,304 @@ $ step ca provisioner remove max@smallstep.com --kid 1234 --ca-config ca.json
...
+		for i, elem := range list {
+			if elem == rem {
+				list[i] = list[len(list)-1]
+				list = list[:len(list)-1]
+				break
+			}
+		}
+	}
+	return list
+}
```
maraino, line 150: This method doesn't remove duplicated elements in `list`. Is this something that can happen?

command/ca/provisioner/provisioner.go
```diff
@@ -72,3 +88,304 @@ $ step ca provisioner remove max@smallstep.com --kid 1234 --ca-config ca.json
...
+		Name:  "allow-renewal-after-expiry",
+		Usage: `Allow renewals for expired certificates generated by this provisioner.`,
+	}
+	enableX509Flag = cli.BoolFlag{
+		Name:  "x509",
+		Usage: `Enable provisioning of x509 certificates.`,
+	}
+	enableSSHFlag = cli.BoolFlag{
+		Name:  "ssh",
+		Usage: `Enable provisioning of ssh certificates.`,
```
maraino, line 219: Document that this is enabled by default, at least I think this is what you did, and say how to disable it `--ssh=false`

command/ca/provisioner/provisioner.go
```diff
@@ -72,3 +88,304 @@ $ step ca provisioner remove max@smallstep.com --kid 1234 --ca-config ca.json
...
+		Usage: `Disable renewal for all certificates generated by this provisioner.`,
+	}
+	allowRenewalAfterExpiryFlag = cli.BoolFlag{
+		Name:  "allow-renewal-after-expiry",
+		Usage: `Allow renewals for expired certificates generated by this provisioner.`,
+	}
+	enableX509Flag = cli.BoolFlag{
+		Name:  "x509",
+		Usage: `Enable provisioning of x509 certificates.`,
+	}
```
maraino, line 216: Although it's implemented in the CA, we should just comment this for now.

command/ca/provisioner/add.go
```diff
@@ -77,23 +109,51 @@ func addCommand() cli.Command {
...
-				Name: "create",
-				Usage: `Create a new ECDSA key pair using curve P-256 and populate a new JWK
-provisioner with it.`,
+				Name:  "create",
+				Usage: `Create the JWK key pair for the provisioner.`,
 			},
-			cli.BoolFlag{
-				Name:  "ssh",
-				Usage: `Enable SSH on the new provisioners.`,
+			cli.StringFlag{
```
maraino, line 149: We have variables for most of the flags, we should create them for all of them.

command/ca/provisioner/provisioner.go
```diff
@@ -72,3 +88,304 @@ $ step ca provisioner remove max@smallstep.com --kid 1234 --ca-config ca.json
...
+		Name:  "ssh-template",
+		Usage: `The x509 certificate template <file>, a JSON representation of the certificate to create.`,
+	}
+	sshTemplateDataFlag = cli.StringFlag{
+		Name:  "ssh-template-data",
+		Usage: `The ssh certificate template data <file>, a JSON map of data that can be used by the certificate template.`,
+	}
+	x509MinDurFlag = cli.StringFlag{
+		Name:  "x509-min-dur",
+		Usage: `The minimum <duration> for an x509 certificate generated by this provisioner.`,
```
maraino, line 171: Document on at least one of the flags, the first one that appears on the list, or in the command what `duration` is and its format.

command/ca/provisioner/provisioner.go
```diff
@@ -72,3 +88,304 @@ $ step ca provisioner remove max@smallstep.com --kid 1234 --ca-config ca.json
...
+		Usage: `The maximum <duration> to grant a certificate in AWS and GCP provisioners.
+A <duration> is sequence of decimal numbers, each with optional fraction and a
+unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns",
+"us" (or "µs"), "ms", "s", "m", "h".`,
+	}
+	iidRootsFlag = cli.StringFlag{
+		Name: "iid-roots",
+		Usage: `The <file> containing the certificates used to validate the
+instance identity documents in AWS.`,
+	}
```
maraino, line 339: This is currently not supported, and if we want to support it as it is it should be the file on the ca server containing the certificates used to ...

command/ca/provisioner/remove.go
```diff
@@ -1,211 +1,56 @@
...
-## POSITIONAL ARGUMENTS
-
-<name>
-: The name field of the provisioner(s) to be removed.
+		Description: `**step ca provisioner remove** removes a provisioner from the CA configuration.
 
 ## EXAMPLES
 
-Remove all provisioners associated with a given name (max@smallstep.com):
+Remove provisioner by name:
```
maraino, line 34: Add example with `--ca-config` too.

command/ca/provisioner/remove.go
```diff
@@ -1,211 +1,56 @@
...
 	return cli.Command{
 		Name:   "remove",
 		Action: cli.ActionFunc(removeAction),
-		Usage:  "remove one, or more, provisioners from the CA configuration",
+		Usage:  "remove a provisioner from the CA configuration",
 		UsageText: `**step ca provisioner remove** <name>
-[**--kid**=<kid>] [**--config**=<file>] [**--all**]`,
+[**--admin-cert**=<file>] [**--admin-key**=<file>] [**--admin-provisioner**=<name>]
+[**--admin-subject**=<subject>] [**--password-file**=<file>] [**--ca-url**=<uri>]
+[**--root**=<file>] [**--context**=<name>]`,
```
maraino, line 18: `--ca-config` is missing.

command/ca/provisioner/update.go
```diff
@@ -18,35 +16,33 @@ import (
...
-	"google.golang.org/protobuf/encoding/protojson"
 )
 
 func updateCommand() cli.Command {
 	return cli.Command{
 		Name:   "update",
 		Action: cli.ActionFunc(updateAction),
 		Usage:  "update a provisioner",
-		UsageText: `**step beta ca provisioner update** <name> [**--public-key**=<file>]
+		UsageText: `**step ca provisioner update** <name> [**--public-key**=<file>]
```
maraino, line 30: `--ca-config` and at least one example with it.

maraino (changes requested):
Mostly ok, but I added a few comments that should be easy to fix.

command/ca/provisioner/add.go
```diff
@@ -190,256 +195,311 @@ will be accepted.`,
...
+				Min:     ctx.String("ssh-user-min-dur"),
+				Max:     ctx.String("ssh-user-max-dur"),
+				Default: ctx.String("ssh-user-default-dur"),
+			},
+			HostDurations: &linkedca.Durations{
+				Min:     ctx.String("ssh-host-min-dur"),
+				Max:     ctx.String("ssh-host-max-dur"),
+				Default: ctx.String("ssh-host-default-dur"),
+			},
+			Enabled: !(ctx.IsSet("ssh") && !ctx.Bool("ssh")),
```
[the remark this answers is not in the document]
dopey: I'm not sure. It looks like we were enabling it by default in `step beta ca provisioner add`. I can't remember if that was a conscious decision.

command/ca/provisioner/nodb.go
```diff
@@ -0,0 +1,172 @@
...
+)
+
+type nodbCRUD struct {
+	configFile string
+	ctx        context.Context
+	auth       *authority.Authority
+}
+
+func newAdminAPIDisabledClient(ctx context.Context, cfg *config.Config, cfgFile string) (*nodbCRUD, error) {
+	a, err := authority.FromOptions(authority.WithConfig(cfg), authority.WithAdminDB(admin.NewNoDB()))
```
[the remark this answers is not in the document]
dopey: Both of those methods initialize the authority. I explicitly do not want to initialize the authority. I just want to initialize the config and then load the admin/provisioner caches.

command/ca/provisioner/provisioner.go
dopey: I suppose it can now, because a user could manually add multiple elements to the config. And I'm not seeing any de-duping on `add`.

command/ca/provisioner/add.go
```diff
@@ -190,256 +195,311 @@ will be accepted.`,
...
-		list, err = addK8sSAProvisioner(ctx, name, provMap)
-	case provisioner.TypeSSHPOP:
-		list, err = addSSHPOPProvisioner(ctx, name, provMap)
+	p.Claims = &linkedca.Claims{
+		X509: &linkedca.X509Claims{
+			Durations: &linkedca.Durations{
+				Min:     ctx.String("x509-min-dur"),
+				Max:     ctx.String("x509-max-dur"),
+				Default: ctx.String("x509-default-dur"),
+			},
```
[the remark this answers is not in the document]
dopey: Why not rely on the validation done in certificates by the Claimer type: https://github.com/smallstep/certificates/blob/4cb74e7d8ba5a70b2b5c6c00f3907ad0899524fe/authority/provisioner/claims.go#L220. 

That get's called as part of both store and update. https://github.com/smallstep/certificates/blob/4cb74e7d8ba5a70b2b5c6c00f3907ad0899524fe/authority/provisioners.go#L189-L191. The errors that are returned are clear and to the point.

I don't see the sense in reimplementing the exact same logic in the cli. 

Instead we should maybe add SSH duration validation to the Claimer.Validate() func.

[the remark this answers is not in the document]
maraino: It's ok to relay on the CA too, the only thing is we can give better errors if we validate some things on the cli.

command/ca/provisioner/nodb.go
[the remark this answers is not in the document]
maraino: What about an option to avoid the initialization of the authority?

command/ca/provisioner/provisioner.go
maraino: It might be confusing, but I guess that was the previous behavior.

command/ca/provisioner/add.go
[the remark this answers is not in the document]
dopey: In this specific case, I feel we're already giving very detailed errors in the CA.

command/ca/provisioner/add.go
```diff
@@ -703,96 +676,124 @@ func addK8sSAProvisioner(ctx *cli.Context, name string, provMap map[string]bool)
...
-	return
+	return &linkedca.ProvisionerDetails{
+		Data: &linkedca.ProvisionerDetails_AWS{
+			AWS: &linkedca.AWSProvisioner{
+				Accounts:               ctx.StringSlice("aws-account"),
+				DisableCustomSans:      ctx.Bool("disable-custom-sans"),
+				DisableTrustOnFirstUse: ctx.Bool("disable-trust-on-first-use"),
+				InstanceAge:            d,
+				// TODO IID Roots
+				// IIDRoots:               ctx.String("iid-roots"),
```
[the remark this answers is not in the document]
dopey: Let's address this in a different PR.

command/ca/provisioner/nodb.go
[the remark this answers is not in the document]
dopey: I changed it to an option.

command/ca/provisioner/update.go
```diff
@@ -0,0 +1,875 @@
...
+				// Encrypt JWK
+				opts := []jose.Option{}
+				if ctx.IsSet("password-file") {
+					opts = append(opts, jose.WithPasswordFile(ctx.String("password-file")))
+				}
+				jwe, err = jose.EncryptJWK(privjwk, opts...)
+				if err != nil {
+					return err
+				}
+			}
```
maraino, line 552: If we're decrypting and encrypting, at least we should verify it. I would be ok trusting the user with the input as long as it's a JWE, without even decrypting it.

command/ca/provisioner/update.go
```diff
@@ -0,0 +1,875 @@
...
+	disableEABSet := ctx.IsSet("disable-eab")
+	if requireEABSet && disableEABSet {
+		return errs.IncompatibleFlagWithFlag(ctx, "require-eab", "disable-eab")
+	}
+	if requireEABSet {
+		details.RequireEab = ctx.Bool("require-eab")
+	}
+	if disableEABSet {
+		details.RequireEab = false
+	}
```
maraino, line 596: Not saying that we should do it, but if we want to reduce boolean flags, we can do  `--require-eab=false` and remove `--disable-eab`. 

In your implementation `--disable-eab` == `--disable-eab=false`, it should be ok.

command/ca/provisioner/update.go
```diff
@@ -0,0 +1,875 @@
...
+	if ctx.IsSet("azure-subscription-id") {
+		details.SubscriptionIds = append(details.SubscriptionIds, ctx.StringSlice("azure-subscription-id")...)
+	}
+	if ctx.IsSet("remove-azure-object-id") {
+		details.ObjectIds = removeElements(details.ObjectIds, ctx.StringSlice("remove-azure-object-id"))
+	}
+	if ctx.IsSet("azure-object-id") {
+		details.ObjectIds = append(details.ObjectIds, ctx.StringSlice("azure-object-id")...)
+	}
+	return nil
```
maraino, line 810: Audience is missing

command/ca/provisioner/caConfigClient.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, errors.Wrapf(err, "error loading authority")
+	}
+	client := &caConfigClient{
+		configFile: cfgFile,
+		ctx:        ctx,
+		auth:       a,
+	}
+	if err := client.auth.ReloadAdminResources(ctx); err != nil {
+		return nil, errors.Wrapf(err, "error loading provisioners from config")
+	}
```
maraino, line 80: Will this fail if we want to change a bad OIDC configuration present in the ca.json?

command/ca/provisioner/caConfigClient.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+type caConfigClient struct {
+	configFile string
+	ctx        context.Context
+	auth       *authority.Authority
+}
+
+func newCaConfigClient(ctx context.Context, cfg *config.Config, cfgFile string) (*caConfigClient, error) {
+	a, err := authority.New(cfg, authority.WithAdminDB(newNoDB()), authority.WithSkipInit())
```
maraino, line 69: Just an idea, instead of this client, would it make sense to create an admin db that edits the ca.json?

maraino (changes requested):

command/ca/ca.go
```diff
@@ -165,7 +166,7 @@ func BetaCommand() cli.Command {
 		Description: `**step beta ca** enables beta access to new step-ca APIs. These
 commands may change, disappear, or be promoted to a different subcommand in the future.`,
 		Subcommands: cli.Commands{
-			admin.Command(),
+			adminbeta.Command(),
```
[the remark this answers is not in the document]
dopey: So, to me, the question is what difference does it make when I plan to just complete remove these directories with the next release. It's more work to point to the new commands from `adminbeta` - and it's completely unnecessary because we will literally remove the files completely.

With `provisionerbeta` I'd rather not point to the new commands because the new commands sometimes use ca.json (if it exists), and that could actually break something for a customer. The old commands always use an API client.

command/ca/provisioner/caConfigClient.go
dopey: It's an interesting idea, but then you would have to http connect to an instance of `step-ca` to make changes to the configuration, if I'm understanding you correctly. Some of our users make modifications to the `ca.json` offline. We would break that functionality.

command/ca/provisioner/caConfigClient.go
dopey: Good catch! I think we have two general options here:

1) I can build the provisioner collection my self and then set it on the Authority using something like With ProvisionerCollection - OR -
2) I can lazy load the provisioner - which could have the benefit of fixing the bug where the CA becomes unusable if, for some reason, the OIDC configuration changes and breaks the currently stored one.

command/ca/provisioner/caConfigClient.go
maraino: I didn't mean that, I meant that having the implementation on the `step-ca` code base might make more sense.

command/ca/provisioner/update.go
```diff
@@ -0,0 +1,890 @@
...
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net/url"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/smallstep/certificates/ca"
+	"github.com/smallstep/cli/crypto/pemutil"
```
maraino, line 15: We should be using `go.step.sm/crypto/pemutil`

maraino (commented):
Just a few minor improvements.

maraino (approved):
lgtm

maraino (approved):
lgtm

Yes, `step ca provisioner add cicd --type JWK --create --no-private-key` didn't make any sense.

PR #677, opened 2022-05-10
Remove unused command context and env vars code

hslatman:
The logic in this file was (largely) moved to https://github.com/smallstep/cli-utils before. No references to the functions I removed.


command/command.go
```diff
@@ -1,215 +1,20 @@
...
-
-// Retrieve returns all commands
-func Retrieve() []cli.Command {
-	return cmds
-}
-
 // ActionFunc returns a cli.ActionFunc that stores the context.
 func ActionFunc(fn cli.ActionFunc) cli.ActionFunc {
 	return func(ctx *cli.Context) error {
-		currentContext = ctx
```
dopey, line 46: Why not use the ActionFunc in cli-utils?

dopey (approved):
In general, everything looks good. I just had one question.

command/command.go
hslatman: This `ActionFunc` was still in use by two commands, so I left that as is. It seems to be the same function; just not refactored when the function was extracted to `cli-utils`, so I can probably clean this out too.

In general, it's a matter of discoverability, clarity and remembering the existence of these internal, shared libraries that affect whether I'll use them or not. It's easier to find code in the same repository, than when having to look for them in dependencies.

hslatman:
@dopey: removed the `ActionFunc`.

dopey:
lgtm

PR #680, opened 2022-05-19
Add --password-file flag for offline mode without password input

dopey:
#### Name of feature:

`--password-file` flag for offline mode

Allows use of `offline` mode without any additional command line input. Useful for scripting.

It's fairly niche, and I don't think we need any additional documentation.

All

Fixes https://github.com/smallstep/cli/pull/673

### DIFF
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -20,13 +20,15 @@ func certificateCommand() cli.Command {
 		Action: command.ActionFunc(certificateAction),
 		Usage:  "generate a new private key and certificate signed by the root certificate",
 		UsageText: `**step ca certificate** <subject> <crt-file> <key-file>
-[**--token**=<token>]  [**--issuer**=<name>] [**--not-before**=<time|duration>]
-[**--not-after**=<time|duration>] [**--san**=<SAN>] [**--set**=<key=value>]
-[**--set-file**=<file>] [**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
+[**--token**=<token>]  [**--issuer**=<name>] [**--provisioner-password-file**=<file>]
+[**--not-before**=<time|duration>] [**--not-after**=<time|duration>]
+[**--san**=<SAN>] [**--set**=<key=value>] [**--set-file**=<file>]
+[**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
 [**--contact**=<email>] [**--http-listen**=<address>] [**--bundle**]
 [**--kty**=<type>] [**--curve**=<curve>] [**--size**=<size>] [**--console**]
 [**--x5c-cert**=<file>] [**--x5c-key**=<file>] [**--k8ssa-token-path**=<file>]
-[**--ca-url**=<uri>] [**--root**=<file>] [**--context**=<name>]`,
+[**--offline**] [**--password-file**] [**--ca-url**=<uri>] [**--root**=<file>]
+[**--context**=<name>]`,
 		Description: `**step ca certificate** command generates a new certificate pair
 
 ## POSITIONAL ARGUMENTS
@@ -157,6 +159,7 @@ multiple SANs. The '--san' flag and the '--token' flag are mutually exclusive.`,
 			flags.NotBefore,
 			flags.Force,
 			flags.Offline,
+			flags.PasswordFile,
 			consoleFlag,
 			flags.X5cCert,
 			flags.X5cKey,
--- a/command/ca/certificate.go
+++ b/command/ca/certificate.go
@@ -75,6 +75,16 @@ files, certificates, and keys created with **step ca init**:
 $ step ca certificate --offline internal.example.com internal.crt internal.key
 '''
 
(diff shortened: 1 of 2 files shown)


command/ca/certificate.go
```diff
@@ -20,13 +20,15 @@ func certificateCommand() cli.Command {
...
-[**--set-file**=<file>] [**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
+[**--token**=<token>]  [**--issuer**=<name>] [**--provisioner-password-file**=<file>]
+[**--not-before**=<time|duration>] [**--not-after**=<time|duration>]
+[**--san**=<SAN>] [**--set**=<key=value>] [**--set-file**=<file>]
+[**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
 [**--contact**=<email>] [**--http-listen**=<address>] [**--bundle**]
 [**--kty**=<type>] [**--curve**=<curve>] [**--size**=<size>] [**--console**]
 [**--x5c-cert**=<file>] [**--x5c-key**=<file>] [**--k8ssa-token-path**=<file>]
-[**--ca-url**=<uri>] [**--root**=<file>] [**--context**=<name>]`,
+[**--offline**] [**--password-file**] [**--ca-url**=<uri>] [**--root**=<file>]
```
hslatman, line 30:
```suggestion
[**--offline**] [**--password-file**=<file>] [**--ca-url**=<uri>] [**--root**=<file>]
```

command/ca/certificate.go
```diff
@@ -20,13 +20,15 @@ func certificateCommand() cli.Command {
 		Action: command.ActionFunc(certificateAction),
 		Usage:  "generate a new private key and certificate signed by the root certificate",
 		UsageText: `**step ca certificate** <subject> <crt-file> <key-file>
-[**--token**=<token>]  [**--issuer**=<name>] [**--not-before**=<time|duration>]
-[**--not-after**=<time|duration>] [**--san**=<SAN>] [**--set**=<key=value>]
-[**--set-file**=<file>] [**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
+[**--token**=<token>]  [**--issuer**=<name>] [**--provisioner-password-file**=<file>]
+[**--not-before**=<time|duration>] [**--not-after**=<time|duration>]
+[**--san**=<SAN>] [**--set**=<key=value>] [**--set-file**=<file>]
+[**--acme**=<file>] [**--standalone**] [**--webroot**=<file>]
```
hslatman, line 26:
```suggestion
[**--acme**=<uri>] [**--standalone**] [**--webroot**=<file>]
```

hslatman (commented):
To be consistent the `--password-file` flag should also be usable in `--offline` mode for other commands. A quick scan reveals `renew`, `revoke`, `sign` and `token` can be used in offline mode. The `renew` command seems to use the `--password-file` for a different purpose. That said, it's of course a fairly niche feature.

dopey:
Yea ... good feedback @hslatman. Was just trynna get something out quickly to "unblock" that particular user. 

As you mentioned, we are already using `password-file` in `renew` and it's a very generic flag name that we use in lots of other places. It's definitely not clear in this scenario that it should be used to decrypt the CA intermediate private key.

What if we used something like `ca-x509-password-file`?

Btw, the problem with changing the flag name is that we'll have to update how this works as well: https://github.com/smallstep/cli/blob/75a695ea753d3e8a265c8f8fe11527cc9601ca98/utils/cautils/offline.go#L59-L66. Maybe that method will have to take a ternary options with a `WithX509PasswordFile(xxx)`, and if it's not set, just use the default `--password-file`. @maraino what do you think?

maraino (commented):
If we need to support different commands we can use the 
`--provisioner-password-file` and `--password-file` distinction. I believe the code that creates the token already uses `--provisioner-password-file` if it exists, and if not it fallback to `--password-file`. 

We can keep this logic and add the `--provisioner-password-file` where necessary.

maraino (approved):
This change looks ok, and we should add the other changes in a different PR

dopey:
@mariano in the case of `step ca renew --offline`, as @hslatman mentioned, it's already using the `--password-file` flag to decrypt the private key (if encrypted). So we can't reuse `--password-file` for the CA intermediate key and we can't use `--provisioner-password-file` because it's completely unrelated. We should figure out how to address this scenario first, because I think the flag name should be uniform across all the `offline` enabled methods.

maraino:
> @maraino in the case of step ca renew --offline, as @hslatman mentioned, it's already using the --password-file flag to decrypt the private key (if encrypted). So we can't reuse --password-file for the CA intermediate key and we can't use --provisioner-password-file because it's completely unrelated. We should figure out how to address this scenario first, because I think the flag name should be uniform across all the offline enabled methods.

I wasn't talking about the renew case, because we generally do mTLS and there's no need for a provisioner password, but others might.

And by the way, in commands with a `--token` flag, like `certificate` or `revoke` there is the workaround of passing the token with `--token $(step ca token ...)`, we will never ask for a provisioner if a token is passed.

PR #681, opened 2022-05-31
Sort context list alphabetically

hslatman:
#### Name of feature:

Sort context list alphabetically

Hard to find contexts when there's no order

I made it default to sorting alphabetically (case-insensitive), because I think that makes sense for most, if not all, use cases.


dopey (approved):
lgtm

PR #683, opened 2022-06-02
Fix signals and offline mode for js/wasm

hslatman:


hslatman:
There's a (recent) open PR for `js/wasm` target for `newrelic/go-agent`: https://github.com/newrelic/go-agent/pull/478. That should work for us too. It's an indirect dependency through `github.com/smallstep/certificates/monitoring`.

PR #686, opened 2022-06-08
Check input password value to not be empty on keypair creation

hslatman:
An empty password doesn't make sense, certainly not given there's the `--no-password` option. This commit adds the empty validation check to the password prompt, always requiring a password to be provided by the user. 

An alternative would be to generate a password if the value is empty. Both methods are currently used in the CLI. The ones that generate a password are mostly used for `ca` commands. That's why I chose to validate the value to not be empty instead of generating one, as `crypto keypair` has nothing to do with the CA directly.


dopey (approved):
lgtm

PR #702, opened 2022-07-15
Force renewal with token

### LINKED ISSUE #701: Add flag to renew without using mTLS on step ca renew
### Description

`step ca renew` uses a token if a certificate is expired, but this token can also renew non-expired certificates, we should add a flag to `step ca renew` to force the use of the token. 

Currently, the only way to do it is using curl directly like this:

```console
curl -s -X POST -H "Authorization: Bearer $(step ca token --renew --x5c-cert internal.crt --x5c-key internal.key internal.smallstep.com)" https://ca.smallstep.com:9000/1.0/renew
```

This can be useful for example when `step-ca` is behind a proxy.

maraino:
### Description

This PR allows disabling the mTLS flow and using the token authorization flow on step ca renew. Fixes #701 

```console
$ step ca renew --force --mtls=false mariano.crt mariano.key
Your certificate has been saved in mariano.crt.
```

### DIFF
--- a/command/ca/renew.go
+++ b/command/ca/renew.go
@@ -62,9 +62,11 @@ fixed period can be set with the **--renew-period** flag.
 The **--daemon** flag can be combined with **--pid**, **--signal**, or **--exec**
 to provide certificate reloads on your services.
 
-The renew command uses mTLS to get the new certificate, but in deployments where
-step-ca is behind a proxy, mTLS might not be possible. To circumvent that, use
-**--mtls=false** to force the token authorization flow.
+The renew command uses mTLS (by default) to authenticate to the step-ca API.
+However, there are scenarios where mTLS is not an option - step-ca is behind a
+proxy or the leaf certificate is not configured to do client authentication. To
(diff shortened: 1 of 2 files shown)


command/ca/renew.go
```diff
@@ -62,6 +62,10 @@ fixed period can be set with the **--renew-period** flag.
 The **--daemon** flag can be combined with **--pid**, **--signal**, or **--exec**
 to provide certificate reloads on your services.
 
+The renew command uses mTLS to get the new certificate, but in deployments where
+step-ca is behind a proxy, mTLS might not be possible. To circumvent that, use
+**--mtls=false** to force the token authorization flow.
```
dopey, line 67: The renew command uses mTLS (by default) to authenticate to the step-ca API. However, there are scenarios where mTLS is not an option - step-ca is behind a proxy or the leaf certificate is not configured to do client authentication. To circumvent the default mTLS authentication use `--mtls=false` to force a flow that uses X5C token based authentication.

maraino: Fixed with d2463ab

dopey (approved):
lgtm

PR #712, opened 2022-08-04
Add basic ACME `device-attest-01` support

hslatman:
The `--attestation-uri` and `--attestation-ca-url` flags can be used when requesting a
certificate from `step-ca`. The ACME `device-attest-01` challenge method will be 
used to authorize the certificate request when `--attestation-uri` indicates usage
of the TPM KMS (`tpmkms`).

The example below will perform the ACME `device-attest-01` challenge verification
by including an AK certificate in the response to the ACME server, configured with 
an ACME provisioner called `attestation`.

```console
step ca certificate my-unique-id id.crt --attestation-uri tpmkms:name=first-key --provisioner attestation --attestation-ca-url https://attca:port
```

If no valid AK certificate is available, the client will try to enroll with the Attestation CA 
at `https://attca:port` first. If an AK is already available, the existing certificate and private 
key will be used. The AK private key to be used will be derived from the EK public key. 

The private key for the certificate issued by the ACME server will be attested by the AK and
persisted in a `step context` aware directory, so that it can be reloaded when required. 

The ACME provisioner must be configured with the trusted root of the Attestation CA, so 
that the AK certificate can be verified.

Two additional flags are present to help connecting to the Attestation CA:
- `--attestation-ca-root` can be used to provide the path to a file with trusted root CAs to 
use when a TLS connection is made to the Attestation CA.
- `--attestation-ca-insecure` can be used to disable TLS validation when connecting
to the Attestation CA.


hslatman:
@maraino: There are still things to be improved w.r.t. using the `CreateAttestor` interface, but I think what exists today can be fit to work like that while being backwards compatible. It just needs a bit more work from my side. I would propose to do that when we gather some feedback from the field on how this is currently used.

utils/cautils/tpm.go
```diff
@@ -0,0 +1,651 @@
...
+	akChainBytes := make([][]byte, len(akChain))
+	for i, cert := range akChain {
+		akChainBytes[i] = cert.Raw
+	}
+
+	obj := &attestationObject{
+		Format: "tpm",
+		AttStatement: map[string]interface{}{
+			"ver":      "2.0",
+			"alg":      int64(-257), // AlgRS256 (COSE identifier); depends on type of the private key
```
maraino, line 295: We're also adding support for ECDSA keys, we should fix this.

utils/cautils/tpm.go
maraino, line 1: Although this does not require CGO support, having some of this logic out of the CLI would be nice. I think right now is ok to have it here, but we should think about other options.

Right now, with a YubiKey, the attestation statement is on `step kms attest`, but in that case is because it requires CGO. We will need to integrate this there too.

Other options are to create an ACMEAttester interface or perhaps just an extension to the Attester interface in the crypto KMS package. So we can use it in `step-kms-plugin` as well as here.

utils/cautils/tpm.go
```diff
@@ -0,0 +1,651 @@
...
+	CreateSignature         []byte `json:"createSignature"`
+}
+
+type attestationRequest struct {
+	TPMInfo      tpmInfo               `json:"tpmInfo"`
+	EKPub        []byte                `json:"ek"`
+	EKCerts      [][]byte              `json:"ekCerts"`
+	AKCert       []byte                `json:"akCert"`
+	AttestParams attestationParameters `json:"params"`
+}
```
maraino, line 531: Some of these might be empty, add omitempty to those.

command/ca/certificate.go
```diff
@@ -153,6 +155,24 @@ $ step ca certificate foo.internal foo.crt foo.key \
...
 multiple SANs. The '--san' flag and the '--token' flag are mutually exclusive.`,
 			},
+			cli.StringFlag{
+				Name:  "attestation-ca-url",
+				Usage: "The base url of the Attestation CA to use",
+			},
+			cli.StringFlag{
+				Name:  "attestation-ca-root",
+				Usage: "The path to the PEM <file> with trusted roots when connecting to the Attestation CA",
+			},
```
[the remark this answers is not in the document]
maraino: Or we can bootstrap them.

maraino (changes requested):
Added a few comments.

utils/cautils/tpm.go
```diff
@@ -0,0 +1,651 @@
...
+	if err != nil {
+		return fmt.Errorf("failed retrieving TPM info: %w", err)
+	}
+
+	ui.Printf("\nTPM INFO:")
+	ui.Printf("\nVersion: %s", info.Version)
+	ui.Printf("\nInterface: %s", info.Interface)
+	ui.Printf("\nManufacturer: %s", info.Manufacturer)
+	ui.Printf("\nVendor info: %s", info.VendorInfo)
+	ui.Printf("\nFirmware version: %s", info.FirmwareVersion)
```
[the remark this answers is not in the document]
hslatman: I've changed the order of the output, so that the TPM info comes first. The current output is kind of in line with the ACME status updates:

```console
✔ Provisioner: attestation (ACME)
TPM INFO:
Version: TPM 2.0
Interface: kernel-managed
Manufacturer: Infineon (IFX, 49465800, 1229346816)
Vendor info: SLB9670
Firmware version: 7.85
Using Device Attestation challenge to validate "localhost" . done!
Waiting for Order to be 'ready' for finalization .. done!
Finalizing Order .. done!
✔ Would you like to overwrite l.crt [y/n]: y
✔ Certificate: l.crt
✔ Private Key: tpmkms:name=newkey2
```

I think it's useful to have this output at this stage for support/debugging. I've been thinking a bit about a `--verbose` option for more logging like this. But I think it's OK to add that later (and to more commands).

utils/cautils/tpm.go
hslatman: Yes, I agree that it would be nice to have it all behind a single interface. The thing that made it less simple, at least for me, is the online Attestation CA involved: there are more parameters that the `step kms attest` with a TPM has to know about. And it probably also needs to keep track of some state, independent of the CLI. The `TPMStore` seems like a good candidate for that, but it somewhat muddies its responsibility, so I'm not entirely sure about it yet.

utils/cautils/tpm.go
hslatman: Good catch 🙂

maraino (approved):

PR #713, opened 2022-08-08
revoke <serial> must be base 10 or indicate appropriate base

### LINKED ISSUE #698: [Bug]: In `step ca revoke` convert serial number to big int for consistency.
### Steps to Reproduce

`step ca revoke 0x000111` will revoke the string `0x000111` rather than converting to a Big Int for consistency.

### Your Environment

* OS - any
* `step` CLI Version - latest

### Expected Behavior

Convert the input to a consistent value as key to the database.

### Actual Behavior

The value is used directly as a key into the database.

### Additional Context

_No response_

### Contributing

dopey:
Fixes #698

### DIFF
--- a/command/ca/revoke.go
+++ b/command/ca/revoke.go
@@ -4,6 +4,8 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
+	"math/big"
 	"net/http"
 	"os"
 	"strconv"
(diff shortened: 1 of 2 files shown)


command/ca/revoke.go
```diff
@@ -260,6 +262,13 @@ func revokeCertificateAction(ctx *cli.Context) error {
 		if err := errs.NumberOfArguments(ctx, 1); err != nil {
 			return err
 		}
+
+		_serial := new(big.Int)
+		_serial, ok := _serial.SetString(serial, 0)
+		if !ok {
+			return fmt.Errorf("'%s' is not a valid serial number - use a base 10 representation or add a prefix indicating the base", serial)
+		}
+		serial = _serial.String()
```
maraino, line 271:
```suggestion
		sn, ok := new(big.Int).SetString(serial, 0)
		if !ok {
			return fmt.Errorf("'%s' is not a valid serial number - use a base 10 representation or add a prefix indicating the base", serial)
		}
		serial = sn.String()
```

maraino (approved):
Looks ok, but variables like `_serial` doesn't look very good in Go code. See my suggestion.

PR #727, opened 2022-08-19
Fix or ignore gosec issues

dopey:
* fix a few other linting issues


.golangci.yml
```diff
@@ -46,28 +82,246 @@ linters-settings:
...
+    - structcheck # Finds unused struct fields
+    - typecheck # Like the front-end of a Go compiler, parses and type-checks Go code
+    - unused # Checks Go code for unused constants, variables, functions and types
+    - varcheck # Finds unused global variables and constants
+    ## disabled by default
+    - gocritic # Provides diagnostics that check for bugs, performance and style issues.
+    - gofmt # [replaced by goimports] Gofmt checks whether code was gofmt-ed. By default this tool runs with -s option to check for code simplification
+    - goimports # In addition to fixing imports, goimports also formats your code in the same style as gofmt.
+    - misspell # Finds commonly misspelled English words in comments
+    - revive # Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint.
```
maraino, line 211: gosec is not enabled.

cmd/step/main.go
```diff
@@ -41,9 +41,6 @@ import (
 	_ "github.com/smallstep/certificates/cas/cloudcas"
 	_ "github.com/smallstep/certificates/cas/softcas"
 	_ "github.com/smallstep/certificates/cas/stepcas"
-
-	// Profiling and debugging
-	_ "net/http/pprof"
```
maraino, line 46: You should remove the profile server below, the code is:
```go
// Start the golang debug logger if environment variable is set.
// See https://golang.org/pkg/net/http/pprof/
debugProfAddr := os.Getenv("STEP_PROF_ADDR")
if debugProfAddr != "" {
	go func() {
		log.Println(http.ListenAndServe(debugProfAddr, nil))
	}()
}
```

command/fileserver/fileserver.go
```diff
@@ -121,6 +122,7 @@ func fileServerAction(ctx *cli.Context) error {
 		tlsConfig = &tls.Config{
 			ClientCAs:  pool,
 			ClientAuth: tls.RequireAndVerifyClientCert,
+			MinVersion: tls.VersionTLS13,
```
maraino, line 125: I'm ok with requiring 1.3, but as this is mainly used for testing 1.2 might be more useful.

maraino (changes requested):
Just a few minor things.

crypto/pemutil/pkcs8.go
```diff
@@ -292,7 +292,7 @@ func MarshalPKCS8PrivateKey(key interface{}) ([]byte, error) {
 // key derived using PBKDF2 over the given password.
 func DecryptPEMBlock(block *pem.Block, password []byte) ([]byte, error) {
 	if block.Headers["Proc-Type"] == "4,ENCRYPTED" {
-		// nolint
+		//nolint:staticcheck // only used for encryption, not decryption.
```
maraino, line 295: Well, just the opposite, just decryption.

crypto/pemutil/pem.go
```diff
@@ -418,7 +418,7 @@ func Serialize(in interface{}, opts ...Options) (*pem.Block, error) {
 			}
 		} else {
 			var err error
-			//nolint
+			//nolint:staticcheck // not being used for cryptographic purposes.
```
maraino, line 421: Well, yes, and as we haven't moved to PKCS8 by default, this is the only option.

maraino: We can convert keys to use PKCS8, but we're not storing them by default.

crypto/pemutil/pkcs8.go
dopey: oops

crypto/pemutil/pem.go
dopey: Added another comment.

make/common.mk
```diff
@@ -75,7 +75,7 @@ integration: bin/$(BINNAME)
 #########################################
 
 fmt:
-	$Q gofmt -l -w $(SRC)
+	$Q goimports -local github.com/golangci/golangci-lint -l -w $(SRC)
```
maraino, line 78: Should we add `goimports` to `make bootstrap` or just assume it's there because it's installed by most of the Go editor modes?

usage/html.go
```diff
@@ -235,7 +234,6 @@ func (h *htmlHelpHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 			continue
 		}
 		cmd.HelpName = fmt.Sprintf("%s %s", ctx.App.HelpName, strings.Join(args, " "))
-		parent.HelpName = fmt.Sprintf("%s %s", ctx.App.HelpName, strings.Join(args[:last], " "))
```
maraino, line 238: It seems ok, but I've realized we are using the methods in `cli-utils` instead of these.

maraino (changes requested):
There's a `r.Body.Close()` that is not in the right place and a few notes

make/common.mk
dopey: Let's just assume it's there. If we hear about issues we'll add to `bootstrap`.

command/crl/crl_extensions.go
```diff
@@ -92,7 +92,7 @@ func (e *Extension) MarshalJSON() ([]byte, error) {
 	return json.Marshal(e.json)
 }
 
-func (e *Extension) AddDetail(format string, args ...interface{}) {
+func (e *Extension) AddDetailf(format string, args ...interface{}) {
```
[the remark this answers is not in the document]
dopey: Agreed, I'll turn it off.

maraino (approved):
lgtm

PR #733, opened 2022-08-24
--x5c-root → --x5c-roots

tashian:
The flag `--x5c-root` in `provisioner add` and `provisioner update` should be plural, because you can configure multiple root CAs to be trusted.

- [x] This would benefit from backward compatibility. I could add `Aliases` to the `StringFlag`, but I think the singular version should be hidden from help text and I wasn't sure how to do that.

### DIFF
--- a/command/ca/provisioner/add.go
+++ b/command/ca/provisioner/add.go
@@ -53,7 +53,7 @@ OIDC
 
 X5C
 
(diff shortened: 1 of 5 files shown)


maraino:
Another possibility is to change this to a `cli.StringSliceFlag`

tashian:
@dopey @maraino I added an alias here, so this is now backward compat.

dopey (approved):
lgtm

PR #734, opened 2022-08-24
Add category to command reference Markdown front matter

tashian:
I want to distinguish the reference docs from the rest of the documentation, and adding a `category` is a better way to do that than using the `layout` as a proxy. See smallstep/docs#131 for where this will be applied.


dopey (approved):
lgtm

PR #736, opened 2022-08-24
Update renewal link in docs

tashian:


maraino (changes requested):
That URL doesn't have any content

tashian:
It will once smallstep/docs#132 is merged

maraino (approved):
It doesn't work yet, probably because docs is not deployed, but this PR looks ok once docs is deployed.

PR #741, opened 2022-08-30
Implement device-attestations for yubikeys

maraino:
### Description

Currently the yubikey KMS implements the attestation interface, this experimental commit uses that interface to request a certificate using ACME with the device-attest-01 challenge.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -399,8 +399,8 @@ flag exists so it can be configured in $STEPPATH/config/defaults.json.`,
 		Usage: "The <uri> to configure a Cloud KMS or an HSM.",
 	}
 
-	Attest = cli.StringFlag{
-		Name:  "attest",
+	AttestationURI = cli.StringFlag{
+		Name:  "attestation-uri",
 		Usage: "The KMS <uri> used for attestation.",
 	}
 )
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -398,6 +398,11 @@ flag exists so it can be configured in $STEPPATH/config/defaults.json.`,
 		Name:  "kms",
 		Usage: "The <uri> to configure a Cloud KMS or an HSM.",
 	}
+
+	Attest = cli.StringFlag{
+		Name:  "attest",
+		Usage: "The KMS <uri> used for attestation.",
+	}
 )
 
 // ParseTimeOrDuration is a helper that returns the time or the current time
--- a/internal/cryptoutil/cryptoutil.go
+++ b/internal/cryptoutil/cryptoutil.go
@@ -17,7 +17,7 @@ import (
 )
 
 type Attestor interface {
-	Public() crypto.PublicKey
+	crypto.Signer
 	Attest() ([]byte, error)
 }
 
--- a/internal/cryptoutil/cryptoutil.go
+++ b/internal/cryptoutil/cryptoutil.go
@@ -16,7 +16,7 @@ import (
 	"go.step.sm/crypto/pemutil"
 )
 
-// Attestor is the interface implemented by step-kms-plugin using the key, sign
+// Attestor is the interface implemented by step-kms-plugin using the key, sign,
 // and attest commands.
 type Attestor interface {
 	crypto.Signer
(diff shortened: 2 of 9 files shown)


internal/cryptoutil/cryptoutil.go
```diff
@@ -33,6 +38,12 @@ func CreateSigner(kms, name string, opts ...pemutil.Options) (crypto.Signer, err
 	return newKMSSigner(kms, name)
 }
 
+// CreateAttestor creates an attestor that will use `step-kms-plugin` with the
+// given kms and uri.
```
hslatman, line 42: It says `given kms and uri`, but it takes a kms and name.

utils/cautils/acme_flow.go
```diff
@@ -29,11 +29,14 @@ func ACMECreateCertFlow(ctx *cli.Context, provisionerName string) error {
...
-	_, err = pemutil.Serialize(af.priv, pemutil.ToFile(keyFile, 0600))
-	if err != nil {
-		return errors.WithStack(err)
+	// We won't have a private key with attestation certificates
+	if af.priv != nil {
+		_, err = pemutil.Serialize(af.priv, pemutil.ToFile(keyFile, 0600))
+		if err != nil {
+			return errors.WithStack(err)
+		}
+		ui.PrintSelected("Private Key", keyFile)
```
hslatman, line 38: It may be an option to indicate that the private key is created and stored in the KMS for the attestation flow?

hslatman (commented):
Changes and refactoring look good. I'll have another look at it when integrating with my work.

utils/cautils/acmeutils.go
```diff
@@ -193,6 +203,7 @@ func serveAndValidateHTTPChallenge(ctx *cli.Context, ac *ca.ACMEClient, ch *acme
 }
 
 func authorizeOrder(ctx *cli.Context, ac *ca.ACMEClient, o *acme.Order) error {
+	attest := (ctx.String("attest") != "")
```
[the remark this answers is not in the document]
maraino: What I don't like about that is that they are not equivalent. You get `IsSet` set to true if you do `--attest ""`

internal/cryptoutil/cryptoutil.go
maraino: Fixed with 6ad024e

utils/cautils/acme_flow.go
maraino: Fixed with 9a45100

flags/flags.go
```diff
@@ -398,6 +398,11 @@ flag exists so it can be configured in $STEPPATH/config/defaults.json.`,
 		Name:  "kms",
 		Usage: "The <uri> to configure a Cloud KMS or an HSM.",
 	}
+
```
dopey, line 401: Is `attest` descriptive enough? Should it be something like `attestation-uri`?

internal/cryptoutil/cryptoutil.go
```diff
@@ -16,6 +16,13 @@ import (
 	"go.step.sm/crypto/pemutil"
 )
 
+// Attestor is the interface implemented by step-kms-plugin using the key, sign
```
dopey, line 19:
```suggestion
// Attestor is the interface implemented by step-kms-plugin using the key, sign,
```

utils/cautils/acmeutils.go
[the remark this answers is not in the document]
dopey: Can we call this variable `isAttest` or something to indicate that it's a boolean? Normally I wouldn't mind, but the flag input is a string and we're not consistent about how we use it.

dopey (approved):
lgtm

utils/cautils/acmeutils.go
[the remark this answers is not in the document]
maraino: Fixed with 377194e

flags/flags.go
maraino: Fixed with 4fc5893

PR #752, opened 2022-09-24
[action] Update workflows

dopey:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+  workflow_call:
+    secrets:
+      GITLEAKS_LICENSE_KEY:
+        required: true
+
+jobs:
+  ci:
+    uses: smallstep/workflows/.github/workflows/goCI.yml@main
+    with:
+      os-dependencies: "libpcsclite-dev"
```
maraino, line 19: This is not a dependency of `step`

.github/workflows/code-scan-cron.yml
```diff
@@ -0,0 +1,9 @@
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
maraino, line 3: You're always using the same schedule, would we get any benefit if we use a different one per project? Perhaps a few minutes appart?

maraino (changes requested):
The os-dependency is not necessary for this project.

.github/workflows/code-scan-cron.yml
dopey: What would the benefit be? It's not running against our servers. Also, I changed every repo that isn't cli, certificates, crypto to only do once a week at midnight on sunday.

maraino (approved):

Issue #577, opened 2021-11-10
Add --enable-admin and --acme flags for the `step ca init` command.


dopey:
This will allow users to set up API mgmt for provisioners, and unblock users in environments where the `ca.json` is not easily accessible.

The --enable-admin flag would create the first provisioner and admin (this code already exists, just behind a boolean). 
The --acme flag would create an ACME provisioner.

Related: https://github.com/smallstep/certificates/issues/737

JonathonRichardson:
The ACME user flag would be particularly useful.  In addition to adding it to the CLI, it should probably be exposed as an option for the docker image via the [entrypoint.sh](https://github.com/smallstep/certificates/blob/c82296b7cdecd1bf4121c9dbb89ddb82d21ad8ef/docker/entrypoint.sh) script, so that we could set an environment variable (e.g. `DOCKER_STEPCA_INIT_ACME_USER=true`) in the container config to have a fully compliant ACME server on startup.

As a workaround for now, I use a modified version of the entrypoint to get the CA to be ready to go for ACME requests on startup/creation by updating the `step_ca_init ` function to call `step ca provisioner add acme --type ACME` before finishing up, but it'd be nice to get back to just launching the image as is from Docker Hub.

```
# Initialize a CA if not already initialized
function step_ca_init () {
    local -a setup_args=(
        --name "${DOCKER_STEPCA_INIT_NAME}"
		--dns "${DOCKER_STEPCA_INIT_DNS_NAMES}"
		--provisioner "${DOCKER_STEPCA_INIT_PROVISIONER_NAME:-admin}"
		--password-file "${STEPPATH}/password"
        --address ":9000"
    )
    if [ -n "${DOCKER_STEPCA_INIT_PASSWORD}" ]; then
        echo "${DOCKER_STEPCA_INIT_PASSWORD}" > "${STEPPATH}/password"
    else
        generate_password > "${STEPPATH}/password"
    fi
    if [ -n "${DOCKER_STEPCA_INIT_SSH}" ]; then
        setup_args=("${setup_args[@]}" --ssh)
    fi
    step ca init "${setup_args[@]}"

    # https://smallstep.com/docs/step-ca/acme-basics#configure-step-ca-for-acme
    step ca provisioner add acme --type ACME
    
    mv $STEPPATH/password $PWDPATH
}
```

PR #758, opened 2022-09-27
Add `--remote-management` and `--acme` flags to `ca init` command

hslatman:
#### Name of feature:

Add `--remote-management` and `--acme` flags to `ca init` command

Configuration being a bit of a hassle in (a.o.) container environments.

This allows users to more easily initialize and configure a `step-ca` instance in environments where the `ca.json` file is not easily accessible or editable. One example is a Docker container running our default `smallstep/step-ca` image.

Just the CLI reference documentation.

The `--acme` flag is (currently) only supported in standalone deployments. The `--remote-management` flag is supported in standalone and linked deployments.

When these flags are supported, we can update the Docker image `entrypoint.sh` to make use of these flags.

The `--acme` flag is not supported in hosted and linked deployments. The `--remote-management` flag is not supported in hosted deployment.

#577 

We may want to look into adding some additional info output after the CA is created stating the name of the first admin. It's not strictly part of this issue, as nothing has changed in this behavior when enabling the Admin API, but I noticed we didn't do this and users may not know about the default `step` admin user.


maraino:
Hi @hslatman, I haven't looked at this, but it's quite possible that you need to do changes [here](https://github.com/smallstep/certificates/blob/master/pki/helm.go) to enable the remote administration.

hslatman:
Makes sense, @maraino. So far I've tried to not sail too close to Helm and its comrades, but I suppose it makes sense to support it there too. I'll see if I can get it done; worst case I can make the flag incompatible.

maraino:
@hslatman See this [tldr](https://github.com/smallstep/helm-charts/tree/master/step-certificates#tldr)
```
helm repo add smallstep https://smallstep.github.io/helm-charts/
step ca init --helm > values.yaml
echo "password" | base64 > password.txt
helm install -f values.yaml \
     --set inject.secrets.ca_password=$(cat password.txt) \
     --set inject.secrets.provisioner_password=$(cat password.txt) \
     --set service.targetPort=9000 \
     step-certificates smallstep/step-certificates
```

hslatman:
An attempt at supporting Helm is here: https://github.com/smallstep/certificates/pull/1075.

If I understand the flow correctly, then this turns out to be harder to support than I thought.

go.mod
```diff
@@ -183,18 +183,19 @@ require (
...
+	google.golang.org/grpc v1.50.0 // indirect
 	gopkg.in/cheggaaa/pb.v1 v1.0.28 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.0.0 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	howett.net/plist v0.0.0-20181124034731-591f970eefbb // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+replace github.com/smallstep/certificates => ./../certificates6
```
dopey, line 201: Think we need to remove this, yea?

dopey (changes requested):
Just need to remove the replace in the go.mod I think, and then this is good to go.

go.mod
maraino: I've been using `go.work` recently to avoid this. In the same directory have a file like this:
```go
go 1.18

use .
use ../certificates
// use ../certinfo
// use ../linkedca
// use ../cli-utils
// use ../crypto


```

command/ca/init.go
```diff
@@ -411,6 +437,13 @@ func initAction(ctx *cli.Context) (err error) {
 			ui.Println()
 			return nil
 		}
+		// When initializing a linked CA, providing the --acme flag doesn't currently
+		// result in the default ACME provisioner being added. We may want to support this
+		// for ease of use, but this seems to require a bit of refactoring when generating
+		// the full CA configuration with DB initialization.
+		if deploymentType != pki.StandaloneDeployment && addDefaultACMEProvisioner {
+			return fmt.Errorf("adding a default ACME provisioner by providing the --acme flag is not supported with deployment type %q", deploymentType.String())
+		}
```
maraino, line 446: Should we say something about using` step ca provisioner add acme --type ACME`?

hslatman: Might be good. This error will only be returned for linked deployments currently, and considering that we want to make it easy to do linked ACME, it's probably good to make this clear to the user, indeed.

hslatman: Fixed in d28911c85ea9cd4d4905bf94abeff523de6f9aae

command/ca/init.go
```diff
@@ -411,6 +437,13 @@ func initAction(ctx *cli.Context) (err error) {
 			ui.Println()
 			return nil
 		}
+		// When initializing a linked CA, providing the --acme flag doesn't currently
+		// result in the default ACME provisioner being added. We may want to support this
+		// for ease of use, but this seems to require a bit of refactoring when generating
+		// the full CA configuration with DB initialization.
+		if deploymentType != pki.StandaloneDeployment && addDefaultACMEProvisioner {
+			return fmt.Errorf("adding a default ACME provisioner by providing the --acme flag is not supported with deployment type %q. Please use `step ca provisioner add acme --type ACME` after initializing your CA", deploymentType.String())
```
maraino, line 445: A `\n` would be nice here:
```suggestion
			return fmt.Errorf("adding a default ACME provisioner by providing the --acme flag is not supported with deployment type %q.\nPlease use `step ca provisioner add acme --type ACME` after initializing your CA", deploymentType.String())
```

maraino (dismissed):
I suggested a \n in an error message, but other than that 👍

maraino (approved):

hslatman:
@maraino had a conflict in `go.mod`/`go.sum`. This is probably going to cause more friction down the road 😅

maraino (approved):

PR #761, opened 2022-09-30
Use go.step.sm/crypto

### LINKED ISSUE #419: Migrate to go.step.sm packages
### Description

To remove dependencies of the cross dependency of the `step` and `step-ca` we created `go.step.sm/crypto` and `go.step.sm/cli-utils`. `step-ca` already uses those ones, but we should to do the same in `step`, so we don't have to maintain the same functionality twice.

maraino:
### Description

This PR attempts to replace as much as possible the use of `github.com/smallstep/cli/crypto` with `go.step.sm/crypto`.

Fixes #419


dopey (approved):
lgtm

Issue #769, opened 2022-10-06
[Bug]: CI does not work well on forks


jsoref:
### Steps to Reproduce

1. Set up an organization where workflows do not have write by default
2. Fork this repo to such a repo
3. Create a PR within that repo

### Your Environment

* organization (check-spelling) where workflows do not have write by default
* https://github.com/check-spelling/smallstep-cli/

https://github.com/check-spelling/smallstep-cli/actions/runs/3194799349
https://github.com/check-spelling/smallstep-cli/actions/runs/3194799311

### Expected Behavior

Workflows should work in forks

### Actual Behavior

https://github.com/check-spelling/smallstep-cli/actions/runs/3194799349
<img width="1007" alt="image" src="https://user-images.githubusercontent.com/2119212/194245328-619c59cd-d27c-4715-8c47-b12326d0c3f7.png">

https://github.com/check-spelling/smallstep-cli/actions/runs/3194799311
<img width="452" alt="image" src="https://user-images.githubusercontent.com/2119212/194245226-8c0eb777-2646-4981-9c80-a3d858b94b77.png">

### Additional Context

https://github.com/jsoref/smallstep-cli/commit/0f6955086859255b904a7e53a2be7e69c3ca8290 may fix some of the problems.

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #776, opened 2022-10-11
Improve Ci in forks

jsoref:
#### Name of feature:
CI in forks w/ restrictive default permissions

#769

💔Thank you!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,6 +8,9 @@ on:
 
 jobs:
   ci:
+    permissions:
+      actions: read
+      security-events: write
     uses: smallstep/cli/.github/workflows/ci.yml@master
     secrets: inherit
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,7 @@ jobs:
   ci:
     permissions:
       actions: read
+      contents: read
       security-events: write
     uses: smallstep/cli/.github/workflows/ci.yml@master
     secrets: inherit
--- a/.github/workflows/triage.yml
+++ b/.github/workflows/triage.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request_target'
     steps:
-    - uses: actions/labeler@v3.0.2
+    - uses: actions/labeler@v4.0.2
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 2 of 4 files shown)


jsoref:
Testing was performed for the triage workflow here: https://github.com/check-spelling/smallstep-cli/actions/workflows/triage.yml

You can see that it didn't fail for runs 2/3/4. For the final run, I added a dummy secret. That resulted in it *trying* to perform the action, but since the secret wasn't proper, I got a failure, which is sufficient proof that if I had a valid secret it would have worked just as it would work here in a repository which has a secret.

The actions/contents/security-events items appear correct, but the CI fails because I don't have codecov configured which is not a permission (but presumably is a secret). I suppose I could look into that, but if I can leave it as an exercise (you can effectively copy the style from the triage project code...), that'd be great.

.github/workflows/triage.yml
```diff
@@ -4,26 +4,39 @@ on:
...
+        shell: bash
+        if: env.HAS_TRIAGE_PAT == 'true'
+        run: |
+          echo '::set-output name=triage::1'
+
       - uses: actions/add-to-project@v0.3.0
+        if: steps.check.outputs.triage
         with:
           project-url: https://github.com/orgs/smallstep/projects/94
           github-token: ${{ secrets.TRIAGE_PAT }}
```
dopey, line 42:
```suggestion
  add-to-project:
    name: Add to Triage Project
    runs-on: ubuntu-latest
    env: 
      HAS_TRIAGE_PAT: ${{ secrets.TRIAGE_PAT != '' }}
    if: env.HAS_TRIAGE_PAT == 'true'
    steps:
      - name 
         uses: actions/add-to-project@v0.3.0
         with:
           project-url: https://github.com/orgs/smallstep/projects/94
           github-token: ${{ secrets.TRIAGE_PAT }}
```

dopey:
Codecov isn't a secret. Public repos don't require a codecov token, and shouldn't fail if you provide an empty one. We've got a bunch of public repos that are not providing a CODECOV token, but all of them use that same flow.

dopey (dismissed):

.github/workflows/release.yml
```diff
@@ -8,6 +8,9 @@ on:
 
 jobs:
   ci:
+    permissions:
+      actions: read
```
dopey, line 12:
```suggestion
      actions: read
      contents: read
```

dopey (approved):

PR #779, opened 2022-10-13
[action] Add bot GH signing key for auto docs update

dopey:


.github/workflows/release.yml
```diff
@@ -180,13 +180,28 @@ jobs:
           repository: smallstep/docs
           token: ${{ secrets.DOCS_PAT }}
           path: './docs'
+      - name: Setup bot SSH signing key
+        uses: webfactory/ssh-agent@v0.5.3
+        env:
+          has_SSH_PRIVATE_KEY: ${{ secrets.STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY != '' }}
+        if: ${{ env.has_SSH_PRIVATE_KEY != '' }}
```
maraino, line 187: Should it be 
```suggestion
        env:
          HAS_SSH_PRIVATE_KEY: ${{ secrets.STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY != '' }}
        if: ${{ env.HAS_SSH_PRIVATE_KEY != '' }}
```

maraino: Or can you do the if with the secret?
```suggestion
        if: ${{ secrets.STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY != '' }}
```

.github/workflows/release.yml
```diff
@@ -180,13 +180,28 @@ jobs:
...
+        with:
+          ssh-private-key: |
+            ${{ secrets.STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY }}
+            STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY
```
maraino, line 191: Do we need this extra line (`STEP_TRAVIS_CI_GH_PRIVATE_SIGNING_KEY`)?

maraino: Should it be a second SSH key from a secret?

maraino (commented):
I've added a suggestion, and there's one line unclear about what it does.

.github/workflows/release.yml
dopey: Unfortunately you can't use secrets in conditionals. But the code is wrong. I should be comparing to `true`.

maraino (approved):

Issue #767, opened 2022-10-06
Allow x5c commands to acquire keys/certificates from platform keystores (CAPI/Keychain/etc)


unreality:
## X5C Provisioner Improvements

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Windows and MacOS have secure key and certificate storage (TPM and Secure Enclave respectively). Linux can also access the platform TPM if available. It would be useful to allow the x5c provisioner process to sign its token using one of the keys in the platform store instead of only on disk.

On Windows 10, the Crypto API (aka CAPI) and Next-Gen Crypto API (nCrypt) could be used which would allow users to sign the request with smart cards/yubikeys as well. CAPI can be called from golang reasonably easily with `syscall`, but I am unsure of how to call the required Secure Enclave methods on MacOS

## Why is this needed?

This would take advantage of the platform specific secure key storage to avoid having to load and store keys on disk.

maraino:
To support this, we will need to create a KMS or a subset of this for TPM (@hslatman) and Secure Enclave (@maraino) in `go.step.sm/crypto`. Then use `step-kms-plugin` in more places of `step`.

Right now it is possible to create tokens using `step-kms-plugin` with yubikeys, pkcs#11, cloud kms, ... but it's not straightforward. 

We could simplify the flags, but with existing KMSs something like this should be supported:

```shell
step ca certificate \
  --kms 'yubikey:?pin-value=123456' \
  --x5c-cert 'yubikey:slot-id=82' \
  --x5c-key 'yubikey:slot-id=82' \
  --provisioner X5C \
  localhost localhost.crt localhost.key
```

unreality:
@maraino would you accept a pull that added a new kms of 'capi' that used the windows CAPI/nCrypt to find certs/keys and raise notimplemented errors on other platforms?

hslatman:
Adding a `capi` KMS sounds reasonable to me. If you're going to use `syscall`, it can probably live in `step-kms-plugin`. I feel it's somewhat of a bigger lift, so it probably needs @maraino's approval too and maybe some internal discussion. 

I'm working on some TPM things. The library supports Windows too. That means we'll have multiple "interfaces" to the same underlying cryptographic hardware, I guess? I don't think that's a big problem, though.

unreality:
[...]

PR #783, opened 2022-10-24
WIP: Allow x5c to use KMS certs and keys for token generation

unreality:
#### Name of feature:

This is some rough initial work to let users specify KMS certificates and Keys when using an X5C provisioner, for example:
```
step ca certificate \
  --kms "capi:provider=Microsoft Smart Card Key Storage Provider" \
  --x5c-key "capi:key=MYTESTKEY" \
  --x5c-cert "capi:store=My;sha1=b4b90bf984064df18b5aa9f3a1e47dee5b86582d"\
  --x5c-chain "capi:store=CA;sha1=ffb71dade1deb93d48c32fcbb402b8e9873d822c" \
  --provisioner X5C
  test_subject test.crt test.key
```

```
step ca certificate \
  --kms 'yubikey:?pin-value=123456' \
  --x5c-cert 'yubikey:slot-id=82' \
  --x5c-key 'yubikey:slot-id=82' \
  --provisioner X5C \
  localhost localhost.crt localhost.key
```

Only `step ca certificate` and `step ssh certificate` have the added `kms` and `x5c-chain` flags at the moment.

This adds another flag to the CLI; `x5c-chain`. Since the KMS `LoadCertificate` interface only returns a single certificate, a method to specify the intermediate chain certificates was required. You can specify the flag multiple times to create the complete chain.

Please let me know what changes need to occur to push this forward.

Closes https://github.com/smallstep/cli/issues/767
Also requires [some changes to smallstep/crypto](https://github.com/smallstep/crypto/pull/105)

💔Thank you!


maraino (changes requested):
Looks good, just a couple of minor changes.

utils/cautils/token_generator.go
```diff
@@ -161,32 +162,68 @@ func generateK8sSAToken(ctx *cli.Context, p *provisioner.K8sSA) (string, error)
...
+	} else {
+		// Get private key from given key file
+		var opts []jose.Option
+		if passOpt := getProvisionerPasswordOption(ctx); passOpt != nil {
+			opts = append(opts, passOpt)
+		}
+		jwk, err = jose.ReadKey(x5cKeyFile, opts...)
+		if err != nil {
+			return "", err
+		}
```
maraino, line 202: You can do this with `cryptoutil.CreateSigner` as it is, even if that `jose` was used here, the key should be a PEM file.

`cryptoutil.CreateSigner` will use `pemutil.Read()` when `kms == ""` you will need to the `pemutil.Options` using a copy of the `getProvisionerPasswordOption()` function, but returning `pemutil.Options`

command/ca/certificate.go
```diff
@@ -172,8 +172,10 @@ multiple SANs. The '--san' flag and the '--token' flag are mutually exclusive.`,
 			flags.Offline,
 			flags.PasswordFile,
 			consoleFlag,
+			flags.KMSUri,
 			flags.X5cCert,
 			flags.X5cKey,
+			flags.X5cChain,
```
maraino, line 178: Can you add this to `command/ca/sign.go` too

maraino (changes requested):

utils/cautils/token_generator.go
maraino: Yes, but `keyutil.VerifyPair()` will verify the pair comparing the public keys, so at the end is the same.

utils/cautils/token_generator.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/smallstep/cli/internal/cryptoutil"
```
maraino, line 9: Format this file with `goimports`, this import should be ordered in the next block of imports.

token/token.go
```diff
@@ -124,9 +124,14 @@ func DefaultClaims() *Claims {
 
 // GenerateKeyID returns the SHA256 of a public key.
 func GenerateKeyID(priv interface{}) (string, error) {
+	if signer, ok := priv.(jose.OpaqueSigner); ok {
+		return jose.Thumbprint(signer.Public())
+	}
```
maraino, line 129: Ok, but I'm considering adding this to keyutil too.

utils/cautils/token_generator.go
```diff
@@ -161,32 +162,60 @@ func generateK8sSAToken(ctx *cli.Context, p *provisioner.K8sSA) (string, error)
...
 	}
-	jwk, err := jose.ReadKey(x5cKeyFile, opts...)
+
+	var kmsSigner crypto.Signer
+	kmsSigner, err = cryptoutil.CreateSigner(ctx.String("kms"), x5cKeyFile, opts...)
 	if err != nil {
 		return "", err
 	}
 
+	joseSigner := jose.NewOpaqueSigner(&kmsSigner)
```
maraino, line 189: This should be:
```suggestion
	joseSigner := jose.NewOpaqueSigner(kmsSigner)
```

It's not a good idea to use a pointer to an interface. I fixed that in a PR.

maraino (changes requested):

maraino (approved):
Thanks, @unreality

I'm also creating a 0.4.0 release of step-kms-plugin, it should include your KMS. Let me know if you find any problems.

PR #787, opened 2022-10-27
fixup: RELEASE_DATE for release tags

pheiduck:
#### Name of feature:
Fixup: get date for releases 

date is not parsed over next tasks.

https://github.com/nanzm/get-time-action#readme
release.yml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -73,10 +73,16 @@ jobs:
       - name: Write cosign key to disk
         id: write_key
         run: echo "${{ secrets.COSIGN_KEY }}" > "/tmp/cosign.key"
+      - 
+        name: Get current time
+        id: time
+        uses: nanzm/get-time-action@master
+        with:
+          format: 'YYYY-MM-DD'
       - name: Get Release Date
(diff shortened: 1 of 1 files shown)


dopey:
Hey @pheiduck 👋 , thanks for the PR!

As a general rule in our Github actions workflows we want to pull in fewer outside dependencies, especially third-party dependencies. Each new dependency is a potential security threat. In this situation I'm not sure what we'd be gaining by introducing the dependency - the outcome seems the same, but I might be missing something. What's the rationale?

pheiduck:
Hey @dopey, thanks for the warm welcome! :)
The date is not displayed correctly for prereleases.
Bildschirmfoto 2022-10-27 um 20 45 40

dopey:
Appreciate the call out on the mistake. I'm currently working on our release pipeline and introduced a small bug - hence the broken release date there. I just wasn't accessing the env var correctly. Appreciate you pointing it out, and I have since fixed it in my branch.

Still, I'd prefer not to pull in a 3rd party action to handle this one task. 

Cheers 🍻

PR #795, opened 2022-11-10
Add --admin-password-file flag

### LINKED ISSUE #744: [Bug]: `--password-file` in `step ca provisioner add --type JWK` seems to be used for both the provisioning admin and the newly created provisioner
### Steps to Reproduce

```bash
step ca provisioner add "newprovisioner" --type JWK --create --admin-subject admin@foo.com --admin-provisioner "Current provisioner" --password-file admin/password
```

### Your Environment

Smallstep CLI/0.21.0 (darwin/arm64)
Release Date: 2022-07-07 01:07 UTC

### Expected Behavior

IMHO `--password-file` should *either* control which password will be used for the newly created provisioner, *OR* what is the password of the current provisioner being used.

### Actual Behavior

Documentation reads:

```
      --password-file=file
          The path to the file containing the password to encrypt or decrypt the
          private key
```

It is not clear *from the documentation* if this is about the provisioning admin, or the newly created provisioner...

Turns out (unless I messed-up) that *this is being used for both*.
* clearly here for the newly created provisioner: https://github.com/smallstep/cli/blob/master/command/ca/provisioner/add.go#L420
* and probably here for the provisioning one: https://github.com/smallstep/cli/blob/005920ad4122fa5838c338b567d6584fcc33c1cc/utils/cautils/client.go#L176

Using the same flag for both of these seems quite wrong.

If only, it seems to prevent you from creating a new provisioner non-interactively.

Hope I am making sense and this is not a PEBKAC.

Let me know if you need anything else on this?

### Additional Context

_No response_

### Contributing

maraino:
### Description

On commands like "step ca provisioner add", the "--password-file" flag was used with two different purposes, to encrypt a new JWK key and to decrypt the admin JWK key. This change adds an extra flag --admin-password-file to make this different.

Fixes #744

### DIFF
--- a/command/ca/provisioner/update.go
+++ b/command/ca/provisioner/update.go
@@ -39,7 +39,7 @@ ACME
 **step ca provisioner update** <name> [**--force-cn**] [**--require-eab**]
 [**--challenge**=<challenge>] [**--remove-challenge**=<challenge>]
 [**--attestation-format**=<format>] [**--remove-attestation-format**=<format>]
-[**--attestation-roots**=<file>] [**--admin-cert**=<file>] [**--admin-key**=<file>] 
+[**--attestation-roots**=<file>] [**--admin-cert**=<file>] [**--admin-key**=<file>]
 [**--admin-subject**=<subject>] [**--admin-provisioner**=<name>] [**--admin-password-file**=<file>]
 [**--ca-url**=<uri>] [**--root**=<file>] [**--context**=<name>] [**--ca-config**=<file>]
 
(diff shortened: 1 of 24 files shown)


command/ca/provisioner/update.go
```diff
@@ -30,17 +30,17 @@ func updateCommand() cli.Command {
...
+[**--ca-url**=<uri>] [**--root**=<file>] [**--context**=<name>] [**--ca-config**=<file>]
 
 ACME
 
 **step ca provisioner update** <name> [**--force-cn**] [**--require-eab**]
 [**--challenge**=<challenge>] [**--remove-challenge**=<challenge>]
 [**--attestation-format**=<format>] [**--remove-attestation-format**=<format>]
-[**--attestation-roots**=<file>] [**--admin-cert**=<file>] [**--admin-key**=<file>]
-[**--admin-provisioner**=<name>] [**--admin-subject**=<subject>] [**--password-file**=<file>]
+[**--attestation-roots**=<file>] [**--admin-cert**=<file>] [**--admin-key**=<file>]
```
dopey, line 42:
```suggestion
[**--attestation-roots**=<file>] [**--admin-cert**=<file>] [**--admin-key**=<file>]
```

dopey (approved):

PR #804, opened 2022-11-16
Update renew docs to clarify when mTLS is and is not used for renewal

tashian:
This will close https://github.com/smallstep/docs/issues/155


maraino (changes requested):
We should mention something about RAs. Starting with v0.23.0, when the default step-based RA is used (`stepcas`) we need the `--mtls=false` flag, but this will be automatically set in a future version.

cloudcas RA should work both ways, and vaultcas renewal is not supported.

maraino (dismissed):

maraino (approved):

PR #808, opened 2022-12-01
Only restart services which are actually running

darix:
#### Name of feature:

Only restart services which are actually running

A service might be enabled by default but not currently running. With the old code we would restart it in this case. is-active only touches currently running services.

systemd


tashian (approved):
lgtm thanks @darix !

PR #809, opened 2022-12-03
Check for remote admin api before prompting for creds

dopey:


command/ca/provisioner/provisioner.go
```diff
@@ -101,38 +108,68 @@ type crudClient interface {
...
+		TLSClientConfig: tlsConfig,
+	}
+
+	client := &http.Client{Transport: tr}
+	// TODO: Replace this logic with url.JoinPath once go1.18 is no longer supported.
+	u, err := url.Parse(caURL)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing caURL: %w", err)
+	}
+	u.Path = path.Join(u.Path, "admin/admins")
```
maraino, line 146: It's probably better to use the following:
```suggestion
	u = u.ResolveReference(&url.URL{Path: "/admin/admins"})
```

command/ca/provisioner/provisioner.go
```diff
@@ -101,38 +108,68 @@ type crudClient interface {
...
+		return nil, fmt.Errorf("error generating cert pool: %w", err)
 	}
 
-	_, err := os.Stat(cfgFile)
-	switch {
-	case errors.Is(err, os.ErrNotExist):
-		return cautils.NewAdminClient(cliCtx)
-	case err == nil:
+	tlsConfig := &tls.Config{
+		InsecureSkipVerify: false,
```
maraino, line 131: There's no need to set this, it's false by default.

command/ca/provisioner/provisioner.go
```diff
@@ -101,38 +108,68 @@ type crudClient interface {
...
+	resp, err := client.Get(u.String())
+	if err != nil {
+		return nil, fmt.Errorf("error checking for remote configuration API: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// If the response is a 404 then the Admin Remote Management API is not
+	// enabled. Therefore we default to using the local ca.json.
+	switch resp.StatusCode {
+	case http.StatusNotFound:
```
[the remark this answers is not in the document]
maraino: It's probably better to wrap this logic in ca's `AdminClient`, something like 

```go
func (c *AdminClient) IsEnabled() (bool, error) { ... }
```

maraino (changes requested):
It's probably better to wrap this logic in a method of the ca's `AdminClient`.

maraino (approved):
lgtm

PR #812, opened 2022-12-06
Update init.go adding dnsValue Validation at interactive prompt

jacostaperu:
#### Name of feature:
dnsValue validation to prevent panic when it is empty

#### Pain or issue this feature alleviates:
currently when calling _step ca init_   it accept empty string for dnsValue causing panic later

```
Ô×£  certificates git:(master) STEPDEBUG=1 step ca init
Ô£ö Deployment Type: Standalone
What would you like to name your new PKI?
Ô£ö (e.g. Smallstep): SmallStep
What DNS names or IP addresses would you like to add to your new CA?     
Ô£ö (e.g. ca.smallstep.com[,1.1.1.1,etc.]):    <- leave it empty
What IP and port will your new CA bind to?
Ô£ö (e.g. :443 or 127.0.0.1:443): :8443
What would you like to name the CA's first provisioner?
Ô£ö (e.g. you@smallstep.com): you@smallstep.com
Choose a password for your CA keys and first provisioner.
Smallstep CLI/0.21.0 (linux/amd64)
Release Date: 2022-07-06T22:23:54Z

panic: runtime error: index out of range [0] with length 0 [recovered]
        panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.panicHandler()
        ./main.go:139 +0x245
panic({0x12dec00, 0xc00003e1c8})
        runtime/panic.go:838 +0x207
github.com/smallstep/certificates/pki.New({{0x137b07c, 0x7}, {0x0, 0x0}, {0x0, 0x0}, 0x0, {0x0, 0x0}, {0x0, ...}, ...}, ...)
        github.com/smallstep/certificates@v0.21.0/pki/pki.go:382 +0xfe5
github.com/smallstep/cli/command/ca.initAction(0xc00055c2c0)
        github.com/smallstep/cli/command/ca/init.go:563 +0x3419
github.com/urfave/cli.HandleAction({0x11bec80?, 0x1467900?}, 0x4?)
        github.com/urfave/cli@v1.22.5/app.go:522 +0x7f
github.com/urfave/cli.Command.Run({{0x1372454, 0x4}, {0x0, 0x0}, {0x0, 0x0, 0x0}, {0x13a51fc, 0x15}, {0x14122b5, ...}, ...}, ...)
        github.com/urfave/cli@v1.22.5/command.go:173 +0x652
github.com/urfave/cli.(*App).RunAsSubcommand(0xc0006ed880, 0xc00055c000)
[...]

PR #815, opened 2022-12-13
Add check for empty DNS value in ca init

hslatman:


command/ca/init.go
```diff
@@ -845,3 +849,13 @@ func normalize(name string) string {
 	}
 	return name
 }
+
+// allEmpty loops through all strings in the slice and returns if
+// all are empty (length 0).
+func allEmpty(parts []string) bool {
+	sumLength := 0
+	for _, p := range parts {
+		sumLength += len(p)
```
dopey, line 858: This is fine, although seems like you could just stop iterating when you find one that is `> 0`, no? I imagine this list will be pretty short so probably doesn't make much difference.

dopey (dismissed):

hslatman:
I've taken your suggestion, @dopey. It's simpler 🙂 

By the way, there was another PR that addressed the issue at hand: https://github.com/smallstep/cli/pull/812. It didn't accept all things we do now, so we can't merge it. But I think the additional validations for the DNS hostname may be a nice addition to the CLI validators. WDYT?

command/ca/init.go
```diff
@@ -845,3 +849,14 @@ func normalize(name string) string {
 	}
 	return name
 }
+
+// allEmpty loops through all strings in the slice and returns if
+// all are empty (length 0).
+func allEmpty(parts []string) bool {
+	for _, p := range parts {
+		if len(p) > 0 {
```
dopey, line 857: Is their a reason we're specifically checking for `len` here? With string in golang I thought it was more idiomatic to check if it's equivalent to `""`.

hslatman: Fixed in 3dca2ff78a529a346abffaf41075be45a21b8104. You're right; it's more idiomatic that way.

PR #828, opened 2023-01-12
make step ca token usable on hsm-enabled step-ca instance

umegaya:
as title. if kms/pkcs11 is not imported, `kms.New` seems to fail for "kms" key enabled ca.json, because corresponding Register() will not be called via init() function of module. see https://github.com/smallstep/certificates/discussions/1226 for detail.

#### Name of feature:
step ca token improvement

you need to build step cli with CGO, like `make build CGO_OVERRIDE=""`

https://github.com/smallstep/certificates/discussions/1226

even if HSM is enabled, workflow using `step ca token` should work as expected

step cli help

hsm-enabled step-ca instance (usually on k8s pod)

hsm disabled step-ca instance

https://github.com/smallstep/certificates/discussions/1226

💔Thank you!

### DIFF
--- a/command/ca/init.go
+++ b/command/ca/init.go
@@ -25,6 +25,7 @@ import (
 	"go.step.sm/cli-utils/step"
 	"go.step.sm/cli-utils/ui"
 	"go.step.sm/crypto/kms"
+	_ "go.step.sm/crypto/kms/pkcs11"
 	"go.step.sm/crypto/pemutil"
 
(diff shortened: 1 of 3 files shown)


maraino:
The PR https://github.com/smallstep/cli/pull/871 included KMS support to `step ca token` using [step-kms-plugin](https://github.com/smallstep/step-kms-plugin)

PR #832, opened 2023-01-19
Add a basic `step ssh verify` command

hslatman:
This PR adds `verify` as an `ssh` subcommand. It verifies an SSH certificate using the `ssh.CertChecker` struct with basic options applied. Example output:

```console
# failure with wrong CA key:
step ssh verify example-cert.pub wrong_ca_key.pub
ssh certificate signed by "SHA256:AXEctpST7/1MfakrLrE+xrtF8Eixh6YsmqNaxiN6AFI" does not equal ssh CA "SHA256:ezEEbt1V5MzJctHhrfk4ftfQMgOvPL51KaU/9MLouUo"
exit status 1

# success: 
step ssh verify example-cert.pub ca_key.pub
```

Some potential things to add/change:
* make CA key optional? `verify` would not check the signer key if not provided.
* allow multiple CA keys?
* add additional verification, such as revocation check?

### DIFF
--- a/command/ssh/ssh.go
+++ b/command/ssh/ssh.go
@@ -95,6 +95,7 @@ $ ssh internal.example.com
 			rekeyCommand(),
 			renewCommand(),
 			revokeCommand(),
+			verifyCommand(),
 		},
 	}
 
--- a/command/ssh/verify.go
+++ b/command/ssh/verify.go
@@ -1,0 +1,104 @@
+package ssh
+
+import (
+	"bytes"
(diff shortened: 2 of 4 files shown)


weaversam8:
One additional suggestion for this command: ability to add an optional `--principal` or `--principals-file` option that would check the principals in the certificate against an expected principal or a list of principals in a file (assuming the same format as the file that the `AuthorizedPrincipalsFile` directive supports within sshd)

PR #837, opened 2023-01-26
Improve ACME `device-attest-01` challenge validation logic

hslatman:
This PR addresses https://github.com/smallstep/certificates/issues/1233.

When CLI is used with an outdated CA, it will show the error as follows:

```console
step ca certificate --attestation-uri yubikey:slot-id=82 12345678 attest.crt --context acmeda --provisioner acmestep
✔ Provisioner: acmestep (ACME)
Using Device Attestation challenge to validate "12345678" . Error!

Unable to validate challenge: Attestation statement cannot be verified
exit status 1
```

When used with a CA that includes the changes from https://github.com/smallstep/certificates/pull/1235, it shows the following:

```console
step ca certificate --attestation-uri yubikey:slot-id=82 12345678 attest.crt --context acmeda --provisioner acmestep
✔ Provisioner: acmestep (ACME)
Using `device-attest-01` challenge to validate "12345678" . Error!

Unable to validate challenge: challenge identifier "12345678" doesn't match the attested hardware identifier "87654321"
exit status 1
```


hslatman:
Depends on https://github.com/smallstep/certificates/pull/1235

utils/cautils/acmeutils.go
```diff
@@ -494,7 +496,76 @@ func doDeviceAttestation(ctx *cli.Context, ac *ca.ACMEClient, ch *acme.Challenge
...
+			isValid = true
+			break
+		}
+		ui.Printf(".")
+		time.Sleep(2 * time.Second) // NOTE: a bit shorter than the other challenge types
+	}
+	if !isValid {
+		ui.Printf(" Error!\n\n")
+		return errors.New(extractDetailedErrorMessageFromChallenge(vch))
+	}
```
maraino, line 537: It looks like we can refactor this, and use it for http-01 challenges too. Perhaps passing the sleep duration as a parameter.

maraino (changes requested):

utils/cautils/acmeutils.go
hslatman: Fixed in https://github.com/smallstep/cli/pull/837/commits/9df571805ce11deeebf860225105b717b5f4d2a0. The HTTP challenge didn't break early on `invalid`, but I think it makes sense to do it in that case too. The CA will invalidate challenges when the key authoriization isn't correct; that's not going to be fixed by repeating the request.

utils/cautils/acmeutils.go
```diff
@@ -494,7 +496,76 @@ func doDeviceAttestation(ctx *cli.Context, ac *ca.ACMEClient, ch *acme.Challenge
...
+
+	ui.Printf(" done!\n")
+	return nil
+}
+
+func extractDetailedErrorMessageFromChallenge(vch *acme.Challenge) string {
+	switch {
+	case vch == nil:
+		// shouldn't happen in normal flow
+		panic(stderrors.New("no ACME challenge to extract potential error from"))
```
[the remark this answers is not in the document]
hslatman: Fixed in https://github.com/smallstep/cli/pull/837/commits/9df571805ce11deeebf860225105b717b5f4d2a0.

utils/cautils/acmeutils.go
```diff
@@ -494,7 +483,94 @@ func doDeviceAttestation(ctx *cli.Context, ac *ca.ACMEClient, ch *acme.Challenge
...
+		// the retry logic intact for now.
+		if vch.Status == "invalid" {
+			break
+		}
+		if vch.Status == "valid" {
+			isValid = true
+			break
+		}
+		ui.Printf(".")
+		time.Sleep(durationBetweenAttempts) // NOTE: a bit shorter than the HTTP challenge type
```
maraino, line 534: The comment does not make sense now.

```suggestion
		time.Sleep(durationBetweenAttempts)
```

maraino (dismissed):
Looks good but we should remove one comment.

maraino (approved):

PR #839, opened 2023-01-31
Clarify init questions about CA addresses

tashian:


command/ca/init.go
```diff
@@ -523,9 +523,9 @@ func initAction(ctx *cli.Context) (err error) {
 	if pkiOnly {
 		pkiOpts = append(pkiOpts, pki.WithPKIOnly())
 	} else {
-		ui.Println("What DNS names or IP addresses would you like to add to your new CA?",
+		ui.Println("What DNS names and IP addresses will clients use to reach your CA?",
```
dopey, line 526: This makes me think I need DNS names **and** IP addresses.

dopey (commented):
As it's written, the new prompt makes me think I need both an IP address and a DNS name.

command/ca/init.go
tashian: good call, changed it back to `or`

jdoss:
LGTM Carl. You are my hero.

jdoss (approved):

PR #840, opened 2023-01-31
Fix `goimports` error in GitHub actions

hslatman:
Noticed some issues running actions for https://github.com/smallstep/cli/pull/837. Some file wasn't formatted properly and resulted in the [following output](https://github.com/smallstep/cli/actions/runs/4055757550/jobs/6979640264) on GitHub Actions:

```
 Running [/home/runner/golangci-lint-1.50.1-linux-amd64/golangci-lint run --out-format=github-actions --timeout=30m] in [] ...
  Error: File is not `goimports`-ed (goimports)
```

But no further info on what file it was about.

File might've been in the Actions cache, as I hadn't pulled latest `master` yet.

### DIFF
--- a/command/ca/init.go
+++ b/command/ca/init.go
@@ -566,7 +566,7 @@ func initAction(ctx *cli.Context) (err error) {
 		if helm {
 			ui.Println("What IP and port will your new CA bind to (it should match service.targetPort)?", ui.WithValue(ctx.String("address")))
 		} else {
-            ui.Println("What IP and port will your new CA bind to? (:443 will bind to 0.0.0.0:443)", ui.WithValue(ctx.String("address")))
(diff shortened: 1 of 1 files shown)


dopey (approved):

tashian:
Thanks for catching this Herman. Do we need to revisit the action to make it clearer?
I'd love some kind of pre-commit hook that runs `go fmt`. I haven't taken the time to set that up.

hslatman:
Using a (shared) pre-commit could work, but needs some coordination and sometimes clashes with other tools doing the same thing at the same time. VS Code will perform some best practice changes automatically, for example.

So far I haven't seen this happening a lot. The main thing that confused me was that there was no filename in the Actions log output, so not easy to see what had to be fixed. Normally when using the linter it does output filenames, so I don't think the Action needs a fix. We could provide a `make fmt` (backed by `goimports`) that does the right thing and that can be plugged in a local git config, perhaps?

PR #861, opened 2023-03-02
Add a debian:bullseye base image

tashian:
Related to smallstep/certificates#1286

- [ ] build this image as part of the release process
- [ ] tag it with `:bullseye` in Docker Hub

### DIFF
--- a/docker/Dockerfile.debian
+++ b/docker/Dockerfile.debian
@@ -18,8 +18,7 @@ RUN apt-get update \
         && apt-get upgrade \
         && apt-get install -y --no-install-recommends curl \
         && addgroup --gid ${STEPGID} step \
-        && adduser --disabled-password --uid ${STEPUID} --gid ${STEPGID} step \
-        && chown step:step /home/step
+        && adduser --disabled-password --uid ${STEPUID} --gid ${STEPGID} step
 
 
 COPY --from=builder /src/bin/step "/usr/local/bin/step"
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.debian
```diff
@@ -0,0 +1,32 @@
...
+FROM debian:bullseye
+
+ENV STEP="/home/step"
+ENV STEPPATH="/home/step"
+ARG STEPUID=1000
+ARG STEPGID=1000
+
+RUN apt-get update \
+        && apt-get upgrade \
+        && apt-get install -y --no-install-recommends bash curl tzdata \
```
maraino, line 19: `bash` and `tzdate` are in the default `debian:bullseye` image.

I guess curl is a nice-to-have thing.

docker/Dockerfile.debian
```diff
@@ -0,0 +1,32 @@
...
+ENV STEPPATH="/home/step"
+ARG STEPUID=1000
+ARG STEPGID=1000
+
+RUN apt-get update \
+        && apt-get upgrade \
+        && apt-get install -y --no-install-recommends curl \
+        && addgroup --gid ${STEPGID} step \
+        && adduser --disabled-password --uid ${STEPUID} --gid ${STEPGID} step \
+        && chown step:step /home/step
```
maraino, line 22: `chown` is not needed, adduser will do that by default:

```suggestion
        && adduser --disabled-password --uid ${STEPUID} --gid ${STEPGID} step
```

maraino (approved):

PR #873, opened 2023-03-09
Add bundle to x509 needs-renewal command.

### LINKED ISSUE #870: [Bug]: `step certificate needs-renewal` should only run against leaf certificate by default.
### Steps to Reproduce

```
step certificate needs-renewal foo.crt --expires-in 10%
percentUsed = 0.4698455540134039
percentThreshold = 10
percentUsed = 10.521388327147696
percentThreshold = 10
```

### Your Environment

* OS - Mac OS
* `step` CLI Version - 0.23.2

### Expected Behavior

I think the expectation here would be that the needs-renewal is only for the leaf certificate. Otherwise it's difficult to make an automated renewal decision.  E.g., if we return the certificate does need renewal, but it's actually one of the intermediates that trips the comparison, that would surprise me as a user.

### Actual Behavior

All certificates in the chain are checked for needing renewal.

### Additional Context

Feels like `--bundle` should be used if all certificates in the chain are expected to be checked.

### Contributing

dopey:
Fixes #870 

- Fixes unexpected behavior of implicitly applying needs-renewal to every cert in the bundle


command/certificate/needsRenewal.go
```diff
@@ -109,6 +120,11 @@ authenticity of the remote server.
 
     **directory**
 	:  Relative or full path to a directory. Every PEM encoded certificate from each file in the directory will be used for path validation.`,
+			},
+			cli.BoolFlag{
+				Name: `bundle`,
+				Usage: `Check all certificates in the order in which they appear in the bundle.
+By default (without this flag) this command will only check the leaf certificate.`,
```
maraino, line 127: I think the default option can be clarified in the description.

maraino (commented):
The old code has also an exit code `2` that is not documented.

command/certificate/needsRenewal.go
```diff
@@ -22,13 +23,16 @@ func needsRenewalCommand() cli.Command {
 		Action: cli.ActionFunc(needsRenewalAction),
 		Usage:  `Check if a certificate needs to be renewed`,
 		UsageText: `**step certificate needs-renewal** <cert-file or hostname>
-[**--expires-in**=<percent|duration>] [**--roots**=<root-bundle>] [**--servername**=<servername>]`,
+[**--expires-in**=<percent|duration>] [**--bundle] [**--verbose]
```
maraino, line 26: Two missing **

command/certificate/needsRenewal.go
```diff
@@ -48,11 +52,22 @@ Check if certificate.crt has passed 66 percent of its validity period:
 $ step certificate needs-renewal ./certificate.crt
 '''
 
+Check if any certificate in the bundle has passed 66 percent of it's validity period:
```
maraino, line 55: Clarify in the first example that is the leaf certificate. `Check if the leaf certificate.crt ...`

command/certificate/needsRenewal.go
```diff
@@ -185,12 +210,26 @@ func needsRenewalAction(ctx *cli.Context) error {
...
+	return isVerboseExit(false, isVerbose)
+}
+
+func isVerboseExit(needsRenewal, isVerbose bool) error {
+	if isVerbose {
+		fmt.Println(needsRenewal)
+	}
+	if needsRenewal {
+		return nil
+	}
```
maraino, line 233: I didn't know that when the certificate doesn't need renewal shows the following:

```
$ step certificate needs-renewal acme.crt
certificate does not need renewal
```

So perhaps the verbose option is not necessary.

command/ssh/needsRenewal.go
```diff
@@ -130,17 +137,27 @@ func needsRenewalAction(ctx *cli.Context) error {
...
+	return isVerboseExit(false, isVerbose)
+}
+
+func isVerboseExit(needsRenewal, isVerbose bool) error {
+	if isVerbose {
+		fmt.Println(needsRenewal)
+	}
+	if needsRenewal {
+		return nil
+	}
```
maraino, line 161: Same comment as before. Perhaps show something only if needs renewal is false?

command/certificate/needsRenewal.go
```diff
@@ -110,6 +126,14 @@ authenticity of the remote server.
     **directory**
 	:  Relative or full path to a directory. Every PEM encoded certificate from each file in the directory will be used for path validation.`,
 			},
+			cli.BoolFlag{
+				Name:  `bundle`,
+				Usage: `Check all certificates in the order in which they appear in the bundle.`,
+			},
+			cli.BoolFlag{
+				Name:  "verbose, v",
+				Usage: `Return "true" or "false" in the terminal.`,
```
maraino, line 135: The verbose usage is not right. We don't show true or false.

command/ssh/needsRenewal.go
```diff
@@ -66,6 +68,10 @@ character. If using <duration>, the input must be a sequence of decimal numbers,
 each with optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
 Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
 			},
+			cli.BoolFlag{
+				Name:  "verbose, v",
+				Usage: `Return "true" or "false" in the terminal.`,
```
maraino, line 73: ditto.

maraino (changes requested):

maraino (approved):

PR #876, opened 2023-03-10
Docker cross-compilation for 6x build speedup

tashian:
- [x] build a release candidate and pull it down from dockerhub on another platform and make sure it works.
- [x] check the version number and build date (probably broken)

Inspired by [this blog post](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/), I spun up a VM, pulled this repo, and ran the following:
```
docker buildx prune
docker buildx build --platform linux/amd64,linux/386,linux/arm,linux/arm64 -f docker/Dockerfile
```

Using our current Dockerfile, which emulates each platform:
```
[+] Building 1110.2s (43/43) FINISHED
```

Using my cross-compilation mod in this branch:
```
[+] Building 183.1s (34/34) FINISHED
```

I haven't tested that these images actually function! But this seems promising.

Docker version info:
```
Client: Docker Engine - Community
 Version:           23.0.1
 API version:       1.42
 Go version:        go1.19.5
 Git commit:        a5ee5b1
 Built:             Thu Feb  9 19:48:02 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.1
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.5
  Git commit:       bc3805a
  Built:            Thu Feb  9 19:48:02 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.18
  GitCommit:        2456e983eb9e37e47538f59ea18f2043c9a73640
 runc:
  Version:          1.1.4
  GitCommit:        v1.1.4-0-g5fd4c4d
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```


docker/Dockerfile
```diff
@@ -1,12 +1,23 @@
-FROM golang:alpine AS builder
-
-RUN mkdir /src
-ADD . /src
-
-RUN apk add --no-cache make git curl && \
-	cd /src && \
-	mkdir -p output/bin && \
```
tashian, line 8: @dopey this `output/bin` folder never seemed to be used in the rest of the build so I removed this mkdir

docker/Dockerfile
```diff
@@ -1,12 +1,23 @@
...
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/go/pkg \
+    GOOS="${TARGETOS}"; \
+    GOARCH="${TARGETARCH}"; \
+    if [ "${TARGETARCH}" = "arm" ] && [ "${TARGETVARIANT}" ]; then \
+    GOARM="${TARGETVARIANT#v}"; \
```
tashian, line 17: this is a new addition that will allow us to build more arm variants than arm64

docker/Dockerfile
```diff
@@ -1,12 +1,23 @@
...
+WORKDIR /src
+COPY go.mod go.sum .
+RUN go mod download
+COPY . .
+RUN apk add --no-cache make git curl
+
+ARG TARGETOS
+ARG TARGETARCH
+ARG TARGETVARIANT
+RUN --mount=type=cache,target=/root/.cache/go-build \
```
tashian, line 12: These cache mounts really only improve performance on iterative builds of the same image. They won't speed up the release.

docker/Dockerfile
```diff
@@ -1,12 +1,23 @@
...
+    --mount=type=cache,target=/go/pkg \
```
tashian, line 13: These cache mounts really only improve performance on iterative builds of the same image. They won't speed up the release, and we could remove them to keep this file simpler.

tashian:
@dopey I've tested this with [a release candidate](https://github.com/smallstep/cli/actions/runs/4420248402) and saw docker build times go from 40m to 4m. [Here's the final images](https://hub.docker.com/layers/smallstep/step-cli/0.23.5-rc.2/images/sha256-4bd49d3079d81832147252b6a5acbf570b8fb876877a36306ca6a62d4848263c?context=explore) in case you want to try them out.

tashian:
Finally, a basic (non-buildx) build continues to work fine. However, it kicks out some warnings when run with Podman:

```
time="2023-03-14T17:00:57-07:00" level=warning msg="missing \"TARGETOS\" build argument. Try adding \"--build-arg TARGETOS=<VALUE>\" to the command line"
time="2023-03-14T17:00:57-07:00" level=warning msg="missing \"TARGETARCH\" build argument. Try adding \"--build-arg TARGETARCH=<VALUE>\" to the command line"
time="2023-03-14T17:00:57-07:00" level=warning msg="missing \"TARGETVARIANT\" build argument. Try adding \"--build-arg TARGETVARIANT=<VALUE>\" to the command line"
```

If we want to avoid these, I think we'd need another Dockerfile.

docker/Dockerfile
maraino: /go/pkg has a directory `mod` with the source code of the packages, this should common for everybody, and a $GOOS_$GOARCH directory with compiled files, that are only for that specific combination, e.g. `linux_amd64`

maraino: I experimented with something similar but wasn't building everything in the same image. I used a similar cache and a different builder image because I needed dependencies for each architecture. I didn't merge it because, at least for that project, docker was taking more time to COPY from the cache than Go downloading the packages. This was my PR https://github.com/smallstep/step-kms-plugin/pull/34

dopey (approved):

PR #877, opened 2023-03-10
Add --service for running as a windows service for renewal and rekeying

unreality:
#### Name of feature:

Adds a new command line flag `--service` which will run as a Windows Service. Users can install the service by running `sc.exe create step-renew binPath= "path_to_step_cli.exe ca renew --service --ca-url=the_ca_url --root=path_to_root_ca.crt certfile.crt keyfile.key` as an Administrator.

Allows renewal via windows service as an alternative to --daemon on windows hosts

CLI usage examples have been added

Windows hosts

Non-windows hosts (use `--daemon` instead)

💔Thank you!

### DIFF
--- a/command/ca/rekey.go
+++ b/command/ca/rekey.go
@@ -185,6 +185,13 @@ periodically. By default the daemon will rekey a certificate before 2/3 of the
 time to expiration has elapsed. The period can be configured using the
 **--rekey-period** or **--expires-in** flags.`,
 			},
+			cli.BoolFlag{
+				Name: "service",
+				Usage: `Run the rekey command as a windows service, rekeying and overwriting the certificate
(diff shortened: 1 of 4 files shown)


maraino:
I suppose we can also add OS custom flags of files that use go tags.

```go 
#go:build windows

package ca

import "..."

func rekeyExtraFlags() []cli.Flag {
    return []cli.Flag
        cli.BoolFlag{Name: "service", Usage: "..."},
    }
}
```

The problem with adding this is that we cannot keep the order, but those flags won't appear on other OSs.

PreterPant:
Have a look here on how to implement the interface for a Windows service.
https://github.com/golang/sys/blob/master/windows/svc/example/main.go

PR #878, opened 2023-03-10
Fix error when using `--insecure` and creating RSA keys <2048 bits

hslatman:
Before this commit, the CLI would allow specifying RSA key sizes smaller than 2048 when also setting `--insecure`, but creation of the key would still fail with `the size of the RSA key should be at least 2048 bits`, originating from the `keyutil.generateRSAKey` function.

This commit enables the insecure mode for the `keyutil` package when `--insecure` is provided, allowing short RSA keys to be created.

While there are existing tests for checking (in)compatible CLI flags that involve short RSA keys, the command didn't seem to be actually executed as part of the test. That's part of the reason why this bug wasn't found earlier, I think.

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -459,9 +459,9 @@ func createAction(ctx *cli.Context) error {
 		return err
 	}
 
-	insecure := ctx.Bool("insecure")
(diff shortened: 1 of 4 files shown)


maraino (approved):
lgtm

PR #889, opened 2023-03-22
Remove 'not currently supported' from step ca revoke help text for active revocation

tashian:
Should this also say "Active revocation is only supported in Certificate Manager: https://u.step.sm/cm"?


maraino (approved):
step-ca also supports CRLs, so I don't think we need to say anything

PR #904, opened 2023-04-11
Remove duplicated usage package. Real one lives in cli-utils

dopey:


maraino (changes requested):
It looks you need to remove blackfriday too. There seem to be small changes between both, probably related to linting warnings.

maraino (approved):
This is ok, but there are differences between this files and the ones in cli-utils, the main one seems to be the `--hugo` flag. It is present in the cli, not in cli-utils. There are other differences probably related to linting, e.g.:
```diff
-	return cli.ShowAppHelp(ctx)
+	cli.ShowAppHelp(ctx)
+	return nil
```

PR #908, opened 2023-04-12
Add `--certificate` flag to `step ssh fingerprint` command

hslatman:
With the changes from https://github.com/smallstep/crypto/pull/207, the default behavior of `step ssh fingerprint` changes to be like the behavior of `ssh-add` (and similar tools). When a fingerprint is determined for an SSH certificate, the fingerprint will only include the bytes of the public key. With the `--certificate` flag, a user can create a fingerprint for the entire SSH certificate contents. 

In general, users are most likely to want the behavior of `ssh-add`, so that's the default.


azazeal (approved):
LGTM

PR #913, opened 2023-04-13
Update reference manifest after generating new help text

tashian:
Not sure if we want to do this this way, but here's a jq script that will generate and update the `manifest.json` in docs when the reference markdown is generated.

See https://linear.app/smallstep/issue/ZEN-2/updates-needed-for-step-help-reference-update-in-release-workflow

**Do not merge this until ZEN-2 is done.**


dopey (dismissed):

tashian:
We can merge this once smallstep/docs#234 is merged

tashian:
This is ready to be merged now.

dopey (approved):

PR #915, opened 2023-04-14
Small improvements for TPM attestation

hslatman:


maraino (approved):
Looks good.

I wonder if the change we made, calling the endpoint enroll instead of attest was right.

PR #942, opened 2023-05-17
Add `--ca-kms` flag to `step certificate create`

hslatman:
The `--ca-kms` flag can be used to specify a different KMS to be used for the CA signer.

@maraino What are your thoughts on this? Replacing `kms` with `caKMS` is a backwards incompatible change. Not a big issue I think, because this flow is most probably used when invoking `step` manually. I think being able to specify a different KMS for the CA signing the certificate is useful, because the current assumption seems to be that keys will be in the same KMS. That may not be true necessarily, if someone wants their root offline (e.g. an USB token) and the intermediate to be in an online system.

An alternative would be to override the `--kms` flag with the value from `--ca-kms` if it's set. One thing that won't work in that case is when we don't specify a `--ca-kms`, but still want to use the `SoftKMS` implementation. We could catch the `--ca-kms softkms:` case before invoking `step-kms-plugin` to allow for that? Another option might be toprovide another flag (e.g. `--ignore-kms`) to indicate that the `--kms` flag must not be used as the CA KMS. 

This PR includes the changes from https://github.com/smallstep/cli/pull/945 and https://github.com/smallstep/cli/pull/946.


maraino:
As we talked, it makes sense to have both `--kms` and `--ca-kms`, one for creating the CSR and the other for signing it.

A better solution would be not to require them, but this will require changes in other places.

hslatman:
After the changes in https://github.com/smallstep/crypto/pull/240 are available here, `--ca-kms` may no longer be required. A key in a different KMS can then be specified using the `--key <kms:>name` notation. We _could_ keep it around, though.

maraino:
> After the changes in https://github.com/smallstep/crypto/pull/240 are available here, --ca-kms may no longer be required. A key in a different KMS can then be specified using the --key <kms:>name notation. We could keep it around, though.

To remove the need of `--ca-kms` we will need to change the `cryptoutil` package to directly support URIs. See https://github.com/smallstep/cli/blob/c1eb16a6fa1903b4756240027c5f3c31b3185f7d/internal/cryptoutil/cryptoutil.go#L31-L44

maraino (changes requested):
Add new flags at the list of flags at the beginning of `UsageText`.

This is not 100% backward compatible with the previous use of `--kms` but this might be ok. The problem is that you might need to specify both flags when before you only needed one. It was convenient for the same KMS but was problematic for cross-kms signing.

command/certificate/create.go
```diff
@@ -766,9 +812,9 @@ func parseSigner(ctx *cli.Context, defaultSigner crypto.Signer) (*x509.Certifica
 	var (
 		caCert   = ctx.String("ca")
 		caKey    = ctx.String("ca-key")
+		caKMS    = ctx.String("ca-kms")
 		profile  = ctx.String("profile")
 		template = ctx.String("template")
-		kms      = ctx.String("kms")
```
maraino, line 771: This change makes the command incompatible with at least one example in the help:

```shell
# Create an intermediate certificate using step-kms-plugin:

$ step kms create \
  --kms 'pkcs11:module-path=/usr/local/lib/softhsm/libsofthsm2.so;token=smallstep?pin-value=password' \
  'pkcs11:id=4001;object=intermediate-key'
$ step certificate create \
  --profile intermediate-ca \
  --kms 'pkcs11:module-path=/usr/local/lib/softhsm/libsofthsm2.so;token=smallstep?pin-value=password' \
  --ca root_ca.crt --ca-key 'pkcs11:id=4000' \
  --key 'pkcs11:id=4001' \
  'My KMS Intermediate' intermediate_ca.crt
```
```

maraino (changes requested):
The current behavior is not backward compatible as the changelog says, but the help still shows examples that now will fail.

command/certificate/create.go
hslatman: Fixed in 290f81d10390b8cb2ab39d65ce061410628cf8f2

maraino (approved):
lgtm

PR #945, opened 2023-05-25
Pass key to be signed as signer for the CSR instead of signing key

hslatman:
Quick bug fix for the CSR signer.

Note: this builds on top of the `--ca-kms` changes.


maraino (approved):
👍

PR #946, opened 2023-05-25
Add `--skip-csr-signature` option

hslatman:
When creating a certificate for a public key backed by a KMS that doesn't allow the key to also be used for signing, or in cases where the private key isn't readily available to sign the (internally used) CSR, `--skip-csr-signature` can be passed to skip signing the (internally used) CSR.

This option is not compatible with `--csr`, because that requires a CSR with a valid signature to be produced.

Note: this builds on the changes in #945 and depends on the changes in ~~https://github.com/smallstep/crypto/pull/239 to be finalized. After merging the changes from https://github.com/smallstep/crypto/pull/239~~ https://github.com/smallstep/crypto/pull/248. 

Will add example usage when updating with the latest `smallstep/crypto`.

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -447,7 +447,12 @@ the **--ca** flag.`,
 				Hidden: true,
 			},
 			cli.StringFlag{
-				Name: "ca-kms",
+				Name:  "ca-kms",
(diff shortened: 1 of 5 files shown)


maraino (commented):
The API will change.

maraino (approved):
lgtm

PR #950, opened 2023-06-01
Add SCEP decrypter configuration flags

hslatman:


command/ca/provisioner/add.go
```diff
@@ -787,16 +800,77 @@ func createSCEPDetails(ctx *cli.Context) (*linkedca.ProvisionerDetails, error) {
...
+		}
+		if block == nil {
+			return nil, errors.New("failed parsing decrypter certificate: no PEM block found")
+		}
+		if _, err := x509.ParseCertificate(block.Bytes); err != nil {
+			return nil, fmt.Errorf("failed parsing decrypter certificate: %w", err)
+		}
+		decrypter.Certificate = data
+		s.Decrypter = decrypter
+	}
```
maraino, line 839: I don't think we need a flag for passing the base64 encoded certificate. Using a file would be enough.

command/ca/provisioner/add.go
```diff
@@ -787,16 +800,77 @@ func createSCEPDetails(ctx *cli.Context) (*linkedca.ProvisionerDetails, error) {
...
+		data, err := base64.StdEncoding.DecodeString(decrypterKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed base64 decoding decrypter key: %w", err)
+		}
+		if err := validateSCEPDecrypterKey(data); err != nil {
+			return nil, fmt.Errorf("failed validating decrypter key: %w", err)
+		}
+		decrypter.Key = data
+		s.Decrypter = decrypter
+	}
```
maraino, line 862: ditto

command/ca/provisioner/add.go
```diff
@@ -787,16 +800,77 @@ func createSCEPDetails(ctx *cli.Context) (*linkedca.ProvisionerDetails, error) {
...
+	if decrypterKeyPasswordFile := ctx.String("scep-decrypter-key-password-file"); decrypterKeyPasswordFile != "" {
+		decrypterKeyPassword, err := utils.ReadPasswordFromFile(decrypterKeyPasswordFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed reading decrypter key password from %q: %w", decrypterKeyPasswordFile, err)
+		}
+		decrypter.KeyPassword = decrypterKeyPassword
+		s.Decrypter = decrypter
```
maraino, line 869: You can read the password first and validate the decrypter key with it.

maraino (changes requested):

command/ca/provisioner/add.go
hslatman: Fixed in 642b9b907d1e1ec042aed4974260d5bac6d43cb6.

command/ca/provisioner/add.go
hslatman: Fixed in 642b9b907d1e1ec042aed4974260d5bac6d43cb6.

command/ca/provisioner/add.go
hslatman: Thought about doing that locally in the CLI, but it's already happening on the CA side, with the somewhat handy consequence that it's not always necessary to specify the password. If the CA already knows the password, it'll be reused when the provisioner is initialized as part of validation. If there's no password, the "password required in non-interactive context" error is thrown (if not provided).

command/ca/provisioner/add.go
```diff
@@ -787,16 +799,58 @@ func createSCEPDetails(ctx *cli.Context) (*linkedca.ProvisionerDetails, error) {
...
+	if decrypterKey := ctx.String("scep-decrypter-key"); decrypterKey != "" {
+		data, err := base64.StdEncoding.DecodeString(decrypterKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed base64 decoding decrypter key: %w", err)
+		}
+		if err := validateSCEPDecrypterKey(data); err != nil {
+			return nil, fmt.Errorf("failed validating decrypter key: %w", err)
+		}
+		decrypter.Key = data
+		s.Decrypter = decrypter
```
maraino, line 841: This should read a file.

maraino (changes requested):

command/ca/provisioner/add.go
hslatman: There's another flag for that. Removed this one in dc047bbf1367488f03e21738aed7e85606c50747.

maraino (approved):
lgtm

PR #953, opened 2023-06-06
Detect OIDC tokens issued by Kubernetes

areed:
Service account tokens issued by the Kubernetes API server in recent versions are valid OIDC tokens but do not contain an email in the payload. Update the Payload Type test to interpret tokens with [all five required claims](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) as OIDC tokens even if they do not have an email claim.

An example Kubernetes OIDC token:
```json
{
  "aud": [
    "step-ca"
  ],
  "exp": 1686076066,
  "iat": 1686072466,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "kubernetes.io": {
    "namespace": "default",
    "pod": {
      "name": "pod",
      "uid": "98844dbc-d248-4ddb-b385-81643892908d"
    },
    "serviceaccount": {
      "name": "default",
      "uid": "e79de63f-8842-46bc-9931-c277c4993994"
    }
  },
  "nbf": 1686072466,
  "sub": "system:serviceaccount:default:default"
}
```

#### Name of feature:
Detect OIDC tokens without email claim.

OIDC tokens issued by Kubernetes are interpreted as type unknown and trying to get a certificate results in error `token is not supported`.

💔Thank you!

### DIFF
--- a/token/parse.go
+++ b/token/parse.go
@@ -74,8 +74,6 @@ func (p Payload) Type() Type {
 		return K8sSA
 	case len(p.SHA) > 0 || len(p.SANs) > 0:
 		return JWK
-	case p.Email != "":
-		return OIDC
 	case len(p.Audience) > 0 && p.Issuer != "" && p.Subject != "" && !p.Expiry.Time().IsZero() && !p.IssuedAt.Time().IsZero():
 		return OIDC
 	default:
--- a/token/parse.go
+++ b/token/parse.go
@@ -76,6 +76,8 @@ func (p Payload) Type() Type {
 		return JWK
 	case p.Email != "":
 		return OIDC
+	case len(p.Audience) > 0 && p.Issuer != "" && p.Subject != "" && !p.Expiry.Time().IsZero() && !p.IssuedAt.Time().IsZero():
+		return OIDC
 	default:
 		return Unknown
 	}
--- a/token/parse_test.go
+++ b/token/parse_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"go.step.sm/crypto/jose"
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


token/parse_test.go
```diff
@@ -290,20 +291,22 @@ func TestPayload_Type(t *testing.T) {
...
-		{"Unknown", fields{"", nil, "", nil, nil, nil}, Unknown},
+		{"JWK", fields{"a-sha", []string{"foo.bar.zar"}, "", nil, nil, nil, nil}, JWK},
+		{"JWK no sans", fields{"a-sha", nil, "", nil, nil, nil, nil}, JWK},
+		{"JWK no sha", fields{"", []string{"foo.bar.zar"}, "", nil, nil, nil, nil}, JWK},
+		{"OIDC", fields{"", nil, "mariano@smallstep.com", nil, nil, nil, nil}, OIDC},
+		{"GCP", fields{"", nil, "", &GCPGooglePayload{}, nil, nil, nil}, GCP},
+		{"AWS", fields{"", nil, "", nil, &AWSAmazonPayload{}, nil, nil}, AWS},
+		{"Azure", fields{"", nil, "", nil, nil, &AzurePayload{}, nil}, Azure},
+		{"Unknown", fields{"", nil, "", nil, nil, nil, nil}, Unknown},
+		{"OIDC Kubernetes", fields{"", nil, "", nil, nil, nil, &jose.Claims{Audience: jwt.Audience{"step-ca"}, Issuer: "https://kubernetes.default.svc.cluster.local", Subject: "system:serviceaccount:default:default", Expiry: jwt.NewNumericDate(time.Now()), IssuedAt: jwt.NewNumericDate(time.Now())}}, OIDC},
```
maraino, line 309: You can use or wrapper and use `jose.NewNumericDate()` it will be easier to move to go-jose/v3

token/parse.go
```diff
@@ -76,6 +76,8 @@ func (p Payload) Type() Type {
 		return JWK
 	case p.Email != "":
 		return OIDC
+	case len(p.Audience) > 0 && p.Issuer != "" && p.Subject != "" && !p.Expiry.Time().IsZero() && !p.IssuedAt.Time().IsZero():
+		return OIDC
```
maraino, line 80: Any well-formed token will be like this, wouldn't I make more sense to look for `kubernetes.io`?

areed: That would work but what if there are other issuers besides Kubernetes that don't use email in the claims? With this being the last check before interpreting the type as unknown would there be any drawbacks to saying that if it's a valid OIDC token then it gets the OIDC type?

maraino: Sure, we could even remove the case p.Email != "" and just use that, those fields are required per OIDC spec.

token/parse_test.go
areed: 948d88c4

token/parse.go
areed: 948d88c4

token/parse_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"go.step.sm/crypto/jose"
+	"gopkg.in/square/go-jose.v2/jwt"
```
maraino, line 11: You can avoid this import using jose.Audience. `"go.step.sm/crypto/jose"` has aliases for almost everything we use.

maraino (changes requested):
Let's try to get rid of the gopkg.in/square/go-jose.v2 import.

token/parse_test.go
areed: a5e944ec

maraino (approved):
👍

PR #962, opened 2023-06-13
[action] use stable go version | bump setup-go version

dopey:


maraino (approved):
lgtm.
FYI: according to this [note](https://github.com/actions/setup-go#using-stableoldstable-aliases), `check-latest: true` might be unnecessary.

PR #976, opened 2023-07-06
Look for Windows plugins with executable extensions in plugins

hslatman:
This PR adds support for locating Windows plugins with executable extensions. The logic is somewhat similar to what's in `exec.LookPath`, but only for the `$(step path)/plugins` directory.

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -4,6 +4,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 
 	"github.com/urfave/cli"
 	"go.step.sm/cli-utils/step"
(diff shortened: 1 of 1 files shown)


areed (approved):
Is there any benefit to adding a check on Windows for the raw filename in case it already has an extension, like [the standard library](https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/os/exec/lp_windows.go;l=42)?

hslatman:
I don't think so, unless I'm misunderstanding your remark.

The plugin filename is constructed in code based on just the plugin name. There's no extension added to the constructed filename. Placing a binary without an executable extensions in the `$(step path)/plugins` directory on Windows results in the plugin not being discovered as an executable (and I think the same holds for `PATH`) by the CLI.

PR #980, opened 2023-07-11
Create API tokens with team slug

areed:
#### Name of feature:
API auth tokens with team slug

Previous you needed to supply the team UUID to get an auth token for the API.

Getting an auth token from the Gateway API.

💔Thank you!

### DIFF
--- a/command/api/token/create.go
+++ b/command/api/token/create.go
@@ -21,7 +21,7 @@ func createCommand() cli.Command {
 		Name:   "create",
 		Action: cli.ActionFunc(createAction),
 		Usage:  "create a new token",
-		UsageText: `**step api token create** <team-id> <crt-file> <key-file>
+		UsageText: `**step api token create** <team-id>|<team-slug> <crt-file> <key-file>
 [**--api-url**=<url>] [**--audience**=<name>]
 `,
 		Flags: []cli.Flag{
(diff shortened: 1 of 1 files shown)


command/api/token/create.go
```diff
@@ -21,7 +21,7 @@ func createCommand() cli.Command {
 		Name:   "create",
 		Action: cli.ActionFunc(createAction),
 		Usage:  "create a new token",
-		UsageText: `**step api token create** <team-id> <crt-file> <key-file>
+		UsageText: `**step api token create** <team-id>|<team-slug> <crt-file> <key-file>
```
maraino, line 24: I would use just <team> and use the text in the positional arguments to actually describe what that argument means.
```suggestion
		UsageText: `**step api token create** <team> <crt-file> <key-file>
```

command/api/token/create.go
```diff
@@ -32,8 +32,8 @@ func createCommand() cli.Command {
 
 ## POSITIONAL ARGUMENTS
 
-<team-id>
-:  UUID of the team the API token will be issued for. This is available in the Smallstep dashboard.
+<team-id>|<team-slug>
```
maraino, line 35:
```suggestion
<team>
```

maraino (changes requested):
Just a couple of suggestions for the help.

command/api/token/create.go
areed: a1f4fb76

command/api/token/create.go
areed: a1f4fb76

maraino (approved):
lgtm

PR #992, opened 2023-07-31
Find more plugins on Windows

hslatman:
This PR adds support for `PATHEXT` in the `(step path)/plugins` directory, as well as support for `.ps1` execution.

Example PowerShell script in `C:\Users\herman\.step\plugins\step-script-plugin.ps1`:
```powershell
Write-Host 'Hello from PowerShell, step!'
Write-Host $args
```

```console
C:\Users\herman\Desktop>step.exe script bla foo
Hello from PowerShell, step!
bla foo
```

Currently the behavior is to fallback to some well-known executable extensions, including `.ps1`, if `PATHEXT` is not set. If `PATHEXT` is set, it is used to extract the executable extensions. By default, `PATHEXT` does not include `.ps1`, so that should be added if necessary.

Not sure if `PATHEXT` makes sense in `(step path)/plugins`. It can go both ways.

### DIFF
--- a/internal/plugin/plugin.go
+++ b/internal/plugin/plugin.go
@@ -5,6 +5,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 
 	"github.com/urfave/cli"
 	"go.step.sm/cli-utils/step"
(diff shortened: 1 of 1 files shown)


maraino (approved):
lgtm. PowerShell is also available in Linux and macOS. It is probably not very popular in those OSs, we can change  later.

PR #973, opened 2023-06-27
Refactor ACME TPM attestation to use the TPMKMS

hslatman:

PR #1044, opened 2023-10-20
Allow the define a custom tpm device

maraino:
### Description

This commit allows passing the device name option when initializing a TPM in the acme flow. It also allows passing the attestation-ca-url in the URI, as we can use a similar URI in `step kms` commands and `step ca certificate`.


hslatman:
@maraino just came across https://github.com/smallstep/cli/pull/973/files, which I started working on a while ago. I think it does some of the same things you did in this PR, but I'm fine with merging yours.

maraino:
I'll merge this for the moment. I'm using it.

Issue #1051, opened 2023-10-24
[Bug]: document for leaf certificate template contains typo


zhangyoufu:
### Steps to Reproduce

```shell
% step help certificate create | grep -F typeIs
          {{- if typeIs "*sa.PublicKey" .Insecure.CR.PublicKey }}
% step help certificate sign | grep -F typeIs
          {{- if typeIs "*sa.PublicKey" .Insecure.CR.PublicKey }}
```

### Your Environment

```
% step version
Smallstep CLI/0.25.0 (darwin/amd64)
Release Date: 2023-09-27 12:02 UTC
```

### Expected Behavior

The type should be `*rsa.PublicKey`

### Actual Behavior

The output is `*sa.PublicKey`

### Contributing

Vote on this issue by adding a 👍 reaction.
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

hslatman:
Hi @zhangyoufu, thank you for reporting this. I've opened a PR: #1052

PR #1052, opened 2023-10-24
Fix leaf certificate `*rsa.PublicKey` example

hslatman:
As reported in https://github.com/smallstep/cli/issues/1051, the example template printed in the help has `*sa.PublicKey` instead of `*rsa.PublicKey`. This is likely due to the help text preprocessor.

This commit adds an underscore to `*_rsa.PublicKey`, which tricks the help text preprocessor into simply printing `*rsa.PublicKey`, ignoring it to be processed as Markdown.

A better fix might be to change the help text processor, but there are no tests, and it's used widely, so it's a riskier method for not too much gain.

Note that the output in the help text in the CLI is the same as on the website: https://smallstep.com/docs/step-cli/reference/certificate/create/#templates, so this PR fixes that issue too.

### DIFF
--- a/command/certificate/create.go
+++ b/command/certificate/create.go
@@ -84,7 +84,7 @@ Here's the default template used for generating a leaf certificate:
 {
 	"subject": {{ toJson .Subject }},
 	"sans": {{ toJson .SANs }},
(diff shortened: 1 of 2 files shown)


azazeal (approved):
LGTM

PR #1054, opened 2023-10-26
Add step rand command

maraino:
This commit adds a new command `step rand` that generates random strings

### DIFF
--- a/command/rand/rand.go
+++ b/command/rand/rand.go
@@ -1,0 +1,231 @@
+package rand
+
+import (
+	"bufio"
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/urfave/cli"
+	"go.step.sm/cli-utils/command"
+	"go.step.sm/cli-utils/errs"
+	"go.step.sm/crypto/fingerprint"
+	"go.step.sm/crypto/randutil"
+)
+
+func init() {
+	cmd := cli.Command{
+		Name:      "rand",
+		Action:    command.ActionFunc(randAction),
+		Usage:     "generate random strings",
+		UsageText: "**step rand** [<length>] [--format=<format>] [--dictionary=<file>]",
+		Flags: []cli.Flag{
+			cli.StringFlag{
+				Name:  "format",
+				Usage: "The <format> of the output string. See help for list of available formats.",
+			},
+			cli.StringFlag{
+				Name:  "dictionary,dict",
+				Usage: "The <file> to use as a dictionary to get random words.",
+			},
+		},
+		Description: `**step rand** generates random strings that can be used for multiple purposes.
+The <rand> command supports printing stings with different formats. It defaults
+to use the printable characters of the ASCII table; <rand> also supports
+generating a memorable password using words from a provided dictionary.
+
+The length of the random string will default to 32 characters or 6 words
+separated by a dash (-) if a dictionary is used.
+
+The list of supported formats is the following:
+
+* ascii: generates a random string using the 94 printable characters of the
+  ASCII table.
+* alphanumeric: uses the 62 alphanumeric characters in the POSIX/C locale
+  (a-z+A-Z+0-9).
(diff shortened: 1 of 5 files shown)


command/rand/rand.go
```diff
@@ -0,0 +1,231 @@
...
+
+## EXAMPLES
+
+Generate a random string using the default format (ascii) and length (32):
+'''
+$ step rand
+Ijghm(Y?pfZiTPkHv0Z=1@MC<n&gsMe|
+'''
+
+Generate a random memorable string using a dictionary words:
```
dopey, line 78:
```suggestion
Generate a random memorable string using a dictionary of words:
```

command/rand/rand.go
```diff
@@ -0,0 +1,231 @@
...
+
+func randWithDictionary(dictionary string, length int) error {
+	file, err := os.Open(dictionary)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	words := make([]string, 0, 1024)
```
dopey, line 208: why is the capacity here 1024?

dopey:
Overall I like this this addition, but I do want to say that I think it's strange to use dictionary.length and dice.length as the same default value. It's nice for now, but it's going to be a headache to decouple later.

hslatman:
I like the addition too, but would suggest to put it under `step crypto`.

command/rand/rand.go
```diff
@@ -0,0 +1,231 @@
...
+The following special formats are also supported:
+
+* dice: generates a random number between 1 and 6 or the given argument,
+* uuid: generates a UUIDv4.
+
+## POSITIONAL ARGUMENTS
+
+<length>
+:  The length of the random string in characters or words. If the dice format
+is used, the length is the maximum number of the dice.
```
hslatman, line 68:
```suggestion
is used, the length is the maximum number of the die.
```

command/rand/rand.go
```diff
@@ -0,0 +1,231 @@
...
+* lower: uses the 26 lowercase alphabetic characters in the POSIX/C locale
+  (a-z).
+* upper: uses the 26 uppercase alphabetic characters in the POSIX/C locale
+  (A-Z).
+* emoji: uses a curated list of 256 emojis that are not entirely similar.
+* raw: uses random bytes.
+
+The following special formats are also supported:
+
+* dice: generates a random number between 1 and 6 or the given argument,
```
hslatman, line 61: Technically, this is a multiple, but it's just a single die being rolled.

dopey, line 61:
```suggestion
* dice: generates a random number between 1 and 6 (by default), or the 1 and the provided argument.
```

dopey: I kinda agree here - I assumed the command was going to roll multiple dice for me before reading the example.

maraino: @dopey @hslatman should I change the format name to `die`?

dopey: Why even have 'die' or 'dice` as a special format? Like, why not just use `step rand int` or `step rand num` and then allow to specify the start and end inclusive or exclusive.

dopey: It seems like the only reason to use dice is because the number 6 matches the number of default words in the dictionary format - and I don't think that relation makes sense. I already commented about that earlier.

dopey (dismissed):

tashian:
Love this 🎲 and I will use it!
Thank you for including emoji 🥰

command/rand/rand.go
maraino: Fixed with [12263b8](https://github.com/smallstep/cli/pull/1054/commits/12263b880d51bd0a1aa2a391bdab7fe429e36845)

maraino: `step rand int` is a good addition that can be added later, but it requires more flags. 🎲 is fun. A die of 6 is the most common one. The match between the number of words in the dictionary and the default die is just a coincidence. Six looked like a good default.

tashian (approved):

PR #1055, opened 2023-10-27
Fix command completion for zsh

terabyte128:
zsh only reads the first line of a file when searching for `#compdef` lines. The completion script added a comment above this line and caused zsh autocomplete to fail.

https://zsh.sourceforge.io/Doc/Release/Completion-System.html

#### Name of feature

bugfix for zsh completion

zsh autocompletion didn't work before

yes, already exists

anyone using zsh

n/a

https://zsh.sourceforge.io/Doc/Release/Completion-System.html

> When compinit is run, it searches all such files accessible via fpath/FPATH and reads the first line of each of them. This line should contain one of the tags described below. Files whose first line does not start with one of these tags are not considered to be part of the completion system and will not be treated specially. 

💔Thank you!

### DIFF
--- a/command/completion/completion.go
+++ b/command/completion/completion.go
@@ -69,8 +69,7 @@ complete -o bashdefault -o default -F _step_cli_bash_autocomplete step
 
 `
 
-var zsh = `# zsh completion for step
-#compdef step
+var zsh = `#compdef step
 
 _step() {
 


maraino (approved):
Thanks @terabyte128.

PR #1060, opened 2023-11-03
feat: add: --password-file option for step crypto jwe decrypt

ncaq:
#### Name of feature:

add: `--password-file` option for `step crypto jwe decrypt`.

The `encryptedKey` generated by `--type=jwk` in `step ca provisioner add` is in PBES2 format. This is hard-coded and difficult to change.
Therefore, in the implementation before this commit, after the key to decrypt is entered in the standard input, the executor must interactively enter the password specified when generated in the standard input. The `--key` option and others cannot be used at this time because of formatting problems. However, trying to do this programmatically is difficult, as the format of `ui.PromptPassword` is too specific for manual execution, or reading stderr output, etc., and trying to add additional standard input passwords in a programmatic and automatic way is very It was very difficult. Therefore, I added the `--password-file` command line argument to provide an alternative that does not require an interactive operation. Since there was no problem reading the password file when generating it, I decided that there was no harm in reading the password file when decrypting it, and it was a natural implementation.

For example, once this is merged, anyone can easily and automatically run the `gh_token.sh` script for GitHub Enterprise that is on the official site.

[Configuring `step-ca` Templates](https://smallstep.com/docs/step-ca/templates/#ssh-template-examples)

The documentation is not specifically written because it is quite simple and a function that is already in other commands.

For now,

```
step crypto jwe decrypt --help
```

You can check it with ```` step crypto jwe decrypt --help ````.

It would be useful to automatically decrypt JWK's encryptedKey.
The reason why `--key` is not allowed is that, as I said before, it is rejected due to formatting problems.

I needed this feature for my own use in the first place.
As I said before, it was difficult to operate via standard input.
I didn't want to put too much load on the maintainer, so I refrained from doing so, but it was too difficult and I thought it would be a comfortable flag to add, so I added it.

### DIFF
--- a/command/crypto/jwe/decrypt.go
+++ b/command/crypto/jwe/decrypt.go
@@ -47,6 +47,10 @@ used with **--key** the <kid> value must match the **"kid"** member of the JWK. 
 used with **--jwks** (a JWK Set) the KID value must match the **"kid"** member of
 one of the JWKs in the JWK Set.`,
 			},
+			cli.StringFlag{
+				Name: "password-file",
+				Usage: `The path to the <file> containing the password to encrypt the keys.`,
+			},
 		},
 	}
 }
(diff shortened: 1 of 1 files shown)


ncaq:
I just realized that maybe it is not necessary because it can be specified with `step crypto jwt sign`.

maraino (approved):
Thanks @ncaq, lgtm

PR #1076, opened 2023-12-04
Add FIPS build.

joshdrake:
This PR adds a new Dockerfile using an Alpine base with OpenSSL compiled in FIPS-mode and a binary built with `boringcrypto`.

💔Thank you!


maraino (changes requested):
We should use tags a file like this:
```
//go:build fips
// +build fips

package something

import _ "crypto/tls/fipsonly"
```

hslatman:
Closing this, as it has been solved in a different way.

PR #1080, opened 2023-12-05
Docker: Update Debian build to use `debian:bookworm`

tashian:
I've tested that `step` builds and runs properly using the `debian:bookworm` base image and `golang:bookworm` builder image.


maraino (approved):
lgtm 📖 🪱

PR #1085, opened 2023-12-12
Fix Nebula key + cert verification

hslatman:
Nebula added support for using ECDSA with the P-256 curve: https://github.com/slackhq/nebula/pull/769.

This takes #1082, but fixes the breaking change.

I expect they might need some additional parameters in the future to be able to use other ECDSA curves too.


maraino (approved):

azazeal (approved):
LGTM

PR #1092, opened 2023-12-29
Add handling of `cnf` claim

maraino:
Allow to add confirmation claims to tokens

This commit allows passing confirmation claims to tokens to tie the
tokens with a provided CSR or SSH public key.

Fixes smallstep/certificates#1637

Related PR:
- [ ] https://github.com/smallstep/certificates/pull/1660


token/options.go
```diff
@@ -84,6 +87,43 @@ func WithSSH(v interface{}) Options {
...
+			return fmt.Errorf("unsupported fingerprint for %T", vv)
+		}
+
+		kid, err := fingerprint.New(data, crypto.SHA256, fingerprint.Base64RawURLFingerprint)
+		if err != nil {
+			return err
+		}
+		c.Set(ConfirmationClaim, map[string]string{
+			"kid": kid,
+		})
```
hslatman, line 122: For the X509 case, the value maybe shouldn't be put in `kid`, as we generally use that claim to carry the thumbprint of a JWK (used as key identifier); not the hash of a full request. Arguably, the `kid` can be filled arbitrarily, as long as it identifiers the key, so it's not wrong to put the hash of the CSR in there, but I think using a different name might be better.

There's `"cnf":{"x5t#S256":"...."}` in https://datatracker.ietf.org/doc/html/rfc8705#section-appendix.a, which is for certificates. Can we find/use/create a (custom) variant of that for certificate requests? E.g. `"x5rt#S256"`, or something like that?

maraino: Fixed with eb2564

flags/flags.go
```diff
@@ -379,6 +379,21 @@ be stored in the 'sshpop' header.`,
...
+		Name:  "cnf-file",
+		Usage: `The CSR or SSH public key <file> to restrict this token for.`,
+	}
+
+	// ConfirmationKid is a cli.Flag used to add a confirmation claim in the
+	// token.
+	ConfirmationKid = cli.StringFlag{
+		Name:  "cnf-kid",
+		Usage: `The <fingerprint> of the CSR or SSH public key to restrict this token for.`,
+	}
```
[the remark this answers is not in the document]
maraino: Fixed with eb2564

command/ca/token.go
```diff
@@ -186,6 +200,8 @@ multiple principals.`,
 			flags.SSHPOPKey,
 			flags.NebulaCert,
 			flags.NebulaKey,
+			flags.ConfirmationFile,
+			flags.ConfirmationKid,
```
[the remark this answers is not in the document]
maraino: Fixed with eb2564 (with `--cnf`)

token/options.go
```diff
@@ -84,6 +87,43 @@ func WithSSH(v interface{}) Options {
...
+	return func(c *Claims) error {
+		var data []byte
+		switch vv := v.(type) {
+		case *x509.CertificateRequest:
+			data = vv.Raw
+		case ssh.PublicKey:
+			data = vv.Marshal()
+		default:
+			return fmt.Errorf("unsupported fingerprint for %T", vv)
+		}
```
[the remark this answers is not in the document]
maraino: Fixed with eb2564

command/certificate/fingerprint.go
```diff
@@ -55,6 +55,12 @@ Get the fingerprints for a remote certificate with its intermediate:
 $ step certificate fingerprint --bundle https://smallstep.com
 e2c4f12edfc1816cc610755d32e6f45d5678ba21ecda1693bb5b246e3c48c03d
 25847d668eb4f04fdd40b12b6b0740c567da7d024308eb6c2c96fe41d9de218d
+'''
+
+Get the fingerprint for a CSR using base64-url without padding encoding:
+'''
+$ step certificate fingerprint --format base64-url-raw hello.csr
+PJLNhtQoBE1yGN_ZKzr4Y2U5pyqIGiyyszkoz2raDOw
```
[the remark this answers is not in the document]
maraino: Fixed with eb2564

utils/cautils/certificate_flow.go
```diff
@@ -35,16 +36,66 @@ type CertificateFlow struct {
...
+	CertificateRequest      *x509.CertificateRequest
+	ConfirmationFingerprint string
+}
+
 // sharedContext is used to share information between commands.
-var sharedContext = struct {
-	DisableCustomSANs bool
-}{}
+var sharedContext flowContext
+
```
hslatman, line 48: Might be better to pass through the `cli.Context` instead of relying on this shared state, but I remember that might take some more refactoring.

token/options.go
```diff
@@ -84,6 +86,40 @@ func WithSSH(v interface{}) Options {
...
+		c.Set(ConfirmationClaim, map[string]string{
+			"x5rt#S256": fp,
+		})
+		return nil
+	}
+}
+
+// WithFingerprint returns an Options function that the cnf claims with
+// "x5rt#S256" representing the fingerprint of the CSR
+func WithFingerprint(v interface{}) Options {
```
hslatman, line 102:
```suggestion
func WithFingerprint(v any) Options {
```

token/options.go
```diff
@@ -84,6 +86,40 @@ func WithSSH(v interface{}) Options {
...
+		default:
+			return fmt.Errorf("unsupported fingerprint for %T", v)
+		}
+
+		kid, err := fingerprint.New(data, crypto.SHA256, fingerprint.Base64RawURLFingerprint)
+		if err != nil {
+			return err
+		}
+		c.Set(ConfirmationClaim, map[string]string{
+			"x5rt#S256": kid,
```
hslatman, line 117: 👍

flags/flags.go
```diff
@@ -379,6 +379,19 @@ be stored in the 'sshpop' header.`,
...
+		Usage: `The <fingerprint> of the CSR to restrict this token for.`,
+	}
+
+	// ConfirmationFile is a cli.Flag used to add a confirmation claim in the
+	// tokens. It will add a confirmation kid with the fingerprint of the CSR.
+	ConfirmationFile = cli.StringFlag{
+		Name:  "cnf-file",
+		Usage: `The CSR <file> to restrict this token for.`,
+	}
+
```
hslatman, line 394: Should these include something about the SSH public key, or want to keep it simple?

command/certificate/fingerprint.go
```diff
@@ -128,7 +134,15 @@ func fingerprintAction(ctx *cli.Context) error {
...
-			return err
+			// Fallback to parse a CSR
+			csr, csrErr := pemutil.ReadCertificateRequest(crtFile)
+			if csrErr != nil {
+				return err
+			}
+			// We will only need the raw the generate a fingerprint.
+			certs = []*x509.Certificate{
+				{Raw: csr.Raw},
+			}
```
hslatman, line 145:
```suggestion
			// We will only need the raw DER bytes to generate a fingerprint.
			certs = []*x509.Certificate{
				{Raw: csr.Raw},
			}
```

command/ca/token.go
```diff
@@ -295,6 +323,31 @@ func tokenAction(ctx *cli.Context) error {
...
+			if err != nil {
+				return errors.Wrap(err, "error parsing ssh public key")
+			}
+			tokenOpts = append(tokenOpts, cautils.WithSSHPublicKey(sshPub))
+		} else {
+			csr, err := pemutil.ParseCertificateRequest(in)
+			if err != nil {
+				return errors.Wrap(err, "error parsing certificate request")
+			}
+			tokenOpts = append(tokenOpts, cautils.WithCertificateRequest(csr))
```
hslatman, line 345: Instead of passing the CSR or public key, the claim itself could be calculated here, and just pass it as the fingerprint directly? Or is there some flow in which to defer calculation of the fingerprint is necessary? `step ca sign`, maybe?

flags/flags.go
maraino: They used to include it, but due to a [comment](https://github.com/smallstep/certificates/pull/1660#discussion_r1547435804) in the certificates PR that suggested removing the cnf for SSH, I've removed it from here.

command/ca/token.go
maraino: `step ca sign` use `cautils.WithCertificateRequest(csr)` and `step ssh certificate` use `cautils.WithSSHPublicKey(sshPub)`, although the SSH one is currently totally ignored. These fields go to the shared context of the `cautils` package and end up in the `token` package, which takes care of calculating the fingerprint and adding the proper attribute to the final token. Right now, only the CSR gets into the token package, but sending the type allows us to specify the right attribute now that we don't share just `kid`.

utils/cautils/certificate_flow.go
maraino: Unfortunately, the cli.Context doesn't allow adding custom attributes. We would need to extend the cli.Context with a new type and refactor several commands to do this.

hslatman (approved):
Looks OK now. Not sure what's up with the test? It seems to want the old `kid` claim still?

PR #1106, opened 2024-02-11
Teach step ca init about --key-password-file

### LINKED ISSUE #453: "step ca init" with "--root" and "--key": add a way to pass the password to decrypt existing key file
When using "step ca init" with "--root" and "--key", could a functionality be added to pass the password to decrypt the existing key file ? It would not matter how the password would be supplied: the thing what matters to me is to avoid the prompt which asks for the password to decrypt the key file.

I would like to automate this process as much as possible, and therefore entering the password on the command prompt is not an option in my case.

larsks:
This commit permits you, when using the --root and --key options, to pass
the password for decrypting the key in --key-password-file rather than
requiring an interactive prompt.

Example usage:

    step ca init --root root.crt --key root.key \
      --key-password-file root_key_password ...

Closes #453

### DIFF
--- a/command/ca/init.go
+++ b/command/ca/init.go
@@ -37,7 +37,7 @@ func initCommand() cli.Command {
 		Action: cli.ActionFunc(initAction),
 		Usage:  "initialize the CA PKI",
 		UsageText: `**step ca init**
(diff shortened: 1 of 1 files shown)


maraino (approved):
Thanks for the contribution @larsks, it looks good to me

PR #1157, opened 2024-04-20
Allow stdin input of cert for needs-renewal

redrac:
#### Name of feature:
Allow stdin input of cert for needs-renewal

This allows users who have ephemeral certs (certs only in agents) to use needs-renewal without writing the cert to disk. This functionality already exists for inspect; I merely ported it over.

### DIFF
--- a/command/ssh/needsRenewal.go
+++ b/command/ssh/needsRenewal.go
@@ -77,25 +77,36 @@ Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
 }
 
 func needsRenewalAction(ctx *cli.Context) error {
-	if err := errs.NumberOfArguments(ctx, 1); err != nil {
-		return errs.NewExitError(err, 255)
+	if err := errs.MinMaxNumberOfArguments(ctx, 0, 1); err != nil {
+		return errs.NewExitError(errors.Wrap(err, "too many arguments"), 255)
+	}
+
+	var name string
+	switch ctx.NArg() {
+	case 0:
+		name = "-"
+	case 1:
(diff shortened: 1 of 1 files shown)


dopey:
Hey @redrac 👋. Thanks for submitting this PR, and apologies for the radio silence!

Overall, this looks good and useful. However, the error codes that are being returned by this command are intentional and in some cases I think the proposed changes may result in an error code that does not line up with the documentation. I will try to comment on those exact cases.

command/ssh/needsRenewal.go
```diff
@@ -77,25 +77,36 @@ Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
 }
 
 func needsRenewalAction(ctx *cli.Context) error {
-	if err := errs.NumberOfArguments(ctx, 1); err != nil {
-		return errs.NewExitError(err, 255)
+	if err := errs.MinMaxNumberOfArguments(ctx, 0, 1); err != nil {
```
dopey, line 80: I believe we want to exit with code `255` here.

command/ssh/needsRenewal.go
```diff
@@ -77,25 +77,36 @@ Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".`,
...
+	}
+
+	var name string
+	switch ctx.NArg() {
+	case 0:
+		name = "-"
+	case 1:
+		name = ctx.Args().First()
+	default:
+		return errs.TooManyArguments(ctx)
```
dopey, line 91: I believe we want to exit with code `255` here.

dopey (approved):

PR #1158, opened 2024-04-20
Allow users to define certificate comment in agent

redrac:
#### Name of feature:
Allow users to define certificate comment in agent

Added a comment flag which allows users to set the comment for a certificate when it gets added to an agent. It defaults to current behavior if not set, which is it uses the subject as the comment. This allows users who interact with multiple CAs with the same identity (email) to have multiple certificates in the agent. It also allows for use cases when users generate SSH certs with different extensions to load multiple certificates in their agent. One use case we currently have is we allow users to add the github extension to their certificate, but we also disable agent forwarding. A user might want to get a regular (non-github extension) certificate into their agent for regular SSHing with agent forwarding while also having another certificate that has the github extension that cannot be forwarded anywhere in their agent. Currently they cannot do this because step complains about duplicate certificates (because the subject is always the comment).


command/ssh/certificate.go
```diff
@@ -184,6 +184,10 @@ $  step ssh certificate --kty OKP --curve Ed25519 mariano@work id_ed25519
 			sshPrivateKeyFlag,
 			sshProvisionerPasswordFlag,
 			sshSignFlag,
+      flags.KTY,
+			flags.Curve,
+			flags.Size,
+      flags.Comment,
 			flags.KMSUri,
```
dopey, line 191:
```suggestion
			sshSignFlag,
                         flags.KTY,
			flags.Curve,
			flags.Size,
                         flags.Comment,
			flags.KMSUri,
```

dopey (approved):

dopey (approved):

Issue #845, opened 2023-02-12
Support querying OCSP and CRL in `step certificate validate` command


itspngu:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

It'd be useful if the step CLI offered a simple means of (optionally via flags, or by default) querying any CRL and OCSP endpoints listed in validated certificates' AIA fields. Additionally, warnings or errors should be issued if these endpoints are unreachable or behave in incompatible ways, to highlight potential misconfigurations - I'm not sure what kind of behaviour the according specifications mandate in these cases.

## Why is this needed?

Doing this with openssl and curl is cumbersome. Having this functionality as part of the step cli would suit its mission statement of being a swiss army knife for cryptographic operations very well.

dopey:
Hey @itspngu 👋 . Thanks for opening the issue! We think this is a great idea and we've added it to our next milestone. Unfortunately, we don't have a ton of time to devote to open source feature work at the moment, but we do plan on getting this in when we've got some spare cycles.

In the mean time, if someone from the community is interested in picking this one up, we'd be very glad to accept a PR. 
Cheers 🍻

hslatman:
Thanks for the heads up, @redrac 😄

PR #1161, opened 2024-04-23
Add OCSP and CRL support to certificate verify

redrac:
#### Name of feature:
Add OCSP and CRL support to certificate verify

Add args and functionality to certificate verify to check a CRL and OCSP for a certificate based on the extensions. Users can pass flags to enable verification of each (CRL, OCSP). The command will try and get the CRL and OCSP server from the certificate extensions if the endpoints are not provided.

Implements #845

#### Notes
- I tested the logic against a valid certificate, invalid OCSP/CRL endpoints, and a revoked certificate
- I moved some of the functions and structs from crl into its own internal package `internal/crlutil` so it can be reused
- Long term if [this](https://github.com/golang/go/issues/40017) ever is adressed then a lot of this can be removed in favor of using built in functionality in crypto/x509

### DIFF
--- a/command/certificate/verify.go
+++ b/command/certificate/verify.go
@@ -1,15 +1,22 @@
 package certificate
 
 import (
+	"bytes"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
+	"io"
+	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
 	"github.com/smallstep/cli/flags"
+	"github.com/smallstep/cli/internal/crlutil"
 	"github.com/urfave/cli"
 	"go.step.sm/cli-utils/errs"
 	"go.step.sm/crypto/x509util"
+	"golang.org/x/crypto/ocsp"
 )
 
 func verifyCommand() cli.Command {
@@ -18,7 +25,10 @@ func verifyCommand() cli.Command {
 		Action: cli.ActionFunc(verifyAction),
 		Usage:  `verify a certificate`,
 		UsageText: `**step certificate verify** <crt-file> [**--host**=<host>]
-[**--roots**=<root-bundle>] [**--servername**=<servername>]`,
+[**--roots**=<root-bundle>] [**--servername**=<servername>]
+[**--issuing-ca**=<ca-cert-file>] [**--verbose**]
+[**--verify-ocsp**]] [**--ocsp-endpoint**]=url
+[**--verify-crl**] [**--crl-endpoint**]=url`,
 		Description: `**step certificate verify** executes the certificate path
 validation algorithm for x.509 certificates defined in RFC 5280. If the
 certificate is valid this command will return '0'. If validation fails, or if
@@ -65,7 +75,24 @@ Verify a certificate using a custom directory of root certificates for path vali
 '''
 $ step certificate verify ./certificate.crt --roots ./root-certificates/
 '''
-`,
+
+Verify a certificate including OCSP and CRL using CRL and OCSP defined in the certificate
+
+'''
+$ step certificate verify ./certificate.crt --verify-crl --verify-ocsp
+'''
+
+Verify a certificate including OCSP and specifying an OCSP server
+
+'''
+$ step certificate verify ./certificate.crt --verify-ocsp --ocsp-endpoint http://acme.com/ocsp
+'''
+
+Verify a certificate including CRL and specificing a CRL server and providing the issuing CA certificate
+
+'''
+$ step certificate verify ./certificate.crt --issuing-ca ./issuing_ca.pem  --verify-crl --crl-endpoint http://acme.com/crl
+'''`,
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "host",
@@ -87,7 +114,32 @@ authenticity of the remote server.
     **directory**
 	:  Relative or full path to a directory. Every PEM encoded certificate from each file in the directory will be used for path validation.`,
 			},
+			cli.StringFlag{
+				Name:  "issuing-ca",
+				Usage: `The certificate issuer CA <file> needed to communicate with OCSP and verify a CRL. By default the issuing CA will be taken from the cert Issuing Certificate URL extension.`,
+			},
+			cli.BoolFlag{
+				Name:  "verify-ocsp",
+				Usage: "Verify the certificate against it's OCSP.",
+			},
+			cli.StringFlag{
+				Name:  "ocsp-endpoint",
+				Usage: `The OCSP endpoint to use. If not provided step will attempt to check it against the certificate's OCSPServer AIA extension endpoints.`,
+			},
+			cli.BoolFlag{
+				Name:  "verify-crl",
+				Usage: "Verify the certificate against it's CRL.",
+			},
+			cli.StringFlag{
+				Name:  "crl-endpoint",
+				Usage: "The CRL endpoint to use. If not provided step will attempt to check it against the certificate's CRLDistributionPoints extension endpoints.",
+			},
+			cli.BoolFlag{
+				Name:  "verbose, v",
+				Usage: "Print result of certificate verification to stdout on success",
+			},
 			flags.ServerName,
+			flags.Insecure,
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


command/certificate/verify.go
```diff
@@ -180,5 +247,129 @@ func verifyAction(ctx *cli.Context) error {
...
+		issuer, err = x509.ParseCertificate(issuerBlock.Bytes)
+		if err != nil {
+			return errors.Errorf("failed to parse the issuing CA certificate")
+		}
+	}
+
+	if verifyCRL {
+		var endpoint string
+		switch {
+		case crlEndpoint != "" && issuer.Equal(&x509.Certificate{}):
```
maraino, line 275: The second condition will be always false. If `issuerFile` is empty, `issuer` will be `nil`.

command/certificate/verify.go
```diff
@@ -180,5 +247,129 @@ func verifyAction(ctx *cli.Context) error {
...
+		if err != nil {
+			return errors.Wrap(err, "error parsing crl")
+		}
+
+		crlJSON, err := crlutil.ParseCRL(b)
+		if err != nil {
+			return errors.Wrap(err, "error parsing crl into json")
+		}
+
+		if !issuer.Equal(&x509.Certificate{}) && !insecure {
```
maraino, line 310: I think this should be:
```suggestion
		if issuer != nil && !insecure {
```

command/certificate/verify.go
```diff
@@ -102,11 +153,27 @@ func verifyAction(ctx *cli.Context) error {
...
+		insecure         = ctx.Bool("insecure")
 		intermediatePool = x509.NewCertPool()
 		rootPool         *x509.CertPool
 		cert             *x509.Certificate
+		issuer           *x509.Certificate
 	)
 
+	if issuerFile == "" && verifyOCSP {
+		return errors.Errorf("you must provide the issuing CA certificate if you want to verify the certificate with OCSP")
+	}
```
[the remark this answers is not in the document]
maraino: In most cases, the issuer for both the OCSP and CRL will be the same issuer of the certificate, so if the certificate passed as an argument contains an intermediate certificate, we might not need the issuer at all.

command/certificate/verify.go
```diff
@@ -180,5 +247,129 @@ func verifyAction(ctx *cli.Context) error {
...
+
+	if verifyOCSP {
+		var endpoint string
+		switch {
+		case ocspEndpoint != "":
+			endpoint = ocspEndpoint
+		case len(cert.OCSPServer) == 0:
+			return errors.Errorf("no OCSP AIA extension found")
+		default:
+			endpoint = cert.OCSPServer[0]
```
maraino, line 333: We probably want to check against all of them. There will be only one in most cases.

maraino (changes requested):
In most cases, if the certificate passed as a parameter contains an intermediate certificate, we can assume this is the issuer if the --ca flag is not provided.

command/certificate/verify.go
maraino: Sounds good

command/certificate/verify.go
```diff
@@ -87,7 +114,32 @@ authenticity of the remote server.
...
+			cli.BoolFlag{
+				Name:  "verify-crl",
+				Usage: "Verify the certificate against it's CRL.",
+			},
+			cli.StringFlag{
+				Name:  "crl-endpoint",
+				Usage: "The CRL endpoint to use. If not provided step will attempt to check it against the certificate's CRLDistributionPoints extension endpoints.",
+			},
+			cli.BoolFlag{
+				Name:  "verify-verbose",
```
maraino, line 138:
```suggestion
				Name:  "verbose, v",
```

command/certificate/verify.go
```diff
@@ -102,9 +154,18 @@ func verifyAction(ctx *cli.Context) error {
 		host             = ctx.String("host")
 		serverName       = ctx.String("servername")
 		roots            = ctx.String("roots")
+		verifyOCSP       = ctx.Bool("verify-ocsp")
+		ocspEndpoint     = ctx.String("ocsp-endpoint")
+		verifyCRL        = ctx.Bool("verify-crl")
+		crlEndpoint      = ctx.String("crl-endpoint")
+		verbose          = ctx.Bool("verify-verbose")
```
maraino, line 161:
```suggestion
		verbose          = ctx.Bool("verbose")
```

command/certificate/verify.go
```diff
@@ -18,7 +25,10 @@ func verifyCommand() cli.Command {
 		Action: cli.ActionFunc(verifyAction),
 		Usage:  `verify a certificate`,
 		UsageText: `**step certificate verify** <crt-file> [**--host**=<host>]
-[**--roots**=<root-bundle>] [**--servername**=<servername>]`,
+[**--roots**=<root-bundle>] [**--servername**=<servername>]
+[**--issuing-ca**=<ca-cert-file>] [**--verify-verbose**]
```
maraino, line 29:
```suggestion
[**--issuing-ca**=<ca-cert-file>] [**--verbose**]
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
+			case respReceived:
+				return err
+			case verbose:
+				fmt.Println(err)
+			default:
+			}
+		}
+
+		if !crlVerified {
+			return errors.Errorf("could not verify certificate against OCSP server(s)")
```
maraino, line 334:
```suggestion
			return errors.Errorf("could not verify certificate against CRL distribution point(s)")
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
+		}
+	}
+
+	if verbose {
+		fmt.Println(verboseMSG + "certficiate is valid")
+	}
 	return nil
 }
+
+func VerifyOCSPEndpoint(endpoint string, cert *x509.Certificate, issuer *x509.Certificate, httpClient *http.Client) (bool, error) {
```
maraino, line 376:
```suggestion
func VerifyOCSPEndpoint(endpoint string, cert, issuer *x509.Certificate, httpClient *http.Client) (bool, error) {
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
+	case ocsp.Revoked:
+		return true, errors.Errorf("certificate has been revoked according to OCSP %s", endpoint)
+	case ocsp.Good:
+		return true, nil
+	default:
+		return true, errors.Errorf("certificate status is unknown according to OCSP %s", endpoint)
+	}
+}
+
+func VerifyCRLEndpoint(endpoint string, cert *x509.Certificate, issuer *x509.Certificate, httpClient *http.Client, insecure bool) (bool, error) {
```
maraino, line 412:
```suggestion
func VerifyCRLEndpoint(endpoint string, cert, issuer *x509.Certificate, httpClient *http.Client, insecure bool) (bool, error) {
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
+		}
+
+		crlVerified := false
+		for _, endpoint := range endpoints {
+			respReceived, err := VerifyCRLEndpoint(endpoint, cert, issuer, httpClient, insecure)
+			switch {
+			case err == nil:
+				verboseMSG += fmt.Sprintf("certificate not revoked in CRL %s\n", endpoint)
+				crlVerified = true
+				break
```
maraino, line 324:
```suggestion
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
+		}
+
+		ocspVerified := false
+		for _, endpoint := range endpoints {
+			respReceived, err := VerifyOCSPEndpoint(endpoint, cert, issuer, httpClient)
+			switch {
+			case err == nil:
+				verboseMSG += fmt.Sprintf("certificate status is good according OCSP %s\n", endpoint)
+				ocspVerified = true
+				break
```
maraino, line 356:
```suggestion
```

command/certificate/verify.go
```diff
@@ -180,5 +241,212 @@ func verifyAction(ctx *cli.Context) error {
...
 	}
 
+	verboseMSG := "certificate validated against roots\n"
+	if host != "" {
+		verboseMSG += "certificate host name validated\n"
+	}
+
+	switch {
+	case (verifyCRL || verifyOCSP) && roots != "":
+		tlsConfig := &tls.Config{
```
maraino, line 251:
```suggestion
		//nolint:gosec // using default configuration for 3rd party endpoints
		tlsConfig := &tls.Config{
```

maraino (commented):
Hi @redrac, I've added suggestion that I think solves all the linter errors, and renames the verbose flag to `--verbose`, I can apply those and merge, unless you see something wrong with my changes.

maraino (approved):
lgtm

hslatman:
Late to the game, but is it an option to rename the flags to shorter versions, e.g. `--crl` and `--ocsp`? Given that they're used on the `verify` subcommand, I think the shorter names are sufficient. Or is it expected that those will be reused on (many) other subcommands in the future?

PR #1201, opened 2024-06-13
[action] update to common dependabot workflow | use SHAs for action versions

dopey:


maraino (approved):
Safer against take overs, but it makes it hard to update against vulnerabilities.

PR #1214, opened 2024-06-19
Makefile: install to /usr/local/bin, not /usr/bin

### LINKED ISSUE #1213: [Bug]: `make install` should install to /usr/local/bin, /usr/bin
### Steps to Reproduce

Run `make install`.

### Your Environment

* OS - Sonoma 14.5
* `step` CLI Version - master

### Expected Behavior

`make install` should install to /usr/local/bin, not /usr/bin.

It is my understanding that /usr/bin is only intended for OS binaries, and other applications should be installed into /usr/local/bin.

### Actual Behavior

`step` is placed in /usr/bin.

### Additional Context

_No response_

### Contributing

marten-seemann:
Fixes #1213.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,7 +124,7 @@ govulncheck:
 # Install
 #########################################
 
-INSTALL_PREFIX?=/usr/
+INSTALL_PREFIX?=/usr/local/
 
 install: $(PREFIX)bin/$(BINNAME)
 	$Q install -D $(PREFIX)bin/$(BINNAME) $(DESTDIR)$(INSTALL_PREFIX)bin/$(BINNAME)


hslatman:
Technically seems correct, but there might be reasons for installing it in `/usr/bin`. 

This might require updating documentation in some places.

maraino:
This `make install` command was used in the past to build the Debian package, but we now use `goreleaser` for this, so the `install` is no longer required. 

We can keep the command we the proper fixes, or we can simply remove it.

tashian (approved):
This seems fine to me.
`make install` is not used by any automations, as far as I know.
I created https://github.com/smallstep/certificates/pull/1918 to mirror this in `step-ca`.
and https://github.com/smallstep/step-ca-pro/pull/58 for `step-ca-pro`.
There may be other repos where we've used this Makefile as a template.

PR #1215, opened 2024-06-21
Restore check for the option `bundle` in the URL certificate inspection

francescopepe:
This PR reintroduces the check of the option `bundle` when a URL is inspected for certificates. Looks like it was removed [here](https://github.com/smallstep/cli/commit/576d8adba7daa388ff24827117bed7c418d7a923)


dopey:
Hey @francescopepe 👋. Thank you for catching and fixing this!! That was an embarrassing oversight 😳. It's a good thing we don't have tests, otherwise we could have prevented that (/s)

PR #1243, opened 2024-07-23
Goreleaser s3 upload not working

dopey:


tashian (approved):

.goreleaser.yml
```diff
@@ -262,16 +262,29 @@ release:
...
     acl: public-read
-    disable: '{{ ne .Prerelease "" }}'
+    extra_files:
+      - glob: ./dist/s3-versioned_*/**
+    extra_files_only: true
 
+  - provider: s3
+    disable: '{{ if .Prerelease }}true{{ else }}false{{ end }}'
+    ids:
+      - s3-unversioned
```
maraino, line 280: Ok, for now, I think we are building the same binaries twice, we could use `name_template` in the `extra_files` to change the name of the s3_versioned files.

PR #1273, opened 2024-09-12
Use `github.com/smallstep/cli-utils` and reorder imports

hslatman:


azazeal (approved):
much diff; didn't see any `rm -rf /` so I approve of it.

PR #1281, opened 2024-09-18
Sync commands of alpine & debian Dockerfile

768kb:
#### Name of feature:
Bugfix, chown command missin in debian Dockerfile variant.

Debian image of step-ca:hsm, which depends on step-cli:bookworm has permission issues. Cannot be changed as user later because step-ca Dockerfiles use VOLUME command on /home/step.


hslatman (approved):

hslatman:
Replaced by #1284

PR #1294, opened 2024-10-08
Add `--password-file` to `crypto jwe encrypt`

eest:
#### Name of feature:
Add `--password-file` to `crypto jwe encrypt`
Makes it possible to run the command without prompts for input if e.g. running it from a script to modify the password of a user as described in https://smallstep.com/docs/step-ca/provisioners/#changing-a-jwk-provisioner-password. The flag also matches the already available flag for the `decrypt` command. Also the usage string for the `decrypt` command has been updated to mention decryption. This fix basically mirrors how things are done in decrypt.go.
Helpful for scripting with the step client.
The flag includes a usage string.
Any using step cli.
N/A

💔Thank you!

### DIFF
--- a/command/crypto/jwe/decrypt.go
+++ b/command/crypto/jwe/decrypt.go
@@ -50,7 +50,7 @@ one of the JWKs in the JWK Set.`,
 			},
 			cli.StringFlag{
 				Name:  "password-file",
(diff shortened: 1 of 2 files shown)


eest:
@dopey is there something missing here for this to be accepted?

PR #1350, opened 2025-01-24
Handle the redirect-url flag

maraino:
### Description

This commit sets the proper redirect-url parameter in the defaults.json if the flag `--redirect-url` is given in the bootstrap command.

### DIFF
--- a/utils/cautils/bootstrap.go
+++ b/utils/cautils/bootstrap.go
@@ -219,6 +219,15 @@ func BootstrapTeamAuthority(ctx *cli.Context, team, teamAuthority string) error 
 		apiEndpoint = u.String()
 	}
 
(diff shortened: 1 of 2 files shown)


utils/cautils/bootstrap.go
```diff
@@ -219,6 +219,15 @@ func BootstrapTeamAuthority(ctx *cli.Context, team, teamAuthority string) error
 		apiEndpoint = u.String()
 	}
 
+	// Get the --redirect-url flag, If passed, we will use this one event if the
```
dopey, line 222:
```suggestion
	// Get the --redirect-url flag, If passed, we will use this one even if the
```

dopey (dismissed):

dopey (approved):

PR #1361, opened 2025-02-11
Skip opening the default browser if STEP_OPEN_BROWSER=false

jdoss:
This PR adds support for `STEP_OPEN_BROWSER` env var which if set to  `STEP_OPEN_BROWSER=false` it will skip opening the default browser and just output the authURL to stderr.

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -786,19 +786,19 @@ func (o *oauth) DoLoopbackAuthorization() (*token, error) {
 
 	if skipBrowser := os.Getenv("STEP_OPEN_BROWSER") == "0"; skipBrowser {
 		fmt.Fprintln(os.Stderr, authURL)
-		} else {
+	} else {
 		if err := exec.OpenInBrowser(authURL, o.browser); err != nil {
-				fmt.Fprintln(os.Stderr, "Cannot open a web browser on your platform.")
-				fmt.Fprintln(os.Stderr)
-				fmt.Fprintln(os.Stderr, "Open a local web browser and visit:")
-				fmt.Fprintln(os.Stderr)
-				fmt.Fprintln(os.Stderr, authURL)
-				fmt.Fprintln(os.Stderr)
+			fmt.Fprintln(os.Stderr, "Cannot open a web browser on your platform.")
+			fmt.Fprintln(os.Stderr)
+			fmt.Fprintln(os.Stderr, "Open a local web browser and visit:")
+			fmt.Fprintln(os.Stderr)
+			fmt.Fprintln(os.Stderr, authURL)
+			fmt.Fprintln(os.Stderr)
 		} else {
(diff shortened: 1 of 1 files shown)


command/oauth/cmd.go
```diff
@@ -784,18 +784,24 @@ func (o *oauth) DoLoopbackAuthorization() (*token, error) {
...
-		fmt.Fprintln(os.Stderr)
-		fmt.Fprintln(os.Stderr, "Open a local web browser and visit:")
-		fmt.Fprintln(os.Stderr)
-		fmt.Fprintln(os.Stderr, authURL)
-		fmt.Fprintln(os.Stderr)
+	skipBrowser := os.Getenv("STEP_OPEN_BROWSER") == "false"
+
+	if !skipBrowser {
+			if err := exec.OpenInBrowser(authURL, o.browser); err != nil {
+					fmt.Fprintln(os.Stderr, "Cannot open a web browser on your platform.")
```
hslatman, line 791: Don't apply this directly, but can you make it look like this:
```suggestion
	if skipBrowser := os.Getenv("STEP_OPEN_BROWSER") == "false"; skipBrowser {
			// no browser case
	} else {
	                // browser case
	}
```

We don't have many boolean env vars, but for the ones that do exist, we use numeric values 1/0 to toggle the behavior. That should be consistent. We could support both numeric values and strings, but then it should go into a helper func; potentially even in our `cli-utils` package.

There's an existing env var `STEP_INTEGRATION_DEBUG`, which seems to exist for testing purposes. This _could_ become `TEST_INTEGRATION_OPEN_BROWSER` to group that together.

jdoss: I can see use cases where people might want this URL like I do that are not using them for integration tests. I'll leave it as is for now. We can leave it undocumented for the time being to see if we want to change it later.

command/oauth/cmd.go
```diff
@@ -784,18 +784,22 @@ func (o *oauth) DoLoopbackAuthorization() (*token, error) {
...
+	if skipBrowser := os.Getenv("STEP_OPEN_BROWSER") == "0"; skipBrowser {
 		fmt.Fprintln(os.Stderr, authURL)
-		fmt.Fprintln(os.Stderr)
-	} else {
-		fmt.Fprintln(os.Stderr, "Your default web browser has been opened to visit:")
-		fmt.Fprintln(os.Stderr)
-		fmt.Fprintln(os.Stderr, authURL)
-		fmt.Fprintln(os.Stderr)
+		} else {
+		if err := exec.OpenInBrowser(authURL, o.browser); err != nil {
```
hslatman, line 790: Looks like the file needs to be formatted.

The version of the linter to use in CI was updated recently. It needs a rerun with fresh commit, I think.

hslatman:
@jdoss I've rebased your changes.

dopey (approved):

PR #1377, opened 2025-02-28
Allow a few global defaults to be pulled from the CA

dopey:
- min-encryption-password-length
- provisioner

Enforce min-password-length, if set, in the 'step ssh certificate' command.

### DIFF
--- a/flags/flags_test.go
+++ b/flags/flags_test.go
@@ -294,50 +294,6 @@ func TestFirstStringOf(t *testing.T) {
 			inputs: []string{"foo", "bar", "baz"},
 			want:   "",
 		},
-		/*
-			{
-				name:      "negative",
-				minLength: -5,
-				promptRun: promptRunner([]string{"foobar"}, nil),
-				want:      "foobar",
-				wantErr:   false,
-			},
-			{
-				name:      "zero",
-				minLength: 0,
-				promptRun: promptRunner([]string{"foobar"}, nil),
-				want:      "foobar",
-				wantErr:   false,
-			},
-			{
-				name:      "greater-than-min-length",
-				minLength: 5,
-				promptRun: promptRunner([]string{"foobar"}, nil),
-				want:      "foobar",
-				wantErr:   false,
-			},
-			{
-				name:      "equal-min-length",
-				minLength: 6,
-				promptRun: promptRunner([]string{"foobar"}, nil),
-				want:      "foobar",
-				wantErr:   false,
-			},
-			{
-				name:      "less-than-min-length",
-				minLength: 8,
-				promptRun: promptRunner([]string{"pass", "foobar", "password"}, nil),
-				want:      "password",
-				wantErr:   false,
-			},
-			{
-				name:      "ignore-post-whitespace-characters",
-				minLength: 7,
-				promptRun: promptRunner([]string{"pass   ", "foobar ", "password  "}, nil),
-				want:      "password",
-				wantErr:   false,
-			},
-		*/
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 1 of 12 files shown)


utils/cautils/bootstrap.go
```diff
@@ -25,9 +25,11 @@ import (
...
 
 type bootstrapAPIResponse struct {
-	CaURL       string `json:"url"`
-	Fingerprint string `json:"fingerprint"`
-	RedirectURL string `json:"redirect-url"`
+	CaURL             string `json:"url"`
+	Fingerprint       string `json:"fingerprint"`
+	RedirectURL       string `json:"redirect-url"`
+	Issuer            string `json:"issuer"`
+	MinPasswordLength int    `json:"min-password-length"`
```
maraino, line 32: We probably want `omitempty` and use provisioner instead of issuer
```
Provisioner string `json:"provisioner,omitempty"`
```

RedirectURL should also have an omitempty

maraino: This is the response, so it can be ok here.

utils/cautils/bootstrap.go
```diff
@@ -70,10 +86,12 @@ func withRedirectURL(r string) bootstrapOption {
...
-	CA          string `json:"ca-url"`
-	Fingerprint string `json:"fingerprint"`
-	Root        string `json:"root"`
-	Redirect    string `json:"redirect-url"`
+	CA                string `json:"ca-url"`
+	Fingerprint       string `json:"fingerprint"`
+	Root              string `json:"root"`
+	Redirect          string `json:"redirect-url"`
+	Issuer            string `json:"issuer,omitempty"`
+	MinPasswordLength int    `json:"min-password-length,omitempty"`
```
maraino, line 94: Omit empty on redirect-url and Provisioner instead of issuer.

utils/cautils/token_flow.go
```diff
@@ -326,7 +326,7 @@ func provisionerPrompt(ctx *cli.Context, provisioners provisioner.List) (provisi
 	}
 
 	// Filter by issuer (provisioner name)
-	if issuer := ctx.String("issuer"); issuer != "" {
+	if issuer := flags.FirstStringOf(ctx, "issuer", "provisioner"); issuer != "" {
```
maraino, line 329: Here and above, I would preffer `"provisioner", "issuer"`

maraino (dismissed):
lgtm, ideally the api should set `provisioner` too, but ...

flags/flags_test.go
```diff
@@ -230,3 +230,122 @@ func TestParseFingerprintFormat(t *testing.T) {
...
+				wantErr:   false,
+			},
+			{
+				name:      "ignore-post-whitespace-characters",
+				minLength: 7,
+				promptRun: promptRunner([]string{"pass   ", "foobar ", "password  "}, nil),
+				want:      "password",
+				wantErr:   false,
+			},
+		*/
```
maraino, line 340: Can you uncomment or remove these tests?

utils/cautils/bootstrap.go
```diff
@@ -25,9 +25,11 @@ import (
...
 
 type bootstrapAPIResponse struct {
-	CaURL       string `json:"url"`
-	Fingerprint string `json:"fingerprint"`
-	RedirectURL string `json:"redirect-url"`
+	CaURL             string `json:"url"`
+	Fingerprint       string `json:"fingerprint"`
+	RedirectURL       string `json:"redirect-url"`
+	Provisioner       string `json:"provisioner,omitempty"`
+	MinPasswordLength int    `json:"min-password-length,omitempty"`
```
maraino, line 32: Ok, but on responses omitempty is not necessary

maraino (approved):
lgtm

Issue #821, opened 2023-01-03
step cli flags with aliases sometimes report a different alias in error messages


weaversam8:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When using the `step ssh proxycommand` command in step CLI, I tried to configure a `--provisioner` flag, but accidentally provided the wrong provisioner name. When I did so, the error message I received was:

```
invalid value '<wrong provisioner name>' for flag '--issuer'
```

This initially confused me, as I didn't realize that `--issuer` and `--provisioner` are aliases for the same flag:

https://github.com/smallstep/cli/blob/160bd07ed5730f71e02a8f8394d6f63eb1b7fb92/flags/flags.go#L156-L160

Error messages should provide the name of the flag that was used, not the name of an alias that was not used.

## Why is this needed?

Inaccurate error messages are misleading and can cause a developer to believe they've made a mistake in a location other than the actual source of the problem.

dopey:
Hey @weaversam8 👋. Thanks for opening the issue!

First off, you're totally right, and this has also confused me multiple times. 

Second, for anyone reading this, we would definitely accept a PR to resolve this, as long as the PR didn't replace the underlying flag library that is being used (this would be a much broader change).

ap-kulkarni:
I did some analysis for this. From what I have understood, I think the information about which flag alias was passed on the command is lost after the parsing of command line is complete. The flag data structure in which the data about of the flag passed on the command line lies, does not have any information about what was passed on command line. Also neither of the structs `cli.Context`, `cli.App` & `cli.Command` save the actual command line, from which we could have cross-checked/extracted the alias of the flag.
One way to mitigate this could be to amend the help/error text and include all the aliases instead of only one. e.g. The error mentioned in issue description could look like
`invalid value '<wrong provisioner name>' for flag '--issuer'/'--provisioner'`
However, this would be a broader a change (in the sense that all such help/error messages involving aliases will need to be amended). But it won't be a change in logic, so there won't be a much of a testing overhead.

PR #1391, opened 2025-03-20
Fix some provisioner and policy prompt issues

hslatman:
This PR fixes the following issues:

* SCEP provisioners not detected in admin token flows. They now return an error, similar to ACME provisioners.
* Invalid provisioner selection logic when managing provisioner policies. The `--provisioner` flag was used to select a provisioner to authenticate as well as the provisioner to manage policies for.
* Unexpected error messages showing "issuer" instead of "provisioner" flag. In certain situations the CLI would return error messages indicating an issue with the `--issuer` flag value, whereas it was actually supplied in the `--provisioner` flag. This fixes https://github.com/smallstep/cli/issues/821.
* Update some usages of `issuer` in our code to reflect provisioner names.


maraino (commented):
It looks good, but I would try to see if doing this would work:
```go
ctx.Set("provisioner", "")
// and perhaps 
// ctx.Set("issuer", "")
```

maraino (dismissed):
It looks a bit happy, but it would be nice to find a different way to set the flag to `""` for just the things we need. Perhaps with token options instead of getting things from the context in this "flow" methods.

maraino (commented):
Looks good, but I would also set the alias to `""`. This will also fix the issue if the alias is in an environment variable or on defaults.json

maraino (approved):
lgtm

hslatman:
@leonweecs ~could you try the`--admin-provisioner` flag to select the provisioner to use for authentication instead?~

Sorry, I misread that. Let me try it.

hslatman:
@leonweecs in my local testing I'm not seeing the same behavior, unfortunately.

Could you verify the `admin.crt` and `admin.key` are _not_ present when you run the command? The CLI will prompt to overwrite if they do. If they _do_ exist, you could opt to use `--force` to overwrite them automatically. 

It doesn't explain why the provisioner picker still shows up in your example, though. Do you have a `defaults.json` in which an `admin-provisioner` is set, perhaps?

PR #1398, opened 2025-03-27
Add support for TSS2 keys in 'api token create' command.

joshdrake:
Adds support for parsing TSS2-wrapped keys to use in mTLS connection when requesting an API token.

💔Thank you!


joshdrake:
Superceded by https://github.com/smallstep/cli/pull/1399/files

PR #1399, opened 2025-03-27
Add support for creating API tokens backed by KMS signer (WIP)

hslatman:
This PR:

- Adds support for creating API tokens using mTLS with private keys backed by a KMS, such as a TPM. 
  - The use of a KMS is transparently detected by parsing the filename of the key to contain a (supported) KMS scheme.
  - TSS2 PEM encoded private keys are automatically detected based on the PEM header.
- Adds support for supplying a password for the private key using the `--password-file` flag.
- Adds `User-Agent` and `X-Request-Id` HTTP request headers.

Using an ECDSA key:

```console
$ step api token create <team-slug> ecdsa-chain.crt 'tpmkms:name=test-ecdsa;device=/path/to/tpmsimulator.sock'
✔ Token successfully created:
<token>
```

Using an RSA key:

```console
$ step api token create <team-slug> rsa-chain.crt 'tpmkms:name=test-rsa;device=/path/to/tpmsimulator.sock'
✔ Token successfully created:
<token>
```

Using the TSS2 PEM encoded private key (exported using `step tpm key get 'test-rsa' --device '/path/to/tpmsimulator.sock' --storage-directory /path/to/tpmstorage --tss2 > tpmrsa.tss2.pem`), specified as a `tpmkms` URI:

```console
$ step api token create <team-slug> rsa-chain.crt 'tpmkms:path=./tpmrsa.tss2.pem;device=/path/to/tpmsimulator.sock'
✔ Token successfully created:
<token>
```

Using the same TSS2 PEM encoded private key directly:
```console
$ step api token create <team-slug> rsa-chain.crt tpmrsa.tss2.pem --tpm-device '/path/to/tpmsimulator.sock'
✔ Token successfully created:
<token>
```

Note that the `device` and `tpm-device` flags and URI parts can be omitted in case the default system TPM locations are to be used. The use of a TPM simulator is just for demonstration and testing purposes.

PR #1402, opened 2025-04-03
Add initial version of the handshake command

maraino:
### Description

This commit adds the command `step certificate handshake`. This command performs a handshake and displays details about it.

For example:

```console
$ step certificate handshake smallstep.com
Server Name:  smallstep.com
Version:      TLS 1.3
Cipher Suite: TLS_AES_128_GCM_SHA256
KEM:          X25519
Insecure:     false
Verified:     true
```

### DIFF
--- a/command/certificate/certificate.go
+++ b/command/certificate/certificate.go
@@ -86,6 +86,7 @@ $ step certificate uninstall root-ca.crt
 			createCommand(),
 			formatCommand(),
 			inspectCommand(),
+			handshakeCommand(),
 			fingerprintCommand(),
 			lintCommand(),
 			needsRenewalCommand(),
--- a/command/certificate/certificate.go
+++ b/command/certificate/certificate.go
@@ -86,7 +86,6 @@ $ step certificate uninstall root-ca.crt
 			createCommand(),
 			formatCommand(),
 			inspectCommand(),
-			handshakeCommand(),
 			fingerprintCommand(),
 			lintCommand(),
 			needsRenewalCommand(),
--- a/command/certificate/handshake.go
+++ b/command/certificate/handshake.go
@@ -1,0 +1,260 @@
+package certificate
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"reflect"
+
+	"github.com/smallstep/cli-utils/errs"
+	"github.com/smallstep/cli/flags"
+	"github.com/smallstep/cli/internal/cryptoutil"
+	"github.com/urfave/cli"
+	"go.step.sm/crypto/pemutil"
+	"go.step.sm/crypto/x509util"
+)
+
+func handshakeCommand() cli.Command {
+	return cli.Command{
+		Name:        "handshake",
+		Action:      cli.ActionFunc(handshakeAction),
+		Usage:       `print handshake details`,
+		UsageText:   `**step certificate handshake** <url>`,
+		Description: `**step certificate handshake** displays detailed handshake information for a TLS connection.`,
+		Flags: []cli.Flag{
+			flags.ServerName,
+			cli.StringFlag{
+				Name: "tls",
+				Usage: `Defines the TLS <version> in the handshake. By default it will use TLS 1.3 or TLS 1.2.
+: The supported versions are **1.3**, **1.2**, **1.1**, and **1.0**.`,
+			},
+			cli.StringFlag{
+				Name:  "cert",
+				Usage: `The path to the <file> containing the client certificate to use.`,
+			},
+			cli.StringFlag{
+				Name:  "key",
+				Usage: `The path to the <file> or KMS <uri> containing the certificate key to use.`,
+			},
+			cli.StringFlag{
+				Name: "roots",
+				Usage: `Root certificate(s) that will be used to verify the
+authenticity of the remote server.
+
+: <roots> is a case-sensitive string and may be one of:
+
+    **file**
+	:  Relative or full path to a file. All certificates in the file will be used for path validation.
+
+    **list of files**
+	:  Comma-separated list of relative or full file paths. Every PEM encoded certificate from each file will be used for path validation.
+
+    **directory**
+	:  Relative or full path to a directory. Every PEM encoded certificate from each file in the directory will be used for path validation.`,
+			},
+
+			cli.StringFlag{
+				Name:  "password-file",
+				Usage: "The path to the <file> containing the password to decrypt the private key.",
+			},
+			cli.BoolFlag{
+				Name:  "chain",
+				Usage: "Print only the chain of verified certificates.",
+			},
+			cli.BoolFlag{
+				Name:  "peer",
+				Usage: `Print only the peer certificates sent by the server.`,
+			},
+			cli.BoolFlag{
+				Name: "insecure",
+				Usage: `Use an insecure client to retrieve a remote peer certificate. Useful for
+debugging invalid certificates remotely.`,
+			},
+		},
+	}
+}
+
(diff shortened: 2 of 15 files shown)


command/certificate/handshake.go
```diff
@@ -0,0 +1,260 @@
...
+	conn, err := tls.Dial("tcp", addr, tlsConfig)
+	if err != nil {
+		if tlsConfig.InsecureSkipVerify {
+			return tls.ConnectionState{}, fmt.Errorf("error connecting to %q: %w", addr, err)
+		}
+		tlsConfig.InsecureSkipVerify = true
+		return tlsDialWithFallback(addr, tlsConfig)
+	}
+	defer conn.Close()
+	conn.Handshake()
```
hslatman, line 258: Can use `HandshakeContext`, and it would be good to check and return the error. 

In a follow up we could implement some additional error handling logic for more informative errors based on some internal code we have. I think it could be nice to put that in `tlsutil`.

internal/cryptoutil/cryptoutil.go
```diff
@@ -115,6 +116,37 @@ func LoadCertificate(kmsURI, certPath string) ([]*x509.Certificate, error) {
...
+	var signer crypto.Signer
+	if IsKMS(keyName) {
+		if signer, err = CreateSigner(keyName, keyName, opts...); err != nil {
+			return tls.Certificate{}, err
+		}
+	} else {
+		if signer, err = CreateSigner("", keyName, opts...); err != nil {
+			return tls.Certificate{}, err
+		}
+	}
```
hslatman, line 136: I like this. I might be able to use it in #1399 too (when we support creating TSS2 signers from files).

internal/cryptoutil/cryptoutil.go
```diff
@@ -115,6 +116,37 @@ func LoadCertificate(kmsURI, certPath string) ([]*x509.Certificate, error) {
 	return cert, nil
 }
 
+// LoadTLSCertificate returns a [tls.Certificate] from a certificate fine and a
+// key in a file or in a KMS.
```
hslatman, line 120:
```suggestion
// LoadTLSCertificate returns a [tls.Certificate] from a certificate file and a
// key in a file or in a KMS.
```

command/certificate/handshake.go
```diff
@@ -0,0 +1,260 @@
...
+		for _, crt := range cs.PeerCertificates {
+			fmt.Print(string(pem.EncodeToMemory(&pem.Block{
+				Type: "CERTIFICATE", Bytes: crt.Raw,
+			})))
+		}
+		return nil
+	}
+
+	// Check if the certificates is verified
+	var intermediates *x509.CertPool
```
hslatman, line 187:
```suggestion
	// Check if the certificate is valid
	var intermediates *x509.CertPool
```

command/certificate/handshake.go
```diff
@@ -0,0 +1,260 @@
...
+	if u, ok, err := trimURL(addr); err != nil {
+		return err
+	} else if ok {
+		addr = u
+	}
+	if _, _, err := net.SplitHostPort(addr); err != nil {
+		addr = net.JoinHostPort(addr, "443")
+	}
+
+	// Load certificate and if
```
hslatman, line 116:
```suggestion
	// Load client TLS certificate 
```

command/certificate/handshake.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Print only the peer certificates
+	if printPeer {
+		for _, crt := range cs.PeerCertificates {
+			fmt.Print(string(pem.EncodeToMemory(&pem.Block{
+				Type: "CERTIFICATE", Bytes: crt.Raw,
+			})))
+		}
+		return nil
+	}
```
hslatman, line 184: Maybe needs an option to continue down, so that the connection details are shown in addition to these too? Or make that the default, instead of returning early?

command/certificate/certificate.go
```diff
@@ -86,6 +86,7 @@ $ step certificate uninstall root-ca.crt
 			createCommand(),
 			formatCommand(),
 			inspectCommand(),
+			handshakeCommand(),
```
hslatman, line 89: The functionality of this command is less concerned with managing certificates than the other subcommands of `certificate`. There's some semblance with `certificate inspect` and `certificate verify`, but this is more about the TLS handshake than it is about certificates. Thus instead of putting this under `certificate`, it could be an option to define a new `tls` command and put it under that. I don't have ideas for other subcommands immediately, though.

hslatman: @maraino we've discussed this during open source triage and after quite some discussion the new `tls` subcommand came out on top. 

@tashian suggested we could add a `tls inspect` (or similar) command (in addition to `tls handshake`) in the future that would probe the server more extensively than just the initial handshake, similar to e.g. https://www.ssllabs.com/ssltest/ and https://github.com/testssl/testssl.sh. 

Furthermore, we could add `ssh handshake` for the equivalent of this PR, but for SSH handshakes. All out of scope for this PR, though.

maraino: Fixed with 6a13fa3

command/certificate/handshake.go
maraino: I'm not sure about this. I make them independent commands because a certificate, especially if you include all the intermediates, is quite noisy. I also wanted to pipe those certificates to another command.

command/certificate/handshake.go
maraino: Fixed with 6a13fa36c1ed44304247d9a7c64b7f0495fae9e6

command/certificate/handshake.go
hslatman: That's fair, but in practice providing one of the `printPeer` or `printChains` options is doing the same as `step certificate inspect https://example.com` and `step certificate inspect https://example.com --bundle`, respectively. 

If instead execution would continue after printing the details, it would be a bit more like `curl -v` (with more certificate details, of course).

utils/utils_test.go
```diff
@@ -0,0 +1,49 @@
+package utils
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
```
hslatman, line 8:
```suggestion
import (
	"errors"
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"
```

command/tls/tls.go
```diff
@@ -0,0 +1,32 @@
+package tls
+
+import (
+	"github.com/urfave/cli"
+
+	"github.com/smallstep/cli-utils/command"
+)
+
+// Command returns the cli.Command for jwt and related subcommands.
```
hslatman, line 9:
```suggestion
// Command returns the cli.Command for tls and related subcommands.
```

command/tls/handshake.go
```diff
@@ -0,0 +1,261 @@
...
+	"net"
+	"reflect"
+
+	"github.com/smallstep/cli-utils/errs"
+	"github.com/smallstep/cli/flags"
+	"github.com/smallstep/cli/internal/cryptoutil"
+	"github.com/smallstep/cli/utils"
+	"github.com/urfave/cli"
+	"go.step.sm/crypto/pemutil"
+	"go.step.sm/crypto/x509util"
```
hslatman, line 18:
```suggestion
	"github.com/urfave/cli"

	"github.com/smallstep/cli-utils/errs"
	"go.step.sm/crypto/pemutil"
	"go.step.sm/crypto/x509util"

	"github.com/smallstep/cli/flags"
	"github.com/smallstep/cli/internal/cryptoutil"
	"github.com/smallstep/cli/utils"
```

PR #1430, opened 2025-05-29
Prefer VerificationURIComplete if present

### LINKED ISSUE #1424: Use VerificationURIComplete if present for OIDC device flow
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Right now, even if the IdP provides a `verification_uri_complete` parameter, step-cli ignores it. It should instead prefer it if present.

## Why is this needed?

It makes the login flow less tedious, as one only needs to click the link and then click yes in their browser, rather than copy the code, click the link, paste the code, hit enter, click yes.

See [kubelogin](https://github.com/int128/kubelogin/blob/c7342f301bb46775faaee5302725ea412b58f875/pkg/usecases/authentication/devicecode/devicecode.go#L32) for an example of a somewhat similar tool handling the same circumstance.

virtualdxs:
#### Name of feature:

Prefer VerificationURIComplete if present

In the OIDC Device Authorization Flow in command/oauth, present the user with VerificationURIComplete rather than VerificationURI if present, so that the user doesn't have to manually type the code when using an IdP that provides this feature.

No, the feature is automatically used if supported by the IdP

The OIDC Device Authorization Flow with an IdP that provides the `verification_uri_complete` parameter (e.g. Keycloak)

Fixes #1424

💔Thank you!

### DIFF
--- a/command/oauth/cmd.go
+++ b/command/oauth/cmd.go
@@ -892,6 +892,9 @@ func (o *oauth) DoDeviceAuthorization() (*token, error) {
 	}
 
 	switch {
+	case idr.VerificationURIComplete != "":
+		// Prefer VerificationURIComplete if present for user convenience
+		idr.VerificationURI = idr.VerificationURIComplete
 	case idr.VerificationURI != "":
 		// do nothing
 	case idr.VerificationURL != "":


virtualdxs:
I don't have any tests; if this change isn't too small to warrant test coverage I'll need help adding those as I haven't worked with test frameworks before and have barely worked with Go.

PR #1642, opened 2026-06-04
ssh: fix proxycommand stdin deadlock when server closes first

### LINKED ISSUE #1641: step ssh proxycommand hangs ~60s when server closes connection before stdin closes
## Description

`step ssh proxycommand` hangs for approximately 60 seconds when the SSH server closes the connection before the client has closed stdin. This occurs when the server rejects the connection mid-session (e.g. PAM account check failure after certificate authentication succeeds), sending a `SSH2_MSG_USERAUTH_BANNER` followed by a disconnect.

The process eventually exits due to an OS-level timeout, but the hang makes it appear to the user that the connection is still in progress.

## Suspected Cause

The deadlock is in `proxyDirect` / `proxyDirectWithIO` in `command/ssh/proxycommand.go`:

```go
var wg sync.WaitGroup
wg.Add(1)
go func() {
    io.Copy(conn, os.Stdin)   // goroutine 1: blocks reading stdin
    conn.CloseWrite()
    wg.Done()
}()
wg.Add(1)
go func() {
    io.Copy(os.Stdout, conn)  // goroutine 2: exits when server closes
    conn.CloseRead()
    wg.Done()
}()
wg.Wait()                     // waits for both — never returns
```

When the server closes the TCP connection:
1. Goroutine 2 exits and calls `conn.CloseRead()`
2. Goroutine 1 is blocked reading from `os.Stdin`
3. `os.Stdin` is a pipe from the SSH client process, which hasn't closed because it's waiting for the ProxyCommand to exit
4. The ProxyCommand is waiting for both goroutines — deadlock

Calling `os.Stdin.Close()` from goroutine 2 does not reliably interrupt a blocked `read()` syscall on macOS when stdin is a pipe.

## Reproduction

```go
// Start a TCP server that sends data and immediately closes
ln, _ := net.Listen("tcp", "127.0.0.1:0")
go func() {
    conn, _ := ln.Accept()
    conn.Write([]byte("hello"))
    conn.Close()
}()

// Simulate a stdin that never closes (SSH client waiting for ProxyCommand)
stdinR, _ := io.Pipe()  // write end intentionally left open

// This hangs indefinitely
proxyDirectWithIO("127.0.0.1", port, stdinR, io.Discard)
```

## Suggested Fix

In `command/ssh/proxycommand.go`, I think we can return early when either goroutine completes rather than waiting for both. Once one side closes, the ProxyCommand has nothing more to do — returning allows the OS to reclaim the blocked goroutine.

## Test

A regression test could simulate a TCP server that sends data and immediately closes, while holding stdin open indefinitely. Before the fix, `proxyDirectWithIO` would hang; after, it should return promptly.

## Environment

- macOS arm64 (Apple Silicon)
- `step` installed via Homebrew
- OpenSSH 9.9

techwolf359:
Fixes #1641

## Summary

`step ssh proxycommand` hangs when the SSH server closes the connection before the client has closed stdin. The hang lasts until an OS-level timeout (~60s on macOS) kills the process.

## Suspected cause

`proxyDirect` uses `sync.WaitGroup` waiting for both goroutines to finish:

- **Goroutine 1** (`stdin→server`): blocked reading `os.Stdin`, which is a pipe from the SSH client
- **Goroutine 2** (`server→client`): finishes when the server closes the connection

When the server closes, goroutine 2 exits. But goroutine 1 is stuck: `os.Stdin` won't close because the SSH client is waiting for the ProxyCommand to exit, and the ProxyCommand won't exit because goroutine 1 is still running — deadlock.

Calling `os.Stdin.Close()` from goroutine 2 does not reliably interrupt a blocked `read()` syscall on macOS when stdin is a pipe.

## Fix

Return as soon as either goroutine completes. The ProxyCommand's job is to relay bytes; once one side closes there is nothing more to do. The process exits and the OS reclaims the blocked goroutine.

## Changes

- Extracts `proxyDirectWithIO(host, port, in, out)` to enable testing without depending on `os.Stdin`/`os.Stdout`
- Replaces `wg.Wait()` (wait for both) with `<-done` (return on first)
- Adds `TestProxyDirectExitsWhenServerCloses` — reproduces the deadlock and verifies the fix (fails before, passes after)

## Testing

```
go test ./command/ssh/ -run TestProxyDirectExitsWhenServerCloses -v
```

### DIFF
--- a/command/ssh/proxycommand.go
+++ b/command/ssh/proxycommand.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"os"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -228,6 +227,10 @@ func getBastion(ctx *cli.Context, user, host string) (*api.SSHBastionResponse, e
 }
 
 func proxyDirect(host, port string) error {
+	return proxyDirectWithIO(host, port, os.Stdin, os.Stdout)
+}
+
+func proxyDirectWithIO(host, port string, in io.ReadCloser, out io.Writer) error {
 	address := net.JoinHostPort(host, port)
 	addr, err := net.ResolveTCPAddr("tcp", address)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


elliesalimi (approved):

techwolf359:
Closing this PR — we haven't been able to get a response on signing the contributor license agreement. Leaving the issue open with a suggested approach in case a maintainer or other contributor wants to pick it up.

PR #1647, opened 2026-06-09
Fix proxycommand hang when server closes before stdin closes

### LINKED ISSUE #1641: step ssh proxycommand hangs ~60s when server closes connection before stdin closes

tashian:
Fixes #1641 · Linear: PRE-124

## Problem

`step ssh proxycommand` hangs ~60s when the SSH server closes the connection before the client has closed stdin — e.g. when the server rejects the session mid-stream (PAM account check fails after cert auth succeeds).

The deadlock is in `proxyDirect`: it spawns two copy goroutines and `wg.Wait()`s for **both**. When the server closes, the `conn→stdout` goroutine returns, but the `stdin→conn` goroutine stays blocked reading `os.Stdin` — which never reaches EOF because the SSH client holds the proxycommand's stdin open until the proxycommand exits. Neither side gives, so `wg.Wait()` blocks until an OS-level timeout reaps the process.

## Fix

Return as soon as **either** direction finishes instead of waiting for both. For a proxycommand there's nothing left to do once one half of the SSH transport closes — returning lets the process exit and the OS reclaim the still-blocked goroutine. Replaced the `sync.WaitGroup` with a buffered `done` channel + `<-done`, and added `defer conn.Close()` so the surviving copy goroutine is also unblocked on the way out.

## Testing

Extracted a testable `proxyDirectWithIO(host, port, stdin, stdout)` and added `Test_proxyDirectWithIO_serverClosesBeforeStdin`: a TCP server that sends data then closes, with a stdin that never EOFs. Verified it hangs the full timeout against the old logic (RED) and returns promptly after the fix (GREEN, race-clean). `go build ./...`, `go vet`, and the full `command/ssh` package pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/command/ssh/proxycommand.go
+++ b/command/ssh/proxycommand.go
@@ -6,7 +6,6 @@ import (
 	"net"
 	"os"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/pkg/errors"
@@ -228,6 +227,10 @@ func getBastion(ctx *cli.Context, user, host string) (*api.SSHBastionResponse, e
 }
 
 func proxyDirect(host, port string) error {
+	return proxyDirectWithIO(host, port, os.Stdin, os.Stdout)
+}
+
+func proxyDirectWithIO(host, port string, stdin io.Reader, stdout io.Writer) error {
 	address := net.JoinHostPort(host, port)
 	addr, err := net.ResolveTCPAddr("tcp", address)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


maraino (approved):
lgtm

